PR #45, opened 2026-04-01
added setting for specifying language in desktop app

### LINKED ISSUE #48: No way to specify the transcription language in the desktop
There is no way to specify the language in the desktop, feature introduced https://github.com/silverstein/minutes/commit/f11083abde968b2aa629ca75a9b09442ef10df7d, is only limited to the cli.

valkrypton:
Fixes #48

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3036,7 +3036,7 @@ dependencies = [
 
 [[package]]
 name = "minutes-cli"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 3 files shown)


silverstein (changes requested):
Thanks for the contribution — the overall direction here is right, but there’s one behavior regression we should fix before merging. `transcription.language` is currently optional in config, so the desktop UI needs to preserve an explicit Auto-detect / unset state instead of forcing every user into a concrete language code. Right now the new select has only language codes, `cmd_set_setting` always writes `Some(value)`, and the settings load path doesn’t account for `null` cleanly. Please add an Auto-detect option (for example `value=""`), load `None` as that option, and map an empty selection back to `None` when saving. Once that round-trip is preserved, this feature should be in good shape.

Issue #49, opened 2026-04-01
Call detection misses Google Meet


valkrypton:
Problem
Call detection works by scanning running processes for known app names (zoom.us, Microsoft Teams, Webex). This covers native desktop apps but completely misses Google Meet, which runs as a tab in a browser — there is no dedicated Google Meet process to find.

Expected behavior:

When a user joins a Google Meet call (with mic active), Minutes should prompt to start recording, the same way it does for Zoom or Teams.

Current behavior

Google Meet calls are never detected. The browser process (Google Chrome, Safari, etc.) is far too broad to match against — it would trigger on any browser use.

Proposed solution

Add a "google-meet" sentinel to the detection app list. When the sentinel is present, query open browser tabs via AppleScript for meet.google.com (combined with the existing mic-in-use check to keep false positives low). Supported browsers: Chrome, Chrome Canary, Chromium, Safari.

PR #50, opened 2026-04-01
feat: detect Google Meet calls in browser tabs

valkrypton:
Fixes: #49 
Adds a "google-meet" sentinel to the call detection app list. When present, the detector queries open tabs in Chrome (and variants) and Safari via AppleScript for meet.google.com, combined with the existing mic-in-use check to keep false positives low.

- call_detect.rs: check_google_meet_in_browsers() uses the already- fetched process list to pre-filter inactive browsers before calling osascript, avoiding AppleScript overhead when no browser is open. Chromium-family browsers share a single query_chrome_for_meet() path. run_applescript() is a shared helper used by both Chrome and Safari.
- config.rs: "google-meet" added to default CallDetectionConfig apps.
- Tests: sentinel excluded from native process match, no-browser short-circuit, malformed AppleScript returns false gracefully, display_name_for("google-meet") == "Google Meet".

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3036,7 +3036,7 @@ dependencies = [
 
 [[package]]
 name = "minutes-cli"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "anyhow",
  "chrono",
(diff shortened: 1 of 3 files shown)


silverstein:
Closing in favor of #53 (already merged). Ali is credited as co-author on that work since this PR originated the feature.

Co-authored-by: Ali Tariq <74088901+valkrypton@users.noreply.github.com>

PR #116, opened 2026-04-13
fix(parakeet): parse word-level timestamps and group into segments

jotremblay:
## Problem

Running minutes with `engine = "parakeet"` always failed with:

```
parakeet transcription failed: could not parse parakeet output
(no [start - end] timestamps found). First 200 chars: ...
```

Two mismatches between `parse_parakeet_output` in `crates/core/src/transcribe.rs` and what parakeet.cpp actually emits with `--timestamps`:

1. **`s` suffix.** parakeet.cpp prints ``[0.08s - 0.16s] (0.60) Bon``; the parser did `start_str.trim().parse::<f64>()`, which fails on `0.08s`, so `has_timestamps` stayed `false` and every run errored out.
2. **Word-level output.** Each line is a single word, not a sentence. Even with #1 fixed, transcripts came out as one `[0:00]` line per word with the raw `(0.60)` confidence score leaking into the text.

Reproduced with the default `tdt-600m` and `tdt-ctc-110m` models on parakeet.cpp v0.x.

## Fix

- Strip the `s` suffix before parsing start/end seconds.
- Strip a leading `(0.42)` confidence prefix from the text.
- Group consecutive words into sentence-level segments, flushing on terminal punctuation (`.` `!` `?` `。` `！` `？`) or a gap > 0.8s between words.

Downstream filters (`dedup_segments`, `dedup_interleaved`, `strip_foreign_script`, `collapse_noise_markers`, `trim_trailing_noise`) now receive segment-level input matching what the whisper path gives them.

## Before / after

**Before** (parser error, pipeline aborts):
```
Error: parakeet transcription failed: could not parse parakeet output
(no [start - end] timestamps found). First 200 chars: Loading model: tdt-600m
...
```

**After** (17 s fr_CA + en sample, tdt-600m):
```
## Transcript

[0:00] Bon matin, aujourd'hui on va tester le modèle parakeet avec du français québécois...
```

## Test plan

- [x] Built with `cargo build --release -p minutes-cli --features 'minutes-core/parakeet'` on macOS 26 (arm64).
- [x] Ran `minutes process <file>.wav` on a 17 s fr_CA + en `say`-generated clip through `tdt-600m` — transcript is grouped, confidence scores gone, LLM summarization + markdown write downstream work end-to-end.
- [x] Existing parser tests still compile and pass (`cargo test -p minutes-core --features parakeet parse_parakeet`).

No change to the whisper path.

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -23,6 +23,7 @@ cuda = ["minutes-core/cuda"]
 hipblas = ["minutes-core/hipblas"]
 metal = ["minutes-core/metal"]
 vulkan = ["minutes-core/vulkan"]
+parakeet = ["minutes-core/parakeet"]
 
 [dependencies]
 minutes-core = { path = "../core", version = "0.11.0", features = ["streaming"] }
--- a/crates/core/src/transcribe.rs
+++ b/crates/core/src/transcribe.rs
@@ -1206,6 +1206,29 @@ fn parse_parakeet_output(
     let mut lines = Vec::new();
     let mut has_timestamps = false;
 
+    // Parakeet.cpp with --timestamps emits one line per *word*, e.g.
+    //   `  [0.08s - 0.16s] (0.60) Bon`
+    // We group words into sentence-level segments (flush on terminal
+    // punctuation or a gap > 0.8s) for a readable transcript.
+    let mut seg_start: Option<f64> = None;
+    let mut seg_prev_end: f64 = 0.0;
+    let mut seg_words: Vec<String> = Vec::new();
+    const GAP_THRESHOLD_SECS: f64 = 0.8;
+
+    let flush = |seg_start: &mut Option<f64>,
+                 seg_words: &mut Vec<String>,
+                 lines: &mut Vec<String>| {
+        if let Some(start_secs) = *seg_start {
+            if !seg_words.is_empty() {
+                let mins = (start_secs / 60.0) as u64;
+                let secs = (start_secs % 60.0) as u64;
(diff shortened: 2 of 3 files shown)


silverstein:
@jotremblay — great work here. You independently diagnosed both the `s` suffix parsing bug and the confidence prefix leaking into transcript text, and your sentence grouping logic (flush on terminal punctuation or >0.8s gap) is a genuinely good idea that we want to adopt.

Unfortunately, `main` moved fast in the same area over the last 24 hours. We landed comprehensive parakeet support including the timestamp/confidence fixes, model-specific tokenizer resolution, Metal GPU acceleration (`--gpu`), Tauri settings UI, health checks, and a new `parakeet.rs` module. So this PR now conflicts heavily and merging it directly would risk rolling back newer work.

The sentence grouping is the piece we don't have yet and want. Two options:

1. **You rebase onto current `main`** and focus the PR on just the word-to-sentence grouping, adapted to the new `crates/core/src/parakeet.rs` module. We'd love tests for punctuation flush, gap flush, and trailing flush.
2. **We port the grouping logic ourselves** with attribution to you in the commit message.

Whichever works better for you. Either way, your diagnosis was spot on and helped confirm the fixes we were landing independently. Thank you for the contribution.

jotremblay:
Thanks for the thorough review @silverstein — going with option 1. Opened #119 with a fresh branch off current `main`, focused narrowly on the word-to-sentence grouping, now living in `crates/core/src/parakeet.rs` as a pure function. Includes the three tests you asked for: punctuation flush, gap flush, and trailing flush. Closing this one in favor of #119.

jotremblay:
Superseded by #119 (rebased onto current main, narrowed to just the grouping logic).

PR #120, opened 2026-04-14
feat(call-detect): add Arc browser to Google Meet detection

czlonkowski:
## Summary

- Adds `Arc` to the browser list in `detect_google_meet_in_browsers`
- Arc is Chromium-based and its process name contains `"arc"`, so it works with the existing `ChromeLike` AppleScript probe with no additional changes
- Without this, users running Google Meet in Arc never see the call-detected banner

## Test plan

- [ ] Open Google Meet in Arc on macOS
- [ ] Confirm Minutes shows the "call detected" banner
- [ ] Confirm Automation permission prompt appears for Arc (first run)
- [ ] Confirm call capture starts correctly after accepting

### DIFF
--- a/tauri/src-tauri/src/call_detect.rs
+++ b/tauri/src-tauri/src/call_detect.rs
@@ -965,15 +965,16 @@ mod tests {
             cooldown_minutes: 5,
             apps: vec!["google-meet".into()],
         });
+        // Defer the probe so `detect_google_meet_in_browsers` skips the real
+        // AppleScript call to Arc but still records `saw_browser`.
+        detector.defer_browser_probe_for("Arc", "test");
+
         // "Arc" (the browser) must be recognised; "searchpartyd" must not
         // accidentally satisfy the check on its own.
         let running: Vec<String> = vec!["searchpartyd".into(), "Arc".into()];
-        // With Arc in the running list the probe attempts browser automation.
-        // We only assert that the result is NOT NoBrowserProcesses — i.e.
-        // `saw_browser` was set to true by the exact-match Arc entry.
-        assert!(!matches!(
+        assert!(matches!(
             detector.detect_google_meet_in_browsers(&running),
-            BrowserMeetProbe::NoBrowserProcesses
+            BrowserMeetProbe::NoMatch
         ));
     }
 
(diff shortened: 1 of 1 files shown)


silverstein (changes requested):
Thanks for the contribution! One concern before this lands:

The `"arc"` fragment is short enough to cause a false positive on stock macOS. Three always-running system processes contain `"arc"` as a substring:

```
searchpartyd
searchpartyuseragent
TrialArchivingService
```

Because the matcher is `running_lower.iter().any(|p| p.contains(proc_fragment))`, any user who **doesn't** have Arc installed will still trip `saw_browser = true` from `searchpartyd`. Then `query_browser_urls("Arc", ChromeLike)` runs an AppleScript against an app that doesn't exist → returns `AppleScriptProbe::Error` → the function `return`s early at call_detect.rs:499-501.

Since Arc is iterated **before** Safari in the list, the Safari probe never runs. Net effect: this change would silently break Google Meet detection for every Safari user on macOS who doesn't have Arc installed.

Suggested fix — tighten the match to equality for this entry, since `"Arc"` is the literal binary name:

```rust
// Arc's binary is exactly "Arc"; substring match would catch searchpartyd et al.
if !running_lower.iter().any(|p| p == "arc") {
    continue;
}
```

Or factor a small helper so Arc gets exact-match while the others keep substring-match.

Secondary thing worth confirming in the test plan: Arc's AppleScript `URL of tabs` support has historically been inconsistent compared to Chrome/Chromium. The description says it "works with the existing ChromeLike probe with no additional changes" — can you confirm you saw URLs actually returned in the `Urls(...)` branch, not just that the script didn't error?

czlonkowski:
Good catch on both points — addressed in the latest push.

**False positive fix:** Added an `exact` boolean field to the browser table. Arc uses `p == "arc"` (exact match); all existing entries keep `p.contains(proc_fragment)` unchanged. Added two unit tests:
- `arc_exact_match_does_not_fire_on_system_processes` — confirms `searchpartyd`, `searchpartyuseragent`, `TrialArchivingService` no longer set `saw_browser`
- `arc_exact_match_fires_on_arc_process` — confirms `"Arc"` itself is still recognised

**AppleScript confirmation:** Verified locally with `osascript -e 'tell application "Arc" to get URL of tabs of windows'` — Arc returns full URL list via the ChromeLike probe (not just a non-error). `meet.google.com/landing` was visible in the output, and a real meeting URL (`meet.google.com/abc-defg-hij`) would match `looks_like_google_meet_meeting_url`.

silverstein:
Merged, thanks! Pulled the branch locally and ran the call_detect suite. All 15 tests green, including your two new Arc cases. The exact-match guard is a nice fix, and the deterministic probe defer in the positive test is especially clean.

Arc support will ship in the next release.

Issue #114, opened 2026-04-10
Summarize language


ed0c:
Hi!

The transcription can be translated into another language (e.g., French). 
However, even if the transcription is in French, the summary is in English. 
I made the test on mac (gui) and linux(cli) , and the result is the same.

silverstein:
Thanks for the report! This is a straightforward fix. The summarization prompt didn't instruct the LLM to match the transcript language.

Fixed in the next release: the system prompt now tells the LLM to respond in the same language as the transcript (while keeping section headers in English for parsing).

If you're building from source, the fix is in `crates/core/src/summarize.rs`.

ed0c:
Thanks for the quick fix! The new behavior is much better for monolingual transcripts.

However, I'm still seeing the issue with **mixed-language transcripts** — e.g. a French meeting where participants occasionally use English terms or switch briefly to English. In that case, the LLM tends to produce the summary in English (or switch mid-summary), because the instruction to "match the transcript's language" is ambiguous when the transcript itself is mixed.

A `language` option in `[summarization]` would solve this cleanly:

```toml
[summarization]
language = "fr"  # ISO 639-1 — forces summary output language regardless of transcript content
                 # Default: auto (current behavior)
```

This mirrors the existing `[transcription] language` option and would give users an explicit override when the transcript language is ambiguous or mixed.

Happy to submit a PR if this seems like a good direction.

silverstein:
Great suggestion, this would be a welcome addition. Happy to take a PR for it.

Here's what I'd scope into the change based on tracing through the code:

**1. Add `language` to `[summarization]` config**
- New `language: String` field on `SummarizationConfig` in `crates/core/src/config.rs` (default `"auto"`)
- Mirrors the existing `[transcription] language` pattern

**2. Thread it into `SYSTEM_PROMPT`**
- When `language != "auto"`, replace the "Respond in the same language as the transcript" paragraph with an explicit instruction like `"Respond in {language}."`
- The `auto` path stays as-is (current behavior)

**3. Fix multi-chunk synthesis prompts (existing bug)**
- The reduce prompts at lines ~1100 and ~1283 in `summarize.rs` currently say "use the same KEY POINTS / DECISIONS / ACTION ITEMS format" but have zero language guidance. Even monolingual French transcripts can drift back to English during synthesis. The new `language` override should be threaded through here too.
[...]

PR #125, opened 2026-04-16
feat: add summarization.language config option

ed0c:
## Summary

Adds a `language` field to `[summarization]` in `config.toml`, allowing users to force the summary output language regardless of the transcript language.

This addresses the mixed-language transcript issue reported in #114 where the LLM would drift back to English during synthesis even when the per-chunk prompt was correctly set.

## Changes

- Add `language: String` field to `SummarizationConfig` (default: `"auto"`)
- Convert `SYSTEM_PROMPT` constant to `build_system_prompt(language)` function
- Convert `TITLE_REFINEMENT_PROMPT` constant to `build_title_prompt(language)` function
- Thread language instruction through multi-chunk synthesis/reduce prompts
- Add `get_effective_summary_language()` helper that falls back to `transcription.language` when `summarization.language` is `"auto"` — zero-config improvement for users who already set their transcription language
- Add tests for config deserialization and accented character parsing

## Usage

```toml
[summarization]
language = "fr"  # ISO 639-1 — forces summary output language
                 # Default: "auto" (matches transcript language, current behavior)
```

When `language` is `"auto"` but `[transcription] language` is set, the transcription language is used as fallback — so users who already configured their transcription language get better summaries without any extra config.

## Testing

- All existing tests pass
- Added `summarization_language_can_be_set_from_toml`
- Added `summarization_language_defaults_to_auto`
- Added `parse_summary_response_with_accented_characters`

Developed with AI assistance.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -166,6 +166,8 @@ pub struct SummarizationConfig {
     pub ollama_url: String,
     pub ollama_model: String,
     pub mistral_model: String,
+    pub language: String,
+
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -654,6 +656,7 @@ impl Default for SummarizationConfig {
             ollama_url: "http://localhost:11434".into(),
             ollama_model: "llama3.2".into(),
             mistral_model: "mistral-large-latest".into(),
+            language: "auto".into(),
         }
     }
 }
@@ -1383,5 +1386,28 @@ shortcut_enabled = false
         assert!(reloaded.contains("# top comment"));
         assert!(reloaded.contains("mystery = \"keep-me\""));
         assert!(raw_toml_has_section(&reloaded, "palette"));
+    }
+
+    #[test]
+    fn summarization_language_defaults_to_auto() {
+        let config = Config::default();
+        assert_eq!(config.summarization.language, "auto");
+    }
+
+    #[test]
+    fn summarization_language_can_be_set_from_toml() {
+        let dir = TempDir::new().unwrap();
+        let config_path = dir.path().join("config.toml");
+        std::fs::write(
+            &config_path,
+            r#"
+[summarization]
+language = "fr"
+"#,
+        )
+        .unwrap();
+
+        let config = Config::load_from(&config_path);
+        assert_eq!(config.summarization.language, "fr");
     }
 }
--- a/crates/core/src/summarize.rs
+++ b/crates/core/src/summarize.rs
@@ -220,7 +220,17 @@ pub fn format_summary(summary: &Summary) -> String {
     output
 }
 
-pub const TITLE_REFINEMENT_PROMPT: &str = r#"You create concise meeting titles.
+pub fn build_title_prompt(language: &str) -> String {
+    let lang_instruction = if language == "auto" {
+        String::new()
+    } else {
+        format!(
+            "\n- Always respond in {}. Regardless of the transcript language, the title must be in {}.",
+            language, language
+        )
+    };
+    format!(
+        r#"You create concise meeting titles.
 
 Given a meeting summary plus extracted structured content, produce a concise meeting title.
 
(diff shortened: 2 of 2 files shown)


silverstein (commented):
Nice work. I pulled this down and walked through the whole summarization path; the design and threading are correct. A few notes before merge.

## Coverage audit

I traced every code path that sends a prompt to an LLM and here is how the language override lands:

| Backend | Per-chunk prompt | Reduce/synthesis pass | This PR |
|---------|:-:|:-:|:--|
| Agent CLI | single-shot, no chunking | n/a | handled |
| Claude | ok | has synthesis at line 1135 | both threaded |
| Mistral | ok | has synthesis at line 1333 | both threaded |
| OpenAI | ok | no synthesis, concatenates chunks | per-chunk threaded; no reduce to fix |
| Ollama | ok | no synthesis, concatenates chunks | per-chunk threaded; no reduce to fix |

So this PR covers everywhere a language instruction needs to go. OpenAI and Ollama lacking a reduce step is a pre-existing structural issue (long transcripts produce concatenated, unnormalized output) and is correctly out of scope here. Worth filing separately.

## One blocker for CI

The new `get_effective_summary_language` trips `clippy::needless_lifetimes` on Rust 1.94+:

```rust
pub fn get_effective_summary_language<'a>(config: &'a Config) -> &'a str {
```

Clippy's suggested elision:

```rust
pub fn get_effective_summary_language(config: &Config) -> &str {
```

Same semantics, just lets Rust elide the lifetime. With `-D warnings` this currently fails macOS CI.

## Unrelated CI red

Ignore the current macOS failure from `search.rs` and `transcribe.rs`. Those are pre-existing regressions on `main` that affect every open PR. I put up #128 to unblock them. Once #128 lands you can rebase and your CI should go green on its own merits.

## Optional nice-to-haves

Not blockers:

1. Tests for `get_effective_summary_language` itself (explicit summary lang wins, fallback to transcription lang, both unset returns `"auto"`). You covered config parsing and prompt output already, adding three asserts for the fallback function makes the contract explicit.
2. Extra blank line after `pub language: String,` in `SummarizationConfig`. Cosmetic, up to you.

## Good stuff worth calling out

- The `summarization.language == "auto"` falling back to `transcription.language` is a really nice touch. Users who already configured their transcription language get better summaries with zero additional config. That was not in the original issue but it is exactly the right move.
- Accented-characters parser test is great. That kind of thing tends to break silently once a code path expects ASCII.
- Converting `TITLE_REFINEMENT_PROMPT` from a const to a builder function in the same pass was the right call, since titles should also respect the language setting.

Once you address the lifetime elision I will land this. Thanks for the thorough implementation.

ed0c:
Thanks for the thorough review!

Addressed the lifetime elision and added the three unit tests for `get_effective_summary_language` (explicit lang wins, fallback to transcription lang, both unset → auto).

Understood on the pre-existing CI red — will rebase once #128 lands.

silverstein:
CI on main is now unblocked as of #128. Rebase this PR and the red macOS check should clear, leaving just the `clippy::needless_lifetimes` fix on `get_effective_summary_language` to land.

silverstein:
Sorry, missed that you'd already pushed the lifetime elision and the three unit tests before my last comment. Thanks, that is exactly what I was asking for. Just rebase onto main when you get a chance and the macOS check should go green on its own merits. Then I will land it.

silverstein:
Went ahead and did the rebase for you since maintainer edits were enabled on this PR and I didn't want to make you wait on my cycles. Replayed your two commits cleanly on top of current main (which now includes #128 and #134); no conflicts. New head is `3196a3b`.

If you were working locally on this branch, a `git fetch origin && git reset --hard origin/feature/summarization-language` will pick up the rebased commits. (Or `git pull --rebase`, same idea.)

Queueing auto-merge so it lands as soon as CI comes back green.

silverstein:
Landed as `4e8760c`. Thanks ed0c, this was a pleasure to land.

The `summarization.language = "auto"` falling back to `transcription.language` was the standout. Users who already configured their transcription language now get correct-language summaries with zero extra config, and that wasn't in the original issue. Small bit of empathy that makes the feature feel designed rather than bolted on.

Also appreciated how thoroughly the reduce-pass threading was done. Easy to miss the synthesis step when you're only thinking about per-chunk prompts, and the gap would have silently re-surfaced #114 for long transcripts.

Shipping in the next release.

Issue #62, opened 2026-04-02
PipeWire auto-detection for --call on Linux


silverstein:
## Context

From Discussion #43 with @Dieterbe. Stock PipeWire is the default Linux audio stack, and `*.monitor` sources provide system audio capture by default. `minutes record --call` should auto-detect these without manual device setup.

## What needs to happen

1. **Detect PipeWire monitor sources** in `capture.rs` device enumeration. `*.monitor` sources from PipeWire should be recognized as system audio / loopback devices.
2. **`call = "auto"` config** should scan for these sources on Linux (the macOS path scans for BlackHole/Loopback/Soundflower).
3. **`minutes sources`** diagnostic command should categorize PipeWire devices correctly (Microphones / System Audio / Virtual Devices).

## Relevant code

- `crates/core/src/capture.rs` - device categorization already has some PipeWire/pulse name matching (~line 822)
- `docs/MULTI-SOURCE-CAPTURE.md` - design doc covering the full architecture
- `crates/cli/src/main.rs` - `sources` subcommand scaffolding

## Notes

- This is scoped to device detection and auto-selection, not the full source-aware attribution pipeline
- @Dieterbe confirmed the desktop app builds and runs on Arch/Wayland/Hyprland (Discussion #44)
- PipeWire monitor sources are actually easier to detect than macOS loopback devices since they're available by default

Related: #34, Discussion #43

Dieterbe:
> Stock PipeWire is the default Linux audio stack

if not that, at least pulse-audio; i'm not an expert on it, but it seems pipewire also emulates pulse-audio and i think the PA api's are the best ones to target (whether implemented by just PA or by pipewire), needs confirmation to be sure

Dieterbe:
on current main branch:
```
./target/debug/minutes sources
ALSA lib pcm_dmix.c:1000:(snd_pcm_dmix_open) [error.pcm] unable to open slave
ALSA lib pcm_oss.c:404:(_snd_pcm_oss_open) [error.] Cannot open device /dev/dsp
ALSA lib pcm_oss.c:404:(_snd_pcm_oss_open) [error.] Cannot open device /dev/dsp
ALSA lib pcm_dmix.c:1000:(snd_pcm_dmix_open) [error.pcm] unable to open slave
Microphones:
    sysdefault (44100Hz, 2 ch)
    lavrate (44100Hz, 2 ch)
    samplerate (44100Hz, 2 ch)
    speexrate (44100Hz, 2 ch)
[...]

PR #133, opened 2026-04-16
fix(capture): stabilize pinned-device capture on Linux dual-host cpal

### LINKED ISSUE #130: Linux cpal host race + DeviceMonitor reconnect loop with pinned recording device
**This was generated by Claude, so take with a pinch of salt or feel free to delete if not deemed useful. I don't have any experience with developing Rust packages or Linux audio debugging. That said, the local fix works.**

Surfaced while testing workarounds for #62 / #69. 

## Summary

On Linux with both PipeWire and ALSA cpal backends compiled in, `minutes record` either errors with `audio device not found` or enters a reconnect loop that resets the capture stream every ~2s, dropping samples and producing hallucinated transcripts when loopback audio is the source.

Two bugs in `crates/core/src/capture.rs`:

1. **`cpal::default_host()` is nondeterministic.** Successive calls in the same process return different `HostId`s (observed: `PipeWire` then `Alsa` 650µs apart). `resolve_capture_plan` and `record_to_wav` each call `default_host()` independently. If the pinned `[recording] device = "..."` name only exists on one backend, the second call fails with `requested audio device not found`. Re-creating a PipeWire host via `host_from_id(PipeWire)` sometimes enumerates zero devices on the first `input_devices()` call.

2. **`DeviceMonitor` polls `default_input_device()` with no notion of pinning.** When the user pinned a device via config/`--device`, the monitor still asks cpal for the current default, sees a name mismatch (e.g. tracked=`pulse`, cpal default=`default`), and triggers reconnect every 2s. Spams log and drops audio.

## Repro

Pop!_OS 22.04, PipeWire 1.0.3 + PulseAudio bridge, both backends live. `minutes 0.12.1` built from `main` with default features.

```toml
[recording]
device = "sink_default"   # only exists on PipeWire host
```

```
$ minutes record
INFO cpal host for input device selection host_id=PipeWire
INFO using requested input device device=sink_default
Recording meeting...
INFO cpal host for input device selection host_id=Alsa
ERROR requested audio device not found requested=sink_default
```

With `device = "pulse"` (ALSA pulse plugin) recording starts but then:

```
INFO default audio device changed — will reconnect
INFO using requested input device device=pulse
INFO audio stream reconnected device=pulse
(repeats every 2s until stop)
```

Transcripts from 60s podcast loopback come back as garbled Canadian Aboriginal Syllabics or `[BLANK_AUDIO]` because whisper sees the shredded stream.

## Fix

Implemented locally, happy to PR:

1. `select_input_device` keeps a process-wide `OnceLock<Mutex<Option<HostId>>>` of the host that successfully resolved a pinned device. Subsequent calls short-circuit to that host regardless of what `default_host()` returns.
2. When the primary host lacks the requested device, iterate `cpal::available_hosts()` and try `host_from_id` with a 3-attempt retry (25ms between) on `input_devices()` to work around PipeWire's empty-list first enumeration.
3. `DeviceMonitor::pinned(name)` constructor that always returns `false` from `has_device_changed()`. `record_to_wav` uses it when `device_override` is `Some`.

With the fix on this machine, `device = "PulseAudio Sound Server"` resolves to ALSA host, gets cached, no reconnect loop, loopback captures cleanly.

## System

- OS: Pop!_OS 22.04 LTS, kernel 6.12.10
- Audio: PipeWire 1.0.3 with PulseAudio 35 bridge
- cpal: 0.18.0 git `8b152c39`, features `["pipewire"]`
- rustc: 1.91.0
- minutes: 0.12.1

jamesvrt:
**This was generated by Claude, so take with a pinch of salt or feel free to delete if not deemed useful. I don't have any experience with developing Rust packages or Linux audio debugging. That said, the local fix works.**

## Summary

Fixes pinned-device recording on Linux builds with both PipeWire and ALSA cpal backends compiled in. Two root causes, both called out by the maintainer review:

1. **cpal host nondeterminism.** `cpal::default_host()` can return different `HostId`s on successive calls in the same process. `resolve_capture_plan` and `record_to_wav` each call it independently, so pinned device names (`"sink_default"` on PipeWire, `"pulse"` on ALSA, etc.) could fail with `requested audio device not found` when the second call landed on the other host. Fix caches the first `HostId` that successfully resolves a pinned device in a process-wide `OnceLock<Mutex<Option<HostId>>>` and iterates `cpal::available_hosts()` as a fallback. Also retries `input_devices()` up to 3× (25ms apart) to work around PipeWire sometimes reporting an empty device list on the first enumeration after `host_from_id`.

2. **`DeviceMonitor` ignored pinning.** It polled `default_input_device()` and compared names against the tracked device. When the user pinned `"pulse"` but cpal reported `"default"`, the monitor flagged a change every 2s and reset the stream, dropping samples and producing hallucinated/garbled transcripts on loopback capture. New `DeviceMonitor::pinned()` constructor returns `false` from `has_device_changed()` and skips the macOS CoreAudio listener.

Wired the pinned monitor through all three capture entry points per review feedback: `record_to_wav`, `dictation::run`, `live_transcript::run`, and the dual-source path via `MultiDeviceMonitor::with_pinned`.

## Repro (Pop!_OS 22.04, PipeWire 1.0.3 + PulseAudio bridge)

```toml
[recording]
device = "PulseAudio Sound Server"
```

Before: `minutes record` either errored with `audio device not found` or reconnected every 2s, shredding loopback audio. After: stable capture, no reconnect loop, clean transcript.

## Tests

- `device_monitor::tests::pinned_monitor_never_flags_change` — pinned monitor ignores device-change signal.
- `device_monitor::tests::multi_device_monitor_with_pinned_voice_never_flags` — multi-monitor with both sides pinned reports no changes.
- `capture::tests::preferred_host_cache_round_trips` — `set_preferred_host_id` → `preferred_host_id` round-trips the cached `HostId`. Uses a serial guard (process-wide static) and restores prior state to not pollute parallel tests.

## CI parity

No new `cfg(unix)`-gated paths. Host cache and `select_input_device` changes are cross-platform. `DeviceMonitor::pinned` gates the existing `#[cfg(target_os = "macos")]` CoreAudio listener with an `if pinned` branch, no new platform cfg surface.

## Test plan

- [ ] `cargo test -p minutes-core --no-default-features --lib` passes (new tests included).
- [ ] `cargo fmt --all -- --check` clean.
- [x] On a Linux machine with PipeWire + ALSA both live, `minutes record` with a pinned device records cleanly with no reconnect-loop log spam.
- [ ] On macOS, unpinned recording still auto-reconnects on device change (CoreAudio listener path unchanged for the non-pinned branch).

Resolves #130


silverstein:
Thanks James. This is a great PR. Both root causes addressed cleanly, tests cover the new behavior, and the wiring through all three capture entry points plus `MultiDeviceMonitor::with_pinned` is exactly the shape we were looking for. The `searched_hosts` addition to the error log is a nice touch for anyone debugging a future variant.

Reviewed the structural angles before merging: host cache is race-safe via `OnceLock` + `Mutex` with graceful poisoning fallback, multi-host iteration self-heals a stale cache, no new `cfg(unix)` or Windows-fragile paths, and pinned mode correctly short-circuits all reconnect triggers (debounce, default-device poll, and the CoreAudio listener registration on mac).

Merging now. Appreciate the thorough repro and the explicit CI parity note, it made review fast.

silverstein:
One follow-on while your PipeWire test setup is still warm, if you are interested: #62 (PipeWire auto-detection for `--call` on Linux) is tagged `help wanted` and is the issue your `device = "PulseAudio Sound Server"` workaround is essentially solving by hand. You know these cpal host quirks better than anyone after this PR. No pressure, just flagging in case it appeals.

PR #134, opened 2026-04-16
feat(call-detect): detect Microsoft Teams web meetings in Arc/Safari/Chrome

czlonkowski:
## Summary
- Adds an opt-in `teams-web` sentinel that mirrors the existing `google-meet` plumbing — settings toggle, mic-gated browser probe, sticky re-notify window, per-browser Apple-Events backoff. Native Teams desktop detection is unchanged.
- The browser probe now reads each tab's URL **and** title via AppleScript (Chromium tabs expose `title`, Safari `name`). URL matching catches static patterns (`teams.live.com/meet/<id>`, `…/meetup-join/…`, `…/pre-join-calling/…`); when the URL is the opaque post-join `teams.live.com/v2/`, a localized title-prefix check (Spotkanie / Meeting / Réunion / Besprechung / 会議 / …) disambiguates the meeting tab from the chat home, which always starts with the localized "Chat" prefix in the user's UI locale.
- Probe errors now include the osascript stderr in the backoff log, so future TCC / script issues are diagnosable without instrumenting a build.

## Why
The Teams web client (`teams.live.com`, `teams.microsoft.com`) is the missing major call surface in the existing browser probe — Google Meet works, but Teams users joining via Arc/Safari/Chrome were silently uncaught. The `/v2/` SPA hides its hash route from `URL of t`, so URL matching alone is insufficient; combining it with `document.title` (already cheap to fetch via the same AppleScript pass) gives a reliable disambiguator without enabling JavaScript-from-Apple-Events or any additional permissions.

## Test plan
- [x] `cargo test -p minutes-app call_detect` — 20 passed, 0 failed (3 new tests + the `teams_web_enabled` settings toggle test)
- [x] Dogfood in `~/Applications/Minutes Dev.app`: enabled the new "Microsoft Teams (web) detection (experimental)" toggle, joined a Teams Live meeting in Arc, confirmed "Teams call detected" notification fired within ~15 s
- [x] Negative case: chat tab on `teams.live.com/v2/` (title "Czat | … | Microsoft Teams") with mic active does NOT trigger
- [x] Pre-join URL (`teams.live.com/meet/<id>`) and the launcher URL (`dl/launcher/launcher.html?…&type=meet&…`) caught by the URL matcher; post-join (`teams.live.com/v2/`) caught by the title fallback

## Notes for reviewer
- `TEAMS_MEETING_TITLE_PREFIXES` is a curated lowercased prefix list across ~20 locales. It's intentionally a fixed list rather than a regex / external file — extending it as users report new locales is the right cost/value tradeoff.
- The simplifier pass collapsed the per-provider sticky helpers into shared `remember_sticky` / `sticky_alive` free functions and a `MeetingProvider::names()` table, so a third browser-meeting provider would be one match arm + one sticky field rather than four duplicated methods.
- Permission-denied notification copy is now provider-agnostic ("Browser meeting detection needs access") since the same probe handles both Meet and Teams.

### DIFF
--- a/tauri/src-tauri/src/call_detect.rs
+++ b/tauri/src-tauri/src/call_detect.rs
@@ -55,6 +55,9 @@ pub struct CallDetector {
     /// Recent successful browser-based Meet detection. Prevents fast native-app
     /// polling from immediately relabeling the same active session as Slack.
     recent_google_meet_until: Mutex<Option<Instant>>,
+    /// Recent successful browser-based Teams detection. Same role as the Meet
+    /// sticky field but for Microsoft Teams in a browser tab.
+    recent_teams_web_until: Mutex<Option<Instant>>,
     /// Log mic-gate transitions once instead of spamming every poll.
     last_mic_live: Mutex<Option<bool>>,
 }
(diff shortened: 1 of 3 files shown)


silverstein:
Merged. Thanks Romuald, this one was a pleasure to review.

The `/v2/` SPA disambiguator is exactly the right shape: URL-first, title fallback only when the URL is ambiguous, curated locale prefix list bounded to a reasonable size. The cleanup pass that collapsed the per-provider sticky helpers into `remember_sticky` / `sticky_alive` + `MeetingProvider::names()` is what I'd have asked for on review. A third browser provider genuinely is one match arm now.

Also appreciated the stderr snippet in the backoff log. Next time someone's probe misbehaves we'll actually have something to look at.

Shipping in the next release.

Issue #21, opened 2026-03-25
Transcription loops on non-English (Italian) audio — Greedy best_of=1 causes repetition


magno73:
## Bug

When processing Italian audio files via `minutes process`, the transcription falls into a repetition loop, producing gibberish like:

```
[0:00] "H Damn ya 18 Anno!"
[0:03] "...H- Hug o F- Ok預備 le Macedi diesel Only tr cloudy"
[0:31] *Risate* "H- Hug o F- Ok, bebe le Macedi diesel Only tr cloudy"
[0:55] *Risate* "H- Hug o F- Ok, bebe le Macedi diesel Only tr cloudy"
...
```

The same audio file transcribes perfectly with `whisper-cli` (the whisper.cpp CLI) using the same `ggml-small.bin` model:

```
[00:00:00.000 --> 00:00:06.160]   Sì, è che avrà molto spesso di essere incastrato.
[00:00:06.160 --> 00:00:11.840]   Bene, allora ho sentito Giuseppe che mi ha detto che aveva un problema...
```

## Environment

- minutes 0.7.3 (built from source with `--features metal`)
- macOS 15, Apple M4
- Model: `ggml-small.bin` (466MB)
- Config: `language = "it"` in `[transcription]`
- Audio: 30-minute Italian meeting recording (.m4a, 30MB)

## Root cause

`transcribe.rs` line 91 uses:

```rust
let mut params =
    whisper_rs::FullParams::new(whisper_rs::SamplingStrategy::Greedy { best_of: 1 });
```

The whisper.cpp CLI defaults are `best_of=5`, `beam_size=5`, `entropy_thold=2.4`, `logprob_thold=-1.0`, `temperature=0.0`, `temperature_inc=0.2`. These defaults exist specifically to prevent decoder loops: when a segment's entropy exceeds the threshold, whisper.cpp retries at a higher temperature, breaking the repetition cycle.

With `Greedy { best_of: 1 }` and no entropy/logprob thresholds, the decoder gets stuck on a low-confidence segment and repeats it for the rest of the audio.

## What I tried

I tested three patches locally:

1. **`Greedy { best_of: 5 }`** + entropy/logprob thresholds + `set_temperature(0.0)` + `set_temperature_inc(0.2)` → Still loops (different pattern: "Ok, bene, vedi?" repeated)
2. **`BeamSearch { beam_size: 5, patience: 1.0 }`** + same thresholds → Still loops (pattern: "*Morso*" repeated)
[...]

PR #139, opened 2026-04-16
build(macos): enable Metal in release binaries for 2× faster large-v3

czlonkowski:
## Summary

- macOS release binaries (DMG + published CLI) are compiled without any whisper.cpp GPU backend, so every transcription falls back to BLAS/CPU. This PR flips on `--features metal` in both macOS release workflows and the local build scripts.
- No Rust changes. Only `.github/workflows/release-*.yml`, `scripts/build.sh`, `scripts/install-dev-app.sh`, and the GPU-acceleration section of `README.md`.
- Windows + Linux release builds are unchanged (matrix keeps them on `parakeet` only).

## Motivation

Benchmarked on an M-series Mac, `large-v3` on an 8m40s Polish meeting (`519.6s` audio):

| Build | Wall time | Realtime | User CPU | Peak RSS | Words |
|---|---:|---:|---:|---:|---:|
| CPU-only (current release = `--features parakeet`) | `151.58s` | **3.43×** | `804.40s` | `4.06 GB` | 1284 |
| **Metal (this PR)** | **`75.38s`** | **6.89×** | `5.08s` | `3.54 GB` | 1284 |
| Metal + CoreML cold | `79.43s` | 6.54× | `5.18s` | `5.78 GB` | 1286 |
| Metal + CoreML warm | `65.00s` | 7.99× | `4.69s` | `3.54 GB` | 1286 |

Key observations:
- Wall time halves (`151s → 75s`) just by adding `metal`.
- User CPU time collapses `804s → 5s` — work moves to the GPU, so the M-series cores stay free for the rest of the system.
- Peak RSS drops ~500 MB with Metal.
- Transcripts are byte-identical between CPU and Metal (same 1284 words). CoreML is a minor encoder variation (1286).
- CoreML is faster on the warm path (`65s`), but it's left opt-in in this PR because `whisper.cpp`'s CoreML path silently falls back to CPU unless the companion `ggml-<model>-encoder.mlmodelc` bundle is present next to the `.bin`. `minutes setup` doesn't fetch that today — proposed as a follow-up PR.

## Design doc context

`docs/designs/apple-native-backend-candidates-2026-04-13.md` already concluded (2026-04-14 addendum) that the cheap/shippable first step before any WhisperKit sidecar work is just to turn on `whisper.cpp` GPU acceleration. This PR implements that step. WhisperKit and `parakeet-mlx` paths remain future work per the design doc — MLX itself has no ANE support, so the eventual "true" Apple-native answer is still the Swift WhisperKit sidecar (tracked as `minutes-ty6k-a`), not MLX.

## Changes

- `.github/workflows/release-cli.yml` — lift `features` into the matrix so the `aarch64-apple-darwin` target builds with `parakeet,metal`; Windows/Linux unchanged.
- `.github/workflows/release-macos.yml` — `cargo tauri build --features parakeet,metal --bundles app,dmg`.
- `scripts/build.sh` and `scripts/install-dev-app.sh` — same feature set for local dogfood builds so developer machines match what CI ships.
- `README.md` — \"GPU acceleration\" section now documents that Metal is the default in the macOS release, includes a CoreML note pointing at the required `.mlmodelc` bundle, and leaves the non-Apple backends marked experimental.

## Test plan

- [ ] `Release CLI Binaries` CI passes on `aarch64-apple-darwin` with the new features flag
- [ ] `Release macOS` CI successfully builds and signs the DMG with `--features parakeet,metal`
- [ ] `Release CLI Binaries` CI still passes on Windows + Linux (matrix `features` column defaults them to `parakeet` only)
- [ ] Maintainer confirms Metal is in the produced binary: `otool -L` on the shipped `minutes` CLI should show Metal.framework; running a transcription should log `whisper_init_with_params_no_state: use gpu = 1`.
- [ ] Smoke-test `large-v3` on a real meeting on an Apple Silicon machine and confirm wall time matches the benchmark (~2× faster than the pre-PR release).

## Notes for reviewer

- No `MACOSX_DEPLOYMENT_TARGET` bump is needed. The existing `11.0` target works because `dtolnay/rust-toolchain@stable` installs a rustup toolchain whose prebuilt `compiler_builtins` ships the `__isPlatformVersionAtLeast` compat shim that whisper.cpp's Metal ObjC code calls into. Local Homebrew-rustc builds on macOS 26 hosts are the only case that currently fails (Homebrew rebuilds `compiler_builtins` against the host SDK, losing the shim); that's a contributor-local Rust toolchain issue, not a CI issue.
- Happy to split the CoreML README clarification into a separate PR if preferred — left it here because anyone auditing the feature flags will hit the same \"silently falls back\" surprise I hit while benchmarking.

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -24,14 +24,17 @@ jobs:
             os: macos-latest
             binary: minutes
             asset: minutes-macos-arm64
+            features: parakeet,metal
           - target: x86_64-pc-windows-msvc
             os: windows-latest
             binary: minutes.exe
             asset: minutes-windows-x64.exe
+            features: parakeet
           - target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             binary: minutes
             asset: minutes-linux-x64
+            features: parakeet
 
     steps:
       - uses: actions/checkout@v4
@@ -59,7 +62,7 @@ jobs:
           key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Build release binary
-        run: cargo build --release -p minutes-cli --features parakeet --target ${{ matrix.target }}
+        run: cargo build --release -p minutes-cli --features ${{ matrix.features }} --target ${{ matrix.target }}
 
       - name: Rename binary
         shell: bash
--- a/.github/workflows/release-macos.yml
+++ b/.github/workflows/release-macos.yml
@@ -96,7 +96,7 @@ jobs:
           APPLE_API_KEY: ${{ secrets.APPLE_API_KEY }}
           TAURI_SIGNING_PRIVATE_KEY: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY }}
           TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ""
-        run: cargo tauri build --features parakeet --bundles app,dmg
+        run: cargo tauri build --features parakeet,metal --bundles app,dmg
 
       - name: Upload signed app bundle
         uses: actions/upload-artifact@v4
--- a/README.md
+++ b/README.md
@@ -739,28 +739,31 @@ If `minutes health` flags the mic as missing, the ChromeOS mic toggle is off —
 
 If Crostini support breaks for you, please [open an issue](https://github.com/silverstein/minutes/issues) — Chromebook isn't a first-class test target yet, so real bug reports are the fastest way to harden it.
 
-### GPU acceleration (optional)
+### GPU acceleration
 
-Build with GPU support for significantly faster transcription:
+macOS release binaries (DMG + `cargo install minutes-cli` from published CI
+artifacts) ship with Metal enabled — `large-v3` runs ~2× faster than the
+CPU-only build and offloads nearly all work to the GPU. Other backends remain
+opt-in at build time.
 
-| Backend | Platform | Feature flag | Prerequisites |
-|---------|----------|-------------|---------------|
-| Metal | macOS | `metal` | Xcode Command Line Tools |
-| CoreML | macOS | `coreml` | Xcode Command Line Tools |
-| CUDA | Windows/Linux | `cuda` | [CUDA Toolkit](https://developer.nvidia.com/cuda-toolkit) |
-| ROCm/HIP | Linux | `hipblas` | [ROCm](https://rocm.docs.amd.com/) 6.1+ (`hipcc`, `hipblas`, `rocblas`) |
-| Vulkan | Windows/Linux | `vulkan` | [Vulkan SDK](https://vulkan.lunarg.com/sdk/home) (+ `vulkan-headers` on Arch) |
+| Backend | Platform | Feature flag | Prerequisites | Default in release |
+|---------|----------|-------------|---------------|--------------------|
+| Metal | macOS | `metal` | Xcode Command Line Tools | **Yes** |
+| CoreML | macOS | `coreml` | Xcode Command Line Tools + `.mlmodelc` bundle | No |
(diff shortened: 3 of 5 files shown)


silverstein:
Merged. Big thank you for this one. It's a meaningful quality-of-life jump for every Mac user on the default Whisper path, and the kind of change I should have made months ago. Appreciated the rigor: byte-identical transcripts, user-CPU vs GPU teased apart, peak RSS tracked, and you tracing it back to the minutes-i8u9 step in the design doc made reviewing painless.

One FYI since you're clearly into benchmarking. There's a Parakeet backend behind `--features parakeet` (already compiled into our release builds, opt-in at runtime via `transcription.engine = "parakeet"` in config). Setup is `minutes setup --parakeet`. On a real 21-minute Apple Silicon run we measured about 10.7× realtime with fp16 on, versus the 6.89× you got from Whisper + Metal. Not apples-to-apples (tdt-600m FastConformer, 600M params, subprocess shellout rather than in-process, ~5 GB RSS), but might be fun to put through the same harness you built for this PR.

If you want to take the `minutes setup --coreml` auto-fetch follow-up, it's yours.

czlonkowski:
Thanks — merged while I was sleeping, much appreciated. Two quick replies:

**Parakeet benchmark.** Tried to put it through the same harness on my 8m40s Polish file for apples-to-apples. Blocked on this machine: `parakeet.cpp` via `axiom` needs Apple's `metal` shader compiler (`metallib`), which only ships with full Xcode — Command Line Tools don't include it, so the build dies at `[ 19%] Compiling Metal kernels to temporary library`. Not worth a 40 GB Xcode install just for a data point when your 21-min/10.7× number already makes the case; the 15–30% headroom over Whisper+Metal on a more permissive model is the interesting story regardless of audio. Happy to redo it on the same Polish file later if I end up installing Xcode for something else — especially curious about WER on Polish with `tdt-600m` vs `large-v3`, since Whisper's multilingual head is reliably better than most FastConformer checkpoints on Slavic languages and that's the part the benchmark table can't show.

Two small observations from setup that might be worth a follow-up issue regardless:
- `minutes setup --parakeet` prints a 5-step manual recipe (clone, hf download, convert_nemo.py, tar the tokenizer, cmake+install). That's a real adoption cliff compared to `minutes setup --model large-v3`'s one-shot download. A wrapper script or docs note about the Xcode-vs-CLT requirement would have saved me ~20 minutes.
- The `parakeet` binary isn't on Homebrew and `Frikallo/parakeet.cpp` has no GitHub releases. If that ever changes, `cmd_setup_parakeet` could point at a prebuilt.

**CoreML follow-up — taking it.** I'll send a PR for `minutes setup --model <name> --coreml` that fetches `ggml-<model>-encoder.mlmodelc.zip` next to the `.bin`, plus a warning in `whisper_context_params()` when `feature = "coreml"` is on but the bundle is missing (so the silent-CPU-fallback stops being a footgun). Should be a clean day's work.

silverstein:
Thanks for the detailed follow-up. Filed the two setup frictions as separate issues so they don't get buried in a merged PR thread:

- silverstein/minutes#162 for the `minutes setup --parakeet` automation + Xcode-vs-CLT docs note (Minutes-side)
- Frikallo/parakeet.cpp#21 for the upstream distribution story (releases + Homebrew), with your metallib experience as a concrete data point

On the parakeet.cpp side: I also shipped Frikallo/parakeet.cpp#19 a few days ago adding a warm-transcriber daemon as an example, and just filed Frikallo/parakeet.cpp#20 for a related LM-reload bug in the library. All of that is downstream of #21 actually landing, so pushing on distribution is probably the highest-leverage thread right now.

Looking forward to the CoreML PR whenever you have the day. The silent-CPU-fallback footgun in `whisper_context_params()` is a real one.

PR #147, opened 2026-04-17
RFC 0001: First-class templates for domain-shaped summaries

### LINKED ISSUE #143: Add --prompt and --prompt-file CLI arguments to customize summary prompt
## Summary

Currently, the summarization prompt is hardcoded in `summarize.rs`, requiring a full recompile to change it. This limits flexibility for users who want to tailor the output to specific contexts (language, format, domain).

## Proposed changes

Add two optional CLI arguments:

| Argument | Description |
|---|---|
| `--prompt <STRING>` | Provide a custom prompt inline |
| `--prompt-file <PATH>` | Path to a plain text file containing the prompt |

**Behavior:**
- If neither argument is provided, the current default prompt is used → **fully backward compatible**
- If both are supplied, `--prompt` takes precedence over `--prompt-file`

## Use cases

- Non-English summaries
- Domain-specific output formats (e.g. medical SOAP notes, meeting minutes templates)
- Experimenting with different summarization styles without recompiling

## Implementation notes

Changes would be limited to:
1. `cli.rs` — add both arguments to the `Args` struct via `clap`
2. `summarize.rs` — replace the hardcoded prompt with a parameter, falling back to the current default
3. `main.rs` — propagate the value through to the summarization call

Happy to submit a PR if the direction is approved.

silverstein:
Draft RFC opening the design conversation for **first-class templates** in Minutes. Surfaced from #143.

## TL;DR

Add templates as a first-class primitive. A template is a markdown file with YAML frontmatter that extends Minutes' structured extraction for a domain (medical, legal, standup, sales discovery, etc.). Templates are additive (preserve the baseline `KEY POINTS` / `DECISIONS` / etc. contract), compose with skills, support inheritance (family pattern), carry compliance rules for regulated verticals, and get PR'd into a community `templates/` directory in this repo.

## Why this shape

- **Resolves #143** as a side effect (domain output without recompile, without breaking structured extraction)
- **Community contribution path** for vertical domain experts (medical, legal, therapy) who will never work on Rust internals but can write markdown
- **Agent-native**: templates carry `agent_context` that informs Claude/Codex/Gemini/OpenCode when working with tagged meetings
- **Three-layer architecture** clarifies: capture (audio to transcript), schema (template-aware extraction), interaction (skills + MCP + graph). Templates slot cleanly into the middle layer.

## Reviewers

@ed0c, you're tagged as co-author on the RFC itself. The `medical-fr-base` + `consultation-fr` + `soap-fr` family worked examples are shaped directly by your HDS/RGPD context. Specifically wanted your review on:

- Whether the `extract:` nested structure matches how you'd actually author SOAP fields
- The family-pattern inheritance (my lean was option (c) from our #143 thread; the RFC shows what that looks like concretely)
- Any HDS or RGPD requirements that should be explicit fields in the `compliance:` schema
- Whether you'd rather own separate `consultation-fr` / `meeting-fr` / `soap-fr` templates, or a different grouping

## Open questions

Eight design questions at the bottom of the RFC that are the real areas where outside input changes the shape. Comments on any of them (or anything else) welcome.

## What merging this PR means

The RFC itself gets merged once the design has converged (expect ~2 weeks of feedback). Phase 1 implementation happens on a separate branch with its own PR.

Refs #143.


silverstein:
Update: I added a US SOAP family to the RFC as a second worked example. Pushed as commits 48eb8e9 and 7c65b25 on this branch. The thinking:

1. The RFC needs at least one fully-realized clinical family to pressure-test the design (inheritance, `extract:` shape, `compliance:` schema). The French `medical-fr` family is intentionally minimal pending review, but the RFC shouldn't be blocked on that.
2. SOAP has a widely used published structure (described in StatPearls, an NCBI Bookshelf clinical overview), so the family was drafted from published clinical references rather than from my own non-clinician interpretation, with explicit open questions for clinician review.
3. Two independent families (US + FR) deliberately avoid a shared `clinical-base` parent. US HIPAA obligations and French HDS/RGPD obligations don't share a clean common base yet, and forcing one would lock in lowest-common-denominator design before either family is mature.

The `soap` template ships **identified by default**. A clinical SOAP note is the working medical record for many clinicians, and stripping names, MRNs, or DOBs would defeat the template's purpose. The privacy posture is `require_local_processing: true` (audio and transcript stay on the user's machine) plus `audit_log: true` (local activity logging), and `forbid_in_summary` is a safety net for non-clinical high-risk identifiers (SSNs, credit card numbers, bank accounts, auth secrets) that should never surface in a clinical note regardless of intent. Research and teaching deployments that need de-identified output can override `redact_phi: true`.

Two rounds of adversarial review (Codex + a fresh reviewer) caught factual errors in v1: it over-claimed HIPAA's 6-year rule as audit-log retention (it actually applies to Security Rule documentation under 45 CFR 164.316(b)(2)(i)), miscounted MSE domains, mis-cut the Bright Futures growth-tracking thresholds, and shipped a privacy-by-default that didn't match the primary clinical workflow. v2 corrects these against StatPearls, APA Practice Guidelines, AAP Bright Futures, and HHS HIPAA references. Citations in the new subsection.

Also added Open Question 9 on HIPAA scope: `audit_log: true` and `require_local_processing: true` are tool-level safety behaviors, not deployment-level certifications.

Two things I most want clinical pushback on:

- The MSE in `psych-soap` follows APA Practice Guidelines / StatPearls domains, not the ASEPTIC mnemonic. ASEPTIC is a teaching aid, not a documentation standard, but some clinicians prefer it. Worth supporting both, or stay APA-aligned?
- Bright Futures anticipatory guidance is left as a free-text "visit-specific priorities" field rather than a fixed list, because each visit's priorities differ. A pediatrician using this daily might want the structure encoded per age band. Should the template ship age-banded fields, or stay generic?

ed0c:
Sorry, i've been busy these days because of my work. I'll watch this until the enf of the week

PR #153, opened 2026-04-19
feat(voice): enroll teammates by name, with optional persona label

severeon:
## Summary

Extends `minutes enroll` so the user can enroll voice profiles for teammates, not just themselves, and optionally tag each profile with a free-form persona label.

- `minutes enroll` — unchanged self-enrollment flow
- `minutes enroll --name "Alex Chen"` — enrolls a voice profile for someone other than you; does NOT write to `identity.name` in config
- `minutes enroll --name "Alex Chen" --persona "PM"` — same, plus an optional label stored alongside the profile
- `minutes enroll --name "Alex Chen" --persona "PM" --file ./alex.wav` — any combination works from a pre-recorded file

Persona is metadata only — it's surfaced in `minutes voices`, in the MCP `list_voices` tool, and in the `voices --json` output (`"persona": string | null`). It does **not** change the voice-matching algorithm. One profile per person (slug-keyed); multi-profile-per-person (Option B from the design) was explicitly deferred.

A slug-collision guard prevents a second `--name Alex` from silently blending two humans' embeddings into one profile — if the existing row has the same slug but a different `name`, enrollment refuses with a clear message pointing at `minutes voices --delete`.

## Changes

- **`crates/core/src/voice.rs`** — additive `persona TEXT` column via idempotent `ALTER TABLE`; `persona: Option<&str>` threaded through `save_profile` / `save_profile_blended`; new `lookup_name_for_slug` helper for collision detection; `slugify` exposed as `pub`.
- **`crates/cli/src/main.rs`** — new `--name` and `--persona` flags on `Enroll`; `cmd_enroll` branches on self vs other, collision guard in front of the save, distinct post-save messaging for self vs teammate; `cmd_voices` shows persona in brackets; two `save_profile_blended` call sites in `cmd_confirm_speaker` pass `None`.
- **`crates/mcp/src/index.ts`** — `list_voices` tool description mentions persona; human-readable output shows `name [persona]`; structured content unchanged (profiles flow through raw from `voices --json`, additive field).
- **`README.md`**, **`CLAUDE.md`** — enroll examples + module description updated.

Voices DB migration is additive and idempotent: existing rows read back `persona = null`. Verified on a real local `voices.db` with a pre-existing profile.

## Design notes

Option A from the brainstorm — persona as a single per-profile tag — was chosen over Option B (multiple profiles per person keyed on `(slug, persona)`). Option B would force a matcher policy decision (best-across-personas? surface which persona matched?) and was not required by the actual need. The `(person_slug, persona)` UNIQUE key can be added in a future migration without reshaping rows, so A stays forward-compatible.

## Test plan

- [x] `cargo test -p minutes-core --no-default-features --lib voice::` — 25/25 pass (5 new: round-trip, default-null, upsert-updates, migration-on-legacy-db, idempotency, lookup)
- [x] `cargo clippy --all --no-default-features -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cd crates/mcp && npm run build` — clean
- [x] `cd crates/mcp && npx vitest run` — 8/8 pass
- [x] `./target/debug/minutes enroll --help` — shows `--name` and `--persona`
- [x] `./target/debug/minutes voices --json` against real pre-existing `voices.db` — migration runs cleanly, old row reads back `"persona": null`
- [ ] Reviewer: smoke-test `minutes enroll --name "Test" --file <wav>` and confirm the persona surfaces in `minutes voices`
- [ ] Reviewer: try the collision guard: enroll `--name "Alex"`, then re-enroll `--name "Alex Chen"` (different slug, succeeds). Then try `--name "Alex"` again with a different profile — confirm refusal only when the slug is actually shared across different names

## Out of scope

- Version bump / release notes — deferred pending review + merge decision. This is additive user-visible functionality, so per the pre-commit checklist it warrants a SemVer minor bump (`0.13.1 → 0.14.0`) when it ships.
- Exposing `persona` in the MCP `confirm_speaker --save-voice` path — can be layered on without schema change.
- Multiple profiles per person (design Option B).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -312,13 +312,17 @@ minutes/
 ├── BUILD-STATUS.md            # Build progress tracker
 ├── Cargo.toml                 # Workspace root
 ├── crates/
-│   ├── core/src/              # 34 Rust modules — the engine
+│   ├── core/src/              # 40 Rust modules — the engine
 │   │   ├── capture.rs         # Audio capture (cpal), device categorization, loopback detection
 │   │   ├── resample.rs        # Shared mono-downmix + 16kHz decimation resampler (used by capture + streaming)
 │   │   ├── transcribe.rs      # Transcription: whisper.cpp (default) or parakeet.cpp (opt-in). Delegates to whisper-guard for anti-hallucination, optional nnnoiseless denoise
+│   │   ├── transcription_coordinator.rs # Engine dispatch (whisper vs parakeet); keeps pipeline engine-agnostic
+│   │   ├── parakeet.rs        # parakeet.cpp transcription path (gated behind `parakeet` feature)
+│   │   ├── parakeet_sidecar.rs # Parakeet sidecar process management (unix/non-unix split via cfg stubs — see Pre-Commit Checklist)
 │   │   ├── diarize.rs         # Speaker diarization + attribution types (pyannote-rs, or energy-based from per-source stems)
 │   │   ├── summarize.rs       # LLM summarization + speaker mapping (ureq HTTP client)
-│   │   ├── voice.rs           # Voice profile storage and matching (voices.db, enrollment, cosine similarity)
+│   │   ├── voice.rs           # Voice profile storage and matching (voices.db, enrollment self/teammate via `--name`, optional `--persona` label, cosine similarity)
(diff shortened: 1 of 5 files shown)


severeon:
Closing for local testing — will reopen once I've verified the end-to-end flow.

PR #160, opened 2026-04-21
fix(capture): prevent PipeWire segfault from double host init/teardown

gtheys:
## Summary

- **Fixes a segfault on `minutes record` with PipeWire backend (Linux)** by caching the `cpal::Host` for the process lifetime instead of creating and dropping it multiple times.

## Problem

`minutes record` crashes with a segmentation fault on Linux when using the PipeWire audio host:

```
#0  __strstr_sse2_unaligned ()
#1  load_spa_handle () at ../pipewire/src/pipewire/pipewire.c:235
#2  pw_load_spa_handle () at ../pipewire/pipewire/pipewire.c:304
#3  pw_loop_new () at ../pipewire/src/pipewire/loop.c:65
#4  pw_main_loop_new () at ../pipewire/src/pipewire/main-loop.c:64
```

### Root cause

cpal's `Host` struct on PipeWire holds a `PwInitGuard` that calls `pw_init()` on creation and `pw_deinit()` on drop. Minutes calls `cpal::default_host()` in at least two places in the recording path:

1. `resolve_capture_plan()` (preflight) — creates Host, enumerates devices, then **drops** Host → `pw_deinit()`
2. `record_to_wav()` — creates a new Host → `pw_init()` → `pw_main_loop_new()` → **segfault**

PipeWire cannot safely reinitialize in a process after `pw_deinit()` — internal state is corrupted, causing a NULL pointer passed to `strstr()` inside `load_spa_handle()`.

### Reproduction

A minimal reproducer confirmed this: calling `cpal::default_host()` twice and **dropping the first Host** before creating the second segfaults on PipeWire. Keeping the first Host alive (or only creating one) does not crash.

## Fix

Introduce `cached_default_host()` — a process-wide `OnceLock` that creates the `cpal::Host` once and **intentionally leaks it** via `Box::leak()`. This ensures the PipeWire init guard is never dropped, so `pw_deinit()` is never called between enumeration and stream creation.

All 8 `cpal::default_host()` call sites across `capture.rs`, `device_monitor.rs`, and `streaming.rs` now use `cached_default_host()`.

The intentional leak is negligible: one `Host` per process lifetime.

## Testing

- `minutes record --source "C922 Pro Stream Webcam" -v` — previously segfaulted, now records successfully
- `cargo test -p minutes-core --lib -- capture` — 38 tests pass
- `cargo build -p minutes-cli` — builds clean
- Tested with both PipeWire and ALSA backends

## Files changed

- `crates/core/src/capture.rs` — `HOST_CACHE` + `cached_default_host()` + 6 call sites updated
- `crates/core/src/device_monitor.rs` — 1 call site updated
- `crates/core/src/streaming.rs` — 1 call site updated

### DIFF
--- a/crates/core/src/capture.rs
+++ b/crates/core/src/capture.rs
@@ -469,7 +469,7 @@ fn select_device_with_override(
 }
 
 fn resolve_capture_plan(config: &Config) -> Result<CapturePlan, CaptureError> {
-    let host = cpal::default_host();
+    let host = cached_default_host();
     resolve_capture_plan_with_host(&host, config)
 }
 
@@ -1246,7 +1246,7 @@ pub fn record_to_wav(
     #[cfg(feature = "streaming")]
     crate::streaming::clear_mic_mute_for_new_recording();
 
-    let host = cpal::default_host();
+    let host = cached_default_host();
     let device_override = match &capture_plan {
         CapturePlan::Single(plan) => plan.device_override.as_deref(),
         #[cfg(feature = "streaming")]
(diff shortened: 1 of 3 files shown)


gtheys:
@silverstein Not sure how to contribute. I was running an older version on linux and it ran fine but the new version now introduced an issue. I fixed it and tested it. Works fine now.

PS: also your app worked on linux before ;)

silverstein (commented):
Nice diagnosis. The PipeWire `pw_init`/`pw_deinit` lifecycle issue is a real trap, and the reproducer (two `cpal::default_host()` calls with an intervening drop) nails the root cause. The fix is the right shape: make the `Host` outlive the process so `pw_deinit` never fires.

One tiny nit (non-blocking): `OnceLock<&'static cpal::Host>` + `Box::leak` is slightly redundant. Rust statics aren't dropped at program exit, so `OnceLock<cpal::Host>` would give you the same lifetime guarantee without the manual leak. That said, the explicit `Box::leak` documents intent loudly and the AIDEV-NOTE spells out the "why", which I like. Happy to leave it as-is.

Approving the CI workflow now. It'll run across macOS, Linux, and Windows. Once green, I'll merge through the GitHub flow.

Thanks for the thorough writeup and for testing both PipeWire and ALSA backends.

PR #192, opened 2026-04-28
fix(desktop): polish window persistence, processing dedupe, Recall + toast cleanup

silverstein:
## Summary

Six behavioral polish nits the user hits in normal use of the Tauri app, batched to avoid stair-stepping PRs over UI fixes.

- **Window size persists across launches.** Added `tauri-plugin-window-state`. Popup windows (note, meeting-prompt, dictation-overlay) opt out via `skip_initial_state` so they keep their fixed-size literals.
- **Processing UI no longer duplicates the same job in two places.** `#processing-jobs` is nested inside `#processing-bar`, so previously the singular row and per-job cards both rendered when one job was active. Now `.processing-bar.has-jobs .processing-row { display: none }` collapses to a single rendering path. The "You can keep recording" helper text moved out of the row so it survives in both states.
- **Recall panel no longer dims during recording.** The `.dimmed` toggle on phase=`'listening'` was backwards: Recall is exactly the surface the user is actively coaching with mid-meeting.
- **Recall flips to debrief the moment recording stops.** Both `!isProcessing` guards (expanded-panel branch and auto-expand branch) dropped, so the red "Listening" pulse no longer persists through the processing window.
- **Header swaps the "Minutes" wordmark for a 22x22 brand mark** sourced from the new app-icon Cathryn shipped in #191. Saves ~60px of horizontal space and reinforces the new brand. `flex-wrap` on `.header-actions` and `flex-shrink:0` on `.brand` as a layout safety net.
- **Completion toast simplified to one action.** Create Draft, Show in Folder, and Dismiss removed. Toast auto-dismisses after 12s with a backend clear (`cmd_clear_latest_output`) so status polling at `commands.rs:4506` does not resurrect it. Captures (`preserved-capture`) are exempt from auto-dismiss since they need attention.

Tray icons + favicon deferred to a separate PR. Cathryn's 1024x1024 design has an 84px red dot that becomes ~3.6px at 44x44, so the tray-state icons need their own design pass rather than a quick downscale.

## Files touched

- `tauri/src-tauri/Cargo.toml`: add `tauri-plugin-window-state = "2"`
- `tauri/src-tauri/src/main.rs`: register plugin, drop `.center()` on main window
- `tauri/src/index.html`: processing dedupe + Recall fixes + brand mark + toast simplification
- `tauri/src/icons/brand-mark.png`: 64x64 PNG sourced from `tauri/src-tauri/icons/app-icon.png`, rendered at 22x22 in CSS

## Test plan

- [ ] `./scripts/install-dev-app.sh` and launch `~/Applications/Minutes Dev.app`
- [ ] **Window persistence:** resize and move the main window, quit, reopen. Size + position restored.
- [ ] **Processing dedupe:** start a recording, stop it. During processing, only ONE rendering of the active job (either row OR card, never both). Helper text "You can keep recording while this finishes." stays visible.
- [ ] **Recall un-dimming:** open Recall, start a recording. Recall stays at full opacity during the listen phase.
- [ ] **Recall phase transition:** stop recording. "Listening" + red dot replaced by "Debrief" immediately, even though processing is still running.
- [ ] **Header layout:** resize the window down to ~420px width. Brand mark + chips do not overlap. Chips wrap onto a second row if necessary.
- [ ] **Completion toast:** finish a recording. Toast appears with only an "Open" button. Auto-dismisses after 12s. Click Open: toast dismisses + meeting opens. Path is not resurrected on the next status poll.
- [ ] **Capture toast:** trigger a capture-preserved scenario (mic disconnect mid-record or similar). Toast does NOT auto-dismiss (capture needs attention).
- [ ] `cargo fmt --all -- --check`
- [ ] `cargo clippy --all --no-default-features -- -D warnings` (clean locally)
- [ ] `cargo test -p minutes-core --no-default-features --lib` (525 passed locally)

Generated with [Claude Code](https://claude.com/claude-code)


silverstein:
Follow-up: addressed three issues from the review pass.

1. **Open handler now clears DOM, not just backend** (`tauri/src/index.html` Open click handler). Previously only invoked `cmd_clear_latest_output`, so the toast lingered visibly between the click and the next status poll. Now calls `updateOutputNotice(null)` immediately after the backend clear.

2. **Auto-dismiss gated on `kind === 'saved'`, not `!== 'preserved-capture'`.** `OutputNotice` has six kinds in `commands.rs`: `saved`, `preserved-capture`, `error`, `stale-recording`, `watch-failed`, `working-copy`. Five of those are attention-needed states. The previous `!isCapture` gate would auto-dismiss `error` / `stale-recording` / `watch-failed` / `working-copy` notices after 12s. Now only the happy `saved` path auto-dismisses.

3. **Dismiss button restored for non-saved notices.** Without auto-dismiss for attention-needed kinds, the user had no escape. Button hidden by default; shown only when `kind !== 'saved'`. Saved toast stays clean (Open + auto-dismiss), other kinds get an explicit Dismiss action.

Commit: 0612ddc.

PR #199, opened 2026-04-29
fix(desktop): compact call-detected card to single row

silverstein:
## Summary

When a call was detected, the banner expanded into a 6+ line card that took up most of the visible window. With Recall open the left pane is about 330px wide, which squeezed the text column to about 120px and forced everything to wrap.

After: single row matching the macOS-native call notification pattern.

```
  [icon]  Slack call detected               [Record]  X
```

## What got cut and why

The card had three elements that carried no actionable info:

- **Subtitle** ("Ready to record this call"): redundant with the title.
- **Paragraph** ("Minutes will stay ready while this call is active until you start recording or dismiss this session"): explained what the Record and Dismiss buttons do. Users know what those buttons do.
- **Status chips** ("Mic waiting" / "Call audio waiting"): `setCallHealthChip` was called once per detection with stale placeholder strings. Confirmed in the JS: the chips never updated past "waiting" during the detection phase. Real readiness signals fire in the recording-bar after Record is clicked, where the same chip variants update with `mic_live` / `call_audio_live` state.

## Implementation

- **CSS**: `.call-detected` changes from `align-items: flex-start` with column-stacked info to `align-items: center` with the title `flex:1` (text-overflow: ellipsis on overflow). Padding tightens.
- **HTML**: drops `.call-detected-info` wrapper, subtitle, paragraph, health chips. Adds `.call-detected-close` X button.
- **JS**: drops references to `callSubtitle`, `callDetail`, `callHealthMic`, `callHealthAudio`. Keeps `setCallHealthChip` because the recording-bar still uses it (index.html:5806-5807) for live mic/call-audio status during recording.

## Out of scope

This PR is UI-only. There's a separate logic bug where starting a Google Meet sometimes mis-identifies as "Slack call detected" (because Slack is running in the background and the native-app match at `call_detect.rs:681-702` fires before the browser probe gets to run). That's tracked in a separate issue and needs more careful work plus a codex review on the proposed fix.

## Test plan

- [ ] `./scripts/install-dev-app.sh` (rebuilt on commit)
- [ ] Trigger a call (Zoom, Slack, Meet, Teams)
- [ ] Verify single-row banner: icon + "{App} call detected" + Record button + X dismiss
- [ ] Click Record: starts recording (no regression)
- [ ] Click X: dismisses banner (no regression)
- [ ] Resize window narrow: title truncates with ellipsis, doesn't wrap
- [ ] Open Recall while banner is shown: layout still fits

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tauri/src/index.html
+++ b/tauri/src/index.html
@@ -3219,12 +3219,17 @@
       flex-wrap: wrap;
     }
 
-    /* ── Call detected banner ── */
+    /* ── Call detected banner ──
+       Single-row notification (matches macOS-native call notification pattern):
+       icon + title + Record CTA + corner ✕. Previously had a subtitle, an
+       instructional paragraph, and two readiness chips that never updated past
+       "waiting" — all dropped because they bloated the card to 6+ lines and
+       carried no actionable info. */
     .call-detected {
       display: none;
-      align-items: flex-start;
+      align-items: center;
       gap: 10px;
-      padding: 12px 16px;
+      padding: 10px 12px 10px 16px;
       background: var(--green-tint-soft);
       border-bottom: 1px solid var(--green-border);
     }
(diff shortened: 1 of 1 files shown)


silverstein:
Related logic bug tracked separately: #200

Issue #229, opened 2026-05-11
SIGABRT on Stop when post-stop transcription fails to spawn whisper context; orphan WAV in jobs/ not retried


silverstein:
## Summary

When a recording ends and the post-stop processing path tries to spawn a whisper context to transcribe the resulting WAV, a context-spawn failure path triggers a Rust `Drop` during process exit that calls `abort()`, resulting in a hard `SIGABRT` crash instead of a graceful "transcription failed" error to the UI. The WAV stays in `jobs/` and is not retried on app relaunch.

Reported by @allaire on #227 with a full macOS crash report (https://github.com/silverstein/minutes/issues/227#issuecomment-4422948921), and now with diagnostic data on #229 (comment 4423605694) that pins down the trigger.

## Root cause for Bug A (confirmed via @allaire's diagnostic data, 2026-05-11)

**Bug A is partial / corrupt model files on disk that `minutes setup` does not detect.**

Allaire's bug surfaces only with `transcription.model = "medium"` or `"large"`. `small` works. His `~/.minutes/models/ggml-medium.bin` is 221 MB. Canonical `ggml-medium.bin` is roughly 1.5 GB, so his file is about one seventh of a complete model, almost certainly from a download that got interrupted (network, app kill, disk full).

`minutes setup --model medium` reports "Model already downloaded" because it only checks file existence, not size or integrity. `whisper-rs` then tries to parse a truncated GGML header on load, fails, and the in-process failure cascades into Bug B's Drop panic.

**Fix scope for Bug A:**

- `minutes setup` validates file size against expected per-model size after download, and on subsequent invocations. If the file exists but is truncated, treat as "needs download" not "already downloaded".
- Model load path (`crates/core/src/transcribe.rs` `resolve_model_path` and downstream `WhisperContext::new_with_params` call) checks file size against expected before passing to whisper-rs. If undersized, return a typed `ModelLoadError::Truncated` with a hint to re-download.
- UI surface: a corrupt-model error should offer a "re-download model" button in the Readiness Center, not just a generic error toast.
- A future enhancement (out of scope for v0.17.2): publish per-model SHA256 in a manifest fetched alongside the binary, validate via checksum not just size.

### A. Post-stop whisper context spawn failure (ROOT CAUSE: partial model file)

[...]

PR #230, opened 2026-05-11
fix(worker): harden post-stop transcription against SIGABRT + bounded retry (#229)

silverstein:
## Summary

Issue #229: a post-stop transcription failure crashed the worker with SIGABRT inside `__cxa_finalize_ranges -> exit`, then left an orphan WAV in `~/.minutes/jobs/` that no automatic recovery would handle. This PR addresses Bug B (the crash) and Bug C (the missing retry policy) from that triage. Bug A (the upstream root cause: a truncated `ggml-medium.bin` on disk) will land as a follow-up commit on this branch.

### Bug B: SIGABRT on worker exit

Two complementary defenses:

1. **Worker subprocess exits via `_exit` on macOS.** The Tauri `--process-queue-worker` path used to call `std::process::exit(code)`, which runs C++ static destructors during `__cxa_finalize_ranges`. whisper.cpp / ggml register static state whose destructors can call `abort()` on partially-initialized contexts. The worker now drains the parakeet sidecar manager and calls `exit_process_without_destructors`, skipping that teardown entirely.
2. **`process_pending_jobs` wraps each transcription pass in `catch_unwind`.** A Rust panic from any of the heavy native paths (whisper FFI, parakeet helper, audio decode, LLM summarizer) is converted to `MinutesError::Transcribe(TranscriptionFailed(...))` and the job is marked `Failed` via the existing Err path, instead of unwinding through the worker.

Audited every `impl Drop` in `crates/core/src/`. None of them panic. The abort was C++ static teardown, not a Rust-side destructor.

### Bug C: bounded retry instead of infinite loop

`ProcessingJob` gains a `retry_count: u32` field (with `#[serde(default)]` so on-disk JSON stays compatible). `list_jobs()` now caps automatic stale-claim recovery at `MAX_AUTO_RETRIES = 2`. Past the cap, the job is demoted to `Failed` with an explanatory error message and moved across the active/archive boundary via `update_job_state`. A user-initiated `requeue_job` resets `retry_count` to 0 so explicit manual retry always gets a fresh budget.

The recovery loop now delegates to `update_job_state` per candidate ID (instead of mutating a `list_jobs_raw()` snapshot in memory) so a concurrent worker that re-claims a job between scan and write does not get silently clobbered. `latest_retryable_output_notice()` switches from `list_jobs()` to `display_jobs(None, true)` so retry-cap demotions remain visible on the startup recovery notice path. Both fixes came out of the codex adversarial review rounds.

## Test plan

- [x] `cargo test -p minutes-core --no-default-features --lib` (699 pass, 1 ignored)
- [x] `cargo test -p minutes-app` (188 pass)
- [x] New `list_jobs_demotes_to_failed_when_retry_cap_exceeded` confirms the active to archive transition + user-facing error string
- [x] New `manual_requeue_resets_retry_count` covers the explicit fresh-budget reset
- [x] `list_jobs_recovers_stale_worker_claims` extended to assert retry_count increments
- [ ] Manual: simulate a model-load failure (point config at a missing model) and confirm the worker exits cleanly instead of aborting (target for v0.17.2 dogfood)

Refs #229.

Bug A follow-up will land on this branch before v0.17.2 ships.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -4866,14 +4866,46 @@ fn cmd_setup(model: &str, list: bool, diarization: bool) -> Result<()> {
     std::fs::create_dir_all(model_dir)?;
 
     let dest = model_dir.join(format!("ggml-{}.bin", model));
-    if dest.exists() {
-        let size = std::fs::metadata(&dest)?.len();
-        eprintln!(
-            "Model already downloaded: {} ({:.0} MB)",
-            dest.display(),
-            size as f64 / 1_048_576.0
-        );
+    let expected_min_bytes = minutes_core::transcribe::expected_whisper_model_size_bytes(model);
+    let mb = |bytes: u64| bytes as f64 / 1_048_576.0;
+
+    // Helper: treat an existing file as truncated if it's smaller than the
+    // expected minimum for this model (issue #229 root cause: a partial
+    // download was reported as "already downloaded" and whisper-rs aborted
+    // parsing the truncated GGML header). Returns Ok(true) when the
+    // existing file should be kept, Ok(false) when it was removed and a
+    // re-download is needed.
+    let keep_existing = if dest.exists() {
+        let actual = std::fs::metadata(&dest)?.len();
+        match expected_min_bytes {
+            Some(min_bytes) if actual < min_bytes => {
+                eprintln!(
+                    "Model file at {} is {:.0} MB but the {} model should be at least {:.0} MB.",
+                    dest.display(),
+                    mb(actual),
+                    model,
+                    mb(min_bytes),
+                );
+                eprintln!(
+                    "Looks truncated, probably an interrupted download. Removing and refetching."
+                );
+                std::fs::remove_file(&dest)?;
+                false
+            }
+            _ => {
+                eprintln!(
+                    "Model already downloaded: {} ({:.0} MB)",
+                    dest.display(),
+                    mb(actual),
+                );
+                true
+            }
+        }
     } else {
+        false
+    };
+
+    if !keep_existing {
         let url = format!(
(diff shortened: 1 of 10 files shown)


silverstein:
Update: Bug A (truncated whisper model files silently accepted) is now folded into this PR after allaire posted diagnostic data on #229 pinning the root cause.

Three new commits on this branch since the original PR:

- `0a74756` fix(setup): detect truncated whisper model files (#229 Bug A)
- `b5df908` chore(site): bump test count after #229 Bug A model-size tests
- `9a32db0` chore(docs): regenerate agent docs for ModelTruncated error variant

What Bug A adds:

- `expected_whisper_model_size_bytes(name)` table sourced from real `Content-Length` values on `huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-{name}.bin` (with ~5% headroom).
- `validate_whisper_model_size(path)` wired into every `resolve_model_path*` helper so dictation, live transcript, and batch all size-check before whisper-rs ever sees the file.
- Typed `TranscribeError::ModelTruncated` carrying a user-facing fix hint (`rm "{path}" && minutes setup --model {name}`).
- `minutes setup` re-validates existing files (deletes partials and re-downloads) and bails after a successful HTTP that still ends up undersize.
- Four new tests covering the contract, allaire's truncation scenario, the happy path, and the custom-filename bypass.

Codex adversarial review caught two race conditions in `list_jobs()` during round 1 (already fixed before the original PR) and flagged the new active-to-archive demotion as silently invisible to `latest_retryable_output_notice()` in round 2 (fixed in commit `62bbcc4`). Round 3 challenged the model-size table; I verified the actual canonical sizes via real `Content-Length` HEAD requests and pinned the source-of-truth comment into the code.

All three bugs from #229 are now addressed and ready to bundle into v0.17.2. Tests: 700 core lib + 4 new whisper-gated + 188 Tauri = clean. clippy + fmt clean.

Issue #224, opened 2026-05-11
Love the idea, but every recording is showing as unintelligible


jacobSingh:
Never had a problem with Granola, and use local whisper models regularly, but this one shows as gibberish through most of the conversation. 

Using OSX, external mic, internal speaker, discord call.

Separate issue: 
It decides you're in a meeting and steals focus every time I speak. Since I use superwhisper all day, this doesn't work obviously.

silverstein:
Hey, thanks for trying it and for the detailed report. Two separate things going on, let me take them one at a time.

**On the gibberish:**

The default model is `small` (not `tiny`), so size is probably not the root cause. When everything comes out unintelligible the issue is almost always the audio reaching whisper, not the model.

The most likely culprit in your setup is **external mic + internal speaker + Discord**. The mic picks up your voice, the Discord audio coming out of the speaker, and room echo, all overlapping. Whisper falls apart on heavily mixed/overlapping speech. Minutes currently records mic only by default (no separate system-audio leg), so the far side of the call only reaches the transcript through speaker bleed, which is the worst possible path.

To narrow it down, could you share:

1. A short snippet of the actual gibberish output. The text itself tells me a lot, whether it is whisper hallucination loops, silence being decoded, or something else.
2. Output of \`minutes devices\` and which input was selected for the recording.
3. Whether you were on headphones or speakers.
4. Your config if you have one: \`cat ~/.config/minutes/config.toml\` (or \"none\" if you never created one).

As a quick test before anything else: try the same kind of call **with headphones** instead of internal speakers. If the transcript becomes legible, the issue is acoustic bleed and we can talk about configuring a system-audio source so Discord's far-side audio is captured cleanly instead of going through the speaker-to-mic loop.

**On the focus stealing:**

That should not be happening with Discord by default. The detector only fires on Zoom, Teams, and Webex out of the box, and it only steals focus once per detected call (reminders explicitly skip the focus grab). \"Every time I speak\" suggests something is repeatedly re-triggering. Two questions:

1. Did you add Discord (or anything else) to \`call_detection.apps\` in config?
2. Can you grab any \`call:detected\` lines from \`~/.minutes/logs/minutes.log\` while reproducing?

Quick mitigation if you want it gone immediately: add this to \`~/.config/minutes/config.toml\`:

\`\`\`toml
[call_detection]
enabled = false
\`\`\`

That kills the auto-detect entirely so it can't fight with Superwhisper. We can dig into the root cause separately.

[...]

Issue #216, opened 2026-05-07
Native call capture: long recordings SIGKILLed during stopCapture finalize, .mov is truncated     (no moov atom); WAV stems survive


sjunnesson:
## Summary
                                                                                                        
  On macOS, native call capture truncates the output `.mov` whenever the recording is more than ~1 hour 
  long. The recorder helper (`system_audio_record`) is sent SIGTERM, but the parent
  (`call_capture.rs::CallCaptureSession::stop`) gives it only 15 seconds to exit before issuing SIGKILL.
   For long captures, `try await stream.stopCapture()` in the Swift helper takes longer than that to
  finalize the `SCRecordingOutput` (writing the moov atom for tens-to-hundreds of thousands of media
  samples). SIGKILL can't be caught → moov never written → file ends in
  `…/areas/meetings/failed-captures/<ts>-capture.mov` with `[mov,mp4,m4a,3gp,3g2,mj2 @ …] moov atom not
  found`.

  The parallel PCM stems written by `writeStemSamples` (`<base>.voice.wav` / `<base>.system.wav` in     
  `~/.minutes/native-captures/`) **do** survive — they're finalized synchronously inside
  `sampleQueue.sync { ... }` *before* `stopCapture()` runs.                                             
                                          
  ## Environment

  - macOS 26.4.1 (build 25E253), Apple Silicon                                                          
  - Minutes 0.16.2 (desktop app, native call capture path)
  - `stop_when_call_ends = true` (also reproduces on manual stop of long sessions)                      
                                                                                                        
  ## Repro
                                                                                                        
  1. Start a Teams/Zoom call, accept the call-detect banner to begin recording.                         
  2. Let the call run for at least ~1 hour.
  3. Hang up — auto-stop countdown fires.                                                               
  4. Observe a new file under `<output_dir>/failed-captures/<TS>-capture.mov` and a "Recovery" panel    
  entry. The .mov has `ftyp` + `wide` + `mdat` headers but no `moov`.                                   
                                                                                                        
  I have hit this 4 times in 2 days on captures of 1 h 8 m, 1 h 19 m, 1 h 38 m, and 3 h 02 m. A 7 s     
  capture and a 0.128 s capture from the same recorder build finalized cleanly, so the failure
  correlates with duration, not setup.                                                                  
                                          
  ## Smoking gun

  Log signature on every failure (here: today's 1 h 19 m capture):                                      
   
  ```                                                                                                   
[...]

Issue #231, opened 2026-05-11
v0.17.2 audio probe rejects valid .m4a with moov-after-mdat (iPhone Voice Memos)


jmh1313:
## Summary
Minutes v0.17.2's audio probe rejects valid `.m4a` files whose `moov` atom follows `mdat` (the default iPhone Voice Memos export layout). The same files were accepted under v0.14.0.

## Reproduction
1. Export a recording from iPhone Voice Memos, or use any `.m4a` with `ftyp` major brand `M4A ` and atom order `ftyp → mdat → moov`.
2. Drop into the watched inbox path while running `minutes watch` v0.17.2.

## Expected
File is probed as a valid audio container and enters the transcription pipeline.

## Actual
Watcher logs:
```
INFO stream is seekable with len=<N> bytes
WARN file failed audio probe — not a valid audio container, skipping
```
`ffprobe` accepts the same files cleanly: `probe_score=100`, full stream metadata, valid `mp4a.40.2` AAC, correct Voice Memos `creation_time` and `voice-memo-uuid` tags.

## Verified samples
| Size | Duration | Title | Encoder |
|---|---|---|---|
| 21 MB | 42:05 | "Jake Ursula 8/4" | iPhone Voice Memos 18.5 |
| 6.8 MB | 13:00 | "Pam Jake 10/10" | iPhone Voice Memos 18.6.2 |
| (historical) | — | (processed cleanly under v0.14.0) | iPhone Voice Memos |

The historical sample has the identical atom layout (`ftyp → mdat → moov`) and identical major brand (`M4A `, space-padded) — so the rejection is a probe-behavior change between v0.14.0 and v0.17.x, not a file-content issue.

## Workaround
Container-only rewrite places `moov` first:
```bash
ffmpeg -i in.m4a -c copy -movflags +faststart out.m4a
```
v0.17.2 then accepts and processes the rewritten file normally. Audio data is byte-identical; only the atom order changes.

## Suspected cause
Audio probe library was changed or tightened between v0.14.0 and v0.17.x. Possible angles:
- New probe requires `moov` before `mdat` (streaming-friendly layout), which Apple Voice Memos does not produce by default.
- Major brand `M4A ` (space-padded, Apple's audio-only brand) no longer recognized in addition to `isom`/`mp42` in compatible_brands.

## Environment
- macOS 25.2.0 (Darwin)
- Minutes v0.17.2 (Homebrew tap `silverstein/tap`)
[...]

PR #235, opened 2026-05-12
fix(pipeline): mix WAV stems for transcription to avoid .mov dual-track 2x bug

### LINKED ISSUE #234: native-call capture: .mov with dual AAC audio tracks doubles Whisper input duration to 2.00x on macOS 26
## Summary

On macOS, `native-call` capture produces a `.mov` containing **two independent AAC audio tracks** (system stem and voice stem as separate streams). `pipeline.rs::process_with_progress_and_sidecar` (~line 1601 in v0.17.2) feeds this `.mov` directly to the transcription coordinator. The audio decoder reads the two streams sequentially, doubling the effective audio length sent to Whisper. Output is exactly **2.00x** the real recording duration, with overlapping/garbled content (i.e., "unintelligible").

This is plausibly the same root cause behind #224, where the user is on macOS and the symptom matches exactly. The paired `.voice.wav` / `.system.wav` PCM stems written by `writeStemSamples` are clean and the correct length, same stem-survival path #216 already relies on.

## Repro

`ffprobe -show_streams` on a representative native-call `.mov` from `~/.minutes/native-captures/` (macOS 26.2, Apple Silicon, minutes 0.17.2):

```
Stream 0: aac, audio, 48000 Hz, 1 channel, mono, duration 4419.080000s
Stream 1: aac, audio, 48000 Hz, 1 channel, mono, duration 4419.040000s
```

Two AAC mono streams, each ~4419 seconds (74 min) of real recording.

When this `.mov` is fed to the transcription step, the pipeline log's `step: transcribe` line reports an `audio` duration of approximately 8838s, exactly 2.00x the real source length, with proportionally inflated word counts and segment counts. After processing, the resulting transcript is largely incoherent overlapping speech.

**Confirmed across 7 historical native-call captures** in `~/.minutes/native-captures/`. Every one has two AAC audio streams of near-identical duration, and every one transcribed at exactly 2.00x with mangled content. The pre-existing `.voice.wav` / `.system.wav` stems for each capture are single-stream, correct length, and clean.

## Environment

- macOS 26.2 (build 25C56), Apple Silicon
- minutes 0.17.2 (Homebrew bottle, `silverstein/tap`)
- ScreenCaptureKit native-call backend (`capture_backend = "native-call"` in `minutes.log`)
- ffmpeg 8.1 (used downstream for the mix; not the producer of the `.mov`)
- `config.toml`: `transcription.model = "small"`, `noise_reduction = true`, otherwise defaults

## Root cause

`SCRecordingOutput` in the Swift capture helper writes a `.mov` with system audio and mic audio as **two separate AAC tracks** rather than a pre-mixed single track. This is the natural ScreenCaptureKit output on macOS 26; older OS versions or different capture options may differ. The `.mov` is then handed to the transcription coordinator at `pipeline.rs::process_with_progress_and_sidecar` (the `audio_path` arg, which becomes `transcribe_input`). symphonia/whisper consumes both audio streams sequentially, so the input length doubles.

## Proposed fix

Same conceptual frame as #216: stems are reliable, the `.mov` is fragile. Sniff `audio_path` at the top of the transcription step; if it's a `.mov` with sibling `.voice.wav` / `.system.wav` stems, mix the stems via `ffmpeg amix` to a single PCM and transcribe that. Fall back to the original `.mov` decode if mixing fails or stems are missing.

Patched locally as a free function `prepare_transcription_input(audio_path) -> Option<PathBuf>` (~80 lines) plus a one-line call site at the existing transcribe input lookup. After applying the patch, an inventory meeting that previously transcribed as gibberish came back as **11,139 coherent words** at the correct duration, accurate to the conversation. Reprocessed 6 of 7 historical broken meetings the same way; all recovered to correct duration and intelligible content.

PR follows shortly with the patch.

## Notes

- Diarization was already correct on these captures (it consumes the WAV stems directly per `diarization_from_stems: true` in the log). Only the transcription step was hitting the `.mov`.
- The `file` field in the pipeline log preserves the original `.mov` path even when the mixed stem is what's actually decoded; useful for downstream provenance.
- ffmpeg is already a runtime requirement elsewhere in the toolchain (e.g., `minutes-remux`), so the mix step doesn't introduce a new dependency.

Filed as a separate issue from #224 because the user-visible symptom there ("gibberish") matches but the existing hypothesis (acoustic bleed) is different. Cross-linked from a comment on that thread.

jmh1313:
Fixes #234. Related to #224.

## What this changes

When the transcription step is handed a `.mov` (the format ScreenCaptureKit writes for native-call captures), check for sibling `.voice.wav` / `.system.wav` PCM stems beside it. If both stems are present, mix them to a single 16kHz mono PCM via `ffmpeg amix` and transcribe the PCM. Fall back to direct `.mov` decode if mixing fails or stems are missing, so existing behavior on other platforms and non-native-call captures is preserved.

## Why

On macOS 26 with the `native-call` capture path, `SCRecordingOutput` writes a `.mov` with two independent AAC audio tracks (system + voice). Feeding that `.mov` straight to the transcription coordinator caused the audio decoder to read the two streams sequentially, doubling the effective input length sent to Whisper. Output was exactly 2.00x source duration with overlapping/garbled content (i.e., "unintelligible").

This is the same stem-survival frame used to recover from #216: the WAV stems are written synchronously inside `sampleQueue.sync { ... }` and are pristine even when the `.mov` is fragile or wrong. Diarization already consumes the stems directly (`diarization_from_stems: true`); this change brings transcription into the same approach.

## Implementation

- New free function `prepare_transcription_input(audio_path: &Path) -> Option<PathBuf>` in `pipeline.rs`.
  - Returns `None` (caller falls through to existing decode path) if: not a `.mov`, either stem missing, or ffmpeg fails for any reason.
  - On success, returns a temp PCM path the caller should delete after use.
- One-line wiring at the existing `transcribe_input` lookup in `process_with_progress_and_sidecar`.
- Tempfile uses `std::process::id()` + stem name; restricted to mode 0o600 on Unix.
- ffmpeg is already a runtime requirement (`minutes-remux`, etc.), so this doesn't add a new dependency.

## Verification

- 6 of 7 historical native-call captures recovered to correct duration and intelligible transcripts after applying the patch.
- One specific case: a 74-minute inventory meeting went from gibberish to 11,139 coherent words.
- Pipeline log's `file` field still shows the original `.mov` path, preserving downstream provenance.

## Notes / open questions

- Happy to add a unit test that constructs a fake `.mov` + stem pair and asserts the function returns the temp PCM path. Wasn't sure of the test conventions for this crate (couldn't find an obvious pipeline-level test fixture); will follow whatever direction you prefer.
- Could also gate the workaround behind a config flag if you'd rather not change default behavior. Defaulted to "on" because the .mov path is currently broken in the affected configuration and there's no user-visible downside to using the stems.
- The fallback `tracing::warn!` on mix failure preserves the bug-mode output but at least surfaces a log line for diagnosis.

### DIFF
--- a/crates/core/src/diarize.rs
+++ b/crates/core/src/diarize.rs
@@ -364,13 +364,13 @@ pub struct StemPaths {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-enum SourceAwareDiarizationPlan {
+pub(crate) enum SourceAwareDiarizationPlan {
     FullStems(StemPaths),
     SystemStemOnly(std::path::PathBuf),
     SilentSystemStem(StemPaths),
 }
 
-fn stem_has_audio(path: &Path) -> bool {
+pub(crate) fn stem_has_audio(path: &Path) -> bool {
     let Ok(reader) = hound::WavReader::open(path) else {
         return false;
     };
@@ -557,7 +557,7 @@ fn probe_stem_observed_signal<T>(
     }
 }
 
-fn discover_stem_plan(audio_path: &Path) -> Option<SourceAwareDiarizationPlan> {
+pub(crate) fn discover_stem_plan(audio_path: &Path) -> Option<SourceAwareDiarizationPlan> {
     let stem = audio_path.file_stem()?.to_str()?;
     let dir = audio_path.parent()?;
     let voice = dir.join(format!("{}.voice.wav", stem));
--- a/crates/core/src/error.rs
+++ b/crates/core/src/error.rs
@@ -83,6 +83,14 @@ pub enum TranscribeError {
     #[error("parakeet transcription failed: {0}")]
     ParakeetFailed(String),
 
+    #[error(
+        "native call capture cannot be transcribed: {reason}. \
+         The .mov produced by macOS SCRecordingOutput decodes to ~2x source duration on this \
+         container shape; transcription requires the sibling .voice.wav and .system.wav PCM \
+         stems to be mixed via ffmpeg first."
+    )]
+    NativeCaptureStemMixUnavailable { reason: String },
+
     #[error("I/O error: {0}")]
     Io(#[from] std::io::Error),
 }
--- a/crates/core/src/pipeline.rs
+++ b/crates/core/src/pipeline.rs
@@ -650,9 +650,41 @@ fn prepare_transcription_input(
             );
         }
         None => {
-            // No stems at all. Could be a non-native-call `.mov` (screen
-            // recording, downloaded file) or a native-call capture whose
-            // stems were cleaned up; we cannot distinguish. Conservative:
+            // `discover_stem_plan` returns None for two semantically different
+            // cases: (a) neither stem present (ordinary non-native-call `.mov`
+            // or a native capture whose stems were cleaned up) and (b) voice
+            // stem present and audible but system stem file is entirely
+            // absent from disk (`(true, false) && !system.exists()` branch of
+            // discover_stem_plan at `diarize.rs:600-606`). Case (b) is a
+            // native capture where the system side was lost during recording,
+            // and falling through to the broken `.mov` decoder reproduces the
+            // exact 2x-duration bug this helper exists to prevent. Codex
+            // review of PR #235 v2 caught this.
+            //
+            // Distinguish by independently checking for a usable sibling
+            // voice stem. If one exists, surface the same typed error as
+            // the other should-have-mixed-but-couldn't branches.
+            if let Some(parent) = canonical.parent() {
+                if let Some(stem_name) = canonical.file_stem().and_then(|s| s.to_str()) {
+                    let voice = parent.join(format!("{}.voice.wav", stem_name));
+                    if voice.exists() && crate::diarize::stem_has_audio(&voice) {
+                        return Err(
+                            crate::error::TranscribeError::NativeCaptureStemMixUnavailable {
+                                reason: format!(
+                                    "voice stem present at {} but system stem is missing from disk. \
+                                     Partial-crash signature; mix would substitute silence for the \
+                                     far-side audio.",
+                                    voice.display()
+                                ),
+                            }
+                            .into(),
+                        );
+                    }
+                }
+            }
+            // No usable stems at all. Could be a non-native-call `.mov`
+            // (screen recording, downloaded file) or a native-call capture
+            // whose stems were cleaned up; we cannot distinguish. Conservative:
             // treat as ordinary `.mov` and let the existing decoder handle
             // it. Hard-erroring on every stemless `.mov` would break
             // legitimate non-native-call use cases.
@@ -6540,6 +6572,47 @@ mod prepare_transcription_input_tests {
                 assert!(
                     reason.to_lowercase().contains("voice stem"),
                     "error reason must mention the missing voice stem; got: {}",
+                    reason
+                );
+            }
+            other => panic!(
+                "expected NativeCaptureStemMixUnavailable, got: {:?}",
+                other.map(|opt| opt.is_some())
+            ),
+        }
+    }
+
+    #[test]
+    fn errors_when_voice_present_and_system_stem_file_absent() {
+        // Codex review of PR #235 v2 caught this: `discover_stem_plan`
+        // returns None for both the "no stems at all" case AND the
+        // "voice ok, system absent from disk" case. The second case is
+        // a partial-crash native capture where the system stem was lost
+        // during recording, and falling through to the broken `.mov`
+        // decoder reproduces the exact 2x bug this helper prevents.
+        //
+        // The fix distinguishes the two None cases by independently
+        // checking for a usable sibling voice stem in
+        // `prepare_transcription_input`. This test pins that contract.
+        let dir = tempfile::TempDir::new().unwrap();
+        let mov = dir.path().join("partial-system-missing.mov");
+        let voice = dir.path().join("partial-system-missing.voice.wav");
+        fs::write(&mov, b"x").unwrap();
+        write_audible_wav(&voice);
+        // system.wav deliberately not created (not even zero-byte; the
+        // file doesn't exist on disk at all). This is the case that
+        // returned None from discover_stem_plan and would have silently
+        // fallen through to the broken `.mov` decode without this fix.
+
+        let result = prepare_transcription_input(&mov);
+        match result {
+            Err(MinutesError::Transcribe(
+                crate::error::TranscribeError::NativeCaptureStemMixUnavailable { reason },
+            )) => {
+                assert!(
+                    reason.to_lowercase().contains("system stem")
+                        && reason.to_lowercase().contains("missing"),
+                    "error reason must call out the missing system stem; got: {}",
                     reason
                 );
             }
--- a/crates/core/src/pipeline.rs
+++ b/crates/core/src/pipeline.rs
@@ -533,6 +533,85 @@ fn expected_voice_stem_path(audio_path: &Path) -> Option<std::path::PathBuf> {
     Some(dir.join(format!("{}.voice.wav", stem)))
 }
 
+/// Workaround for the .mov dual-track decode bug.
+///
+/// macOS SCRecordingOutput writes a `.mov` with two audio tracks (system + mic)
+/// plus pristine `.voice.wav` and `.system.wav` stems beside it. Decoding the
+/// `.mov` for transcription produces audio at 2x real duration, causing whisper
+/// to receive garbled samples and emit gibberish.
+///
+/// When stems are present, mix them into a 16kHz mono PCM wav via ffmpeg and
+/// hand that to the transcriber instead. Returns the temp file path on success;
+/// caller must remove it after use. Falls back silently (returns None) on any
+/// error so the original `.mov` decode path stays as a safety net.
+fn prepare_transcription_input(audio_path: &Path) -> Option<std::path::PathBuf> {
+    let ext = audio_path
+        .extension()
+        .and_then(|e| e.to_str())
+        .map(|s| s.to_lowercase());
+    if ext.as_deref() != Some("mov") {
+        return None;
+    }
+    let stem = audio_path.file_stem()?.to_str()?;
+    let dir = audio_path.parent()?;
+    let voice = dir.join(format!("{}.voice.wav", stem));
+    let system = dir.join(format!("{}.system.wav", stem));
+    if !voice.exists() || !system.exists() {
+        return None;
+    }
+
+    let tmp = std::env::temp_dir().join(format!(
+        "minutes-stem-mix-{}-{}.wav",
+        std::process::id(),
+        stem
+    ));
+    #[cfg(unix)]
+    {
+        if let Ok(f) = std::fs::File::create(&tmp) {
+            drop(f);
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(&tmp, std::fs::Permissions::from_mode(0o600)).ok();
+        }
+    }
+
+    let status = std::process::Command::new("ffmpeg")
+        .args([
+            "-y",
+            "-i",
+            system.to_str()?,
+            "-i",
+            voice.to_str()?,
+            "-filter_complex",
+            "[0:a][1:a]amix=inputs=2:duration=longest:normalize=0",
+            "-ac",
+            "1",
+            "-ar",
+            "16000",
+            "-c:a",
+            "pcm_s16le",
+            tmp.to_str()?,
+        ])
+        .stdin(std::process::Stdio::null())
+        .stdout(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
+        .status()
+        .ok()?;
+    if !status.success() {
+        let _ = std::fs::remove_file(&tmp);
+        tracing::warn!(
+            audio = %audio_path.display(),
+            "stem mix via ffmpeg failed; falling back to .mov decode"
+        );
+        return None;
+    }
+    tracing::info!(
+        audio = %audio_path.display(),
+        mixed = %tmp.display(),
+        "using mixed stems instead of .mov for transcription (workaround for dual-track 2x bug)"
+    );
+    Some(tmp)
+}
+
 fn log_attribution_decision(
     audio_path: &Path,
     output_path: &Path,
@@ -1820,14 +1899,21 @@ where
 
     // Step 1: Transcribe (always)
     on_progress(PipelineStage::Transcribing);
-    tracing::info!(step = "transcribe", file = %audio_path.display(), "transcribing audio");
+    // Workaround: if this is a native-call .mov with stems beside it, transcribe
+    // a freshly-mixed PCM from the stems to avoid the .mov dual-track 2x bug.
+    let mixed_stem_path = prepare_transcription_input(audio_path);
+    let transcribe_input = mixed_stem_path.as_deref().unwrap_or(audio_path);
+    tracing::info!(step = "transcribe", file = %transcribe_input.display(), "transcribing audio");
     let step_start = std::time::Instant::now();
     let result = crate::transcription_coordinator::transcribe_path_for_content_with_hints(
-        audio_path,
+        transcribe_input,
         content_type,
         config,
         decode_hints,
     )?;
+    if let Some(path) = mixed_stem_path.as_ref() {
+        let _ = std::fs::remove_file(path);
+    }
     let transcribe_ms = step_start.elapsed().as_millis() as u64;
     let transcript = if content_type == ContentType::Meeting {
         normalize_transcript_for_self_name_participant(
--- a/crates/core/src/pipeline.rs
+++ b/crates/core/src/pipeline.rs
@@ -533,6 +533,33 @@ fn expected_voice_stem_path(audio_path: &Path) -> Option<std::path::PathBuf> {
     Some(dir.join(format!("{}.voice.wav", stem)))
 }
 
+/// RAII handle for a stem-mix temp file. Dropping the value unlinks the
+/// file from /tmp, including on early Err returns from the transcription
+/// coordinator. Belt-and-suspenders against future call-site refactors
+/// that might forget the manual cleanup (#235 review item #1).
+///
+/// The temp file contains raw meeting audio, so leaking it on error is
+/// a privacy issue, not just a cleanliness one. The Drop impl deliberately
+/// swallows the unlink error (the file may already be gone if the OS or
+/// a test harness cleaned it up); leaving a partial file behind is the
+/// only failure mode worth thinking about and `Drop` cannot recover from
+/// it any better than the existing best-effort logic did.
+struct MixedStemTempFile {
+    path: std::path::PathBuf,
+}
+
+impl MixedStemTempFile {
+    fn as_path(&self) -> &Path {
+        &self.path
+    }
+}
+
+impl Drop for MixedStemTempFile {
+    fn drop(&mut self) {
+        let _ = std::fs::remove_file(&self.path);
+    }
+}
+
 /// Workaround for the .mov dual-track decode bug.
 ///
 /// macOS SCRecordingOutput writes a `.mov` with two audio tracks (system + mic)
@@ -541,10 +568,15 @@ fn expected_voice_stem_path(audio_path: &Path) -> Option<std::path::PathBuf> {
 /// to receive garbled samples and emit gibberish.
 ///
 /// When stems are present, mix them into a 16kHz mono PCM wav via ffmpeg and
-/// hand that to the transcriber instead. Returns the temp file path on success;
-/// caller must remove it after use. Falls back silently (returns None) on any
-/// error so the original `.mov` decode path stays as a safety net.
-fn prepare_transcription_input(audio_path: &Path) -> Option<std::path::PathBuf> {
+/// hand that to the transcriber instead. Returns a `MixedStemTempFile` whose
+/// `Drop` impl removes the temp file automatically, so the caller cannot
+/// leak raw meeting audio on a transcription error. Falls back silently
+/// (returns None) on any error so the original `.mov` decode path stays as
+/// a safety net — note that on macOS 26 native captures this fallback hides
+/// the dual-track bug itself, which is the structural concern jmh1313 is
+/// addressing in the v2 of this PR by switching to a typed hard-error
+/// return for the should-have-but-couldn't case.
+fn prepare_transcription_input(audio_path: &Path) -> Option<MixedStemTempFile> {
     let ext = audio_path
         .extension()
         .and_then(|e| e.to_str())
@@ -560,9 +592,20 @@ fn prepare_transcription_input(audio_path: &Path) -> Option<std::path::PathBuf> 
         return None;
     }
 
+    // Tempfile name includes pid + nanosecond timestamp + stem so two
+    // concurrent invocations on the same recording cannot land on the
+    // same path (Tauri's recovery path can spawn a worker thread for a
+    // recording whose `processing` flag is mid-CAS, which would otherwise
+    // collide here). Falls back to pid+stem alone if SystemTime is
+    // unavailable.
+    let unique_suffix = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_nanos())
+        .unwrap_or(0);
     let tmp = std::env::temp_dir().join(format!(
-        "minutes-stem-mix-{}-{}.wav",
+        "minutes-stem-mix-{}-{}-{}.wav",
         std::process::id(),
+        unique_suffix,
         stem
     ));
     #[cfg(unix)]
@@ -574,6 +617,15 @@ fn prepare_transcription_input(audio_path: &Path) -> Option<std::path::PathBuf> 
         }
     }
 
+    // ffmpeg amix defaults: `duration=longest` is the framework default
+    // (specifying it explicitly was redundant); `normalize=1` is the
+    // default and prevents combined-amplitude clipping when one stem is
+    // significantly louder than the other. System audio is usually
+    // hotter than mic, so `normalize=0` could bake clipping into the
+    // PCM before whisper sees it (jmh1313 confirmed the original PR's
+    // recordings had stems at -0.0 and -0.1 dB peak, which would have
+    // clipped on normalize=0). Default normalization is the safer
+    // choice unless we add explicit weights with measured levels.
     let status = std::process::Command::new("ffmpeg")
         .args([
             "-y",
@@ -582,7 +634,7 @@ fn prepare_transcription_input(audio_path: &Path) -> Option<std::path::PathBuf> 
             "-i",
             voice.to_str()?,
             "-filter_complex",
-            "[0:a][1:a]amix=inputs=2:duration=longest:normalize=0",
+            "[0:a][1:a]amix=inputs=2",
             "-ac",
             "1",
             "-ar",
@@ -609,7 +661,7 @@ fn prepare_transcription_input(audio_path: &Path) -> Option<std::path::PathBuf> 
         mixed = %tmp.display(),
         "using mixed stems instead of .mov for transcription (workaround for dual-track 2x bug)"
     );
-    Some(tmp)
+    Some(MixedStemTempFile { path: tmp })
 }
 
 fn log_attribution_decision(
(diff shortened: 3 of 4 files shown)


silverstein (changes requested):
First off, this is an incredible contribution. The investigation alone was high-quality work: 7 historical captures all showing exactly 2.00x, ffprobe evidence, and the recovery data is concrete (74-minute meeting from gibberish to 11,139 coherent words is the kind of validation I rarely get on bug reports). Thank you. Genuinely.

I ran the diff through codex adversarial review and there are six things I want to nail down before merging. None of them block the conceptual approach (stems first, fall back to mov), but a few are sharper edges than they look.

## 1. Tempfile cleanup leaks raw meeting audio on transcription error

The cleanup at the call site runs after `transcribe_path_for_content_with_hints(...)?`, so any Err short-circuits before `fs::remove_file`. The temp PCM is raw meeting audio sitting in `/tmp` until the OS reaps it. This is the one that worries me most, since meeting audio is sensitive.

Minimal fix that does not need a new dep:

```rust
let mixed_stem_path = prepare_transcription_input(audio_path);
let transcribe_input = mixed_stem_path.as_deref().unwrap_or(audio_path);
let result = crate::transcription_coordinator::transcribe_path_for_content_with_hints(
    transcribe_input,
    content_type,
    config,
    decode_hints,
);
if let Some(path) = mixed_stem_path.as_ref() {
    let _ = std::fs::remove_file(path);
}
let result = result?;
```

Or wrap in a `Drop`-impl struct that owns the path. Either is fine.

## 2. Tempfile naming collision under Tauri recovery race

`minutes-stem-mix-${pid}-${stem}.wav` is unique per process per recording, but `commands.rs:6235-6258` (Tauri recovery) does a non-atomic `processing.load` / `processing.store(true)` before spawning the worker thread. A double-invocation race on the same recording would overwrite the same temp file. Either add a nanosecond suffix (`std::time::SystemTime::now().duration_since(UNIX_EPOCH)`) or switch to `tempfile::NamedTempFile` for the unique-by-construction property.

## 3. The stated decode root-cause does not match the code I see

This is the one I want your read on more than a code change. Per `transcribe.rs:1132-1164` the symphonia fallback picks the first non-null audio track and skips packets whose `track_id` differs from the chosen one. Per `transcribe.rs:947-954` the normal `.mov` path prefers ffmpeg first, and ffmpeg auto-stream-selects a single audio stream absent `-map`. So the "whisper reads both streams sequentially" theory does not obviously hold against the code.

The fix clearly produces correct output (your 6/7 recovery data is hard evidence). What I cannot reconcile is *why* the unpatched path was producing 2x output. Possibilities I have not ruled out:

- A third decode path I am missing, possibly via an older ffmpeg invocation that uses `-map 0:a` and grabs all audio streams.
- whisper-cli vs whisper-rs behavior diverging on a multi-track container.
- Something specific about how `pipeline.rs::transcribe_to_artifact` constructs the decode call before reaching the transcription coordinator.

If you saved the log output from one of your before-fix runs (a `step: transcribe` line with the inflated duration), or if you can briefly check which of `transcribe_with_ffmpeg_pipe` vs `transcribe_with_symphonia` actually fired on your machine, that would unlock the diagnosis. The fix may still be correct even if my decode-path read is right (it strictly avoids the entire `.mov` decode), but I want to know what we actually fixed.

## 4. Silent fallback swallows failures in the exact case this targets

When stems are missing or ffmpeg fails on a `.mov`, the helper returns None and the caller silently uses the `.mov`. On macOS 26 native captures that means 2x gibberish still gets to the user with no signal beyond the warn log. Two options:

- Return `Result<Option<PathBuf>>` and hard-error on the path that should-have-mixed-but-could-not.
- Keep `Option<PathBuf>` but escalate to a typed error via the pipeline so the UI shows "stem mix failed, native call captures need ffmpeg and stems, please report" instead of silently producing nonsense.

I lean toward (1) because the fallback is masking exactly the bug we are trying to fix.

## 5. `amix=normalize=0` risks baking in clipping

ffmpeg amix defaults to `normalize=1` (auto-attenuate so combined amplitude stays sane). System audio is usually louder than the mic, so `normalize=0` on two correlated stems can produce clipping in the PCM before whisper-guard or any normalization step Minutes does. Suggest defaulting to `normalize=1` or specifying explicit weights with measured levels.

Also `duration=longest` is ffmpeg's default; no value in being explicit.

## 6. Background job path is uncovered

`jobs.rs:1173-1181` calls `pipeline::transcribe_to_artifact` at `pipeline.rs:894-949`, which does not go through `process_with_progress_and_sidecar`. Anything that lands in `~/.minutes/jobs/` and gets picked up by the worker (including #229's orphan-WAV retry path) will still hit the unpatched `.mov` decode. The wiring needs to live in `transcribe_to_artifact` or in a shared helper both call into.

## 7. Diarization checked, unaffected

Just confirming for the record: diarization passes the original `audio_path` and discovers stems independently via `diarize_with_context -> diarize_from_source_aware_stems`, so the mixed PCM is transcription-only and the diarization path is untouched by this change. Codex verified, all good.

## On tests

For the test fixture question, my honest preference is integration-style: a fake `.mov` plus a matched `.voice.wav` and `.system.wav` pair in a `tempdir`, assert that `prepare_transcription_input` returns Some(path), assert that path exists and starts with the right magic bytes for PCM WAV, drop the function, assert cleanup. The `.mov` can be a 1-byte file because the function only sniffs extension + stem siblings, it does not parse the container. The wavs can be 16kHz mono silence generated with `hound` in the test setup. That gives deterministic coverage without needing real audio.

---

Net of all this: the conceptual approach is right, the data is solid, and I want to land this. Happy to take a v2 push from you with (1) (2) (4) (5) (6) addressed, plus your read on (3). If you would rather I push a follow-up commit to your branch with the mechanical pieces (1, 2, 5) and leave the structural ones (3, 4, 6) for you, that works too. Your call. Either way, this is going to ship.

Thank you again for the work on this whole bundle today, between #231 + #234 + this PR you have moved the macOS 26 story forward more than any single commit of mine would have.

Mathieu

jmh1313:
@silverstein thank you for this. The review caught things I would have missed and gave me a sharper bug to chase.

Before responding to each item, the headline you should know:

**Item #4 already bit me in production, today.** This morning's 42-minute meeting on my own machine landed as gibberish despite the patched binary being the one the watcher loaded. The log shows `step: transcribe` reported `audio: 5033.7s` against a 2516s real recording (exactly 2.00x). The "stem mix" log line never fired. The "stem mix failed" fallback log line also never fired. The helper returned `None` silently at the sibling-stems existence check, the caller used the `.mov` as-is, and Whisper produced 6,073 words of overlapping nonsense with `*Grunting*` / `(gurgling)` / fabricated phrases.

So your point about the fallback masking exactly the bug we're trying to fix was correct in advance. It masked it for me, eight hours after I shipped the PR.

**Plus a related disclosure I owe you up front:** my "6 of 7 historical captures recovered" data in the issue body was misattributed to this patch. Re-reading my own log entries today, every one of those reprocessed runs operated on a `/tmp/<name>-mixed.wav` file that had been pre-mixed outside the pipeline. The first thing `prepare_transcription_input` does is `if ext != Some("mov") { return None; }`, so on those reprocess runs the helper short-circuited on the extension check and the caller used the pre-mixed `.wav` directly. The recovery I cited was validating a manual mix workflow that uses the patched binary as a passthrough, not the patch's own logic. The patched logic had never run end-to-end against a fresh native-call capture until today, and when it did, it failed.

**Plus a bigger structural flaw than #4 alone:** the sibling-stems lookup is wrong for the live capture path entirely. In my live pipeline run today, the `.mov` lived at `~/.minutes/native-captures/2026-05-12-100439-call.mov` and the stems were written from the start to `~/meetings/<calendar-derived-title>.voice.wav` and `.system.wav`, with totally different base names and a different directory. They were never siblings of the `.mov`. `find ~ -name "2026-05-12-100439-call*"` returns nothing for the `.voice.wav` / `.system.wav` patterns. So even before #4 swallowed the failure, the helper couldn't have found anything to mix.

The good news from today's evidence: the stems themselves are pristine (voice: mean -30.9 dB, peak -14.5 dB mid-meeting; system: similar), and a manual `ffmpeg amix` (default `normalize=1`) of those stems is currently reprocessing successfully on my end. So the conceptual approach is still right; the locator is just looking in the wrong place.

---

On your seven points:

**#1 (tempfile leak on transcription error).** Agree, fixing in v2 with the `let result = ...; cleanup; result?` pattern you wrote. Meeting audio in `/tmp` is unambiguously a privacy issue. I'll also add a `Drop` impl on the path holder as a belt-and-suspenders so future call-site refactors can't reintroduce the leak.

**#2 (tempfile name collision under Tauri recovery race).** Agree. Switching to `tempfile::NamedTempFile` (you already have `tempfile` in the dep tree via `whisper-rs-sys` transitively, so no new direct dep). Falling back to nanosecond suffix if you'd prefer to avoid adding a direct dependency on `tempfile`. Your call.

**#3 (the decode-path question, your read vs my claim).** This is the one I owe you the most on, especially now that I have a fresh in-production reproduction. Today's `step: transcribe` log line for the failed run:
```
duration_ms: 1655630
diagnosis: audio: 5033.7s, whisper produced 872 segments, no_speech rescue: 1 segments saved (would have been blank), dedup: -52 → 820, script filter: -15 → 805, final: 6073 words
file: /Users/jakehirschfeld/.minutes/native-captures/2026-05-12-100439-call.mov
step: transcribe
```
5033.7s for a 2516.74s real recording. No `whisper-cli stderr` / `transcribe_with_ffmpeg_pipe` / `transcribe_with_symphonia` tags in the same log block, so I can't tell from this alone which decoder fired. I'll dig into `transcribe.rs:947-954` and `transcribe.rs:1132-1164` with a tracing patch and confirm, then update this PR with the answer. Hypotheses I want to actively rule in or out, in your order:
- A third decode path: I'll grep for any `Command::new("ffmpeg")` / `Command::new("ffprobe")` constructions in the audio-input layer that use `-map 0:a` or `-map 0` (which would grab all audio streams). If that exists, it's the candidate.
- whisper-cli vs whisper-rs divergence: I'm on macOS 26.2 Apple Silicon, Homebrew bottle. Will report which one ships in the bottle and how it handles a multi-track input directly via a 30-second isolated repro.
- `transcribe_to_artifact` constructing the call differently: tied to your #6 below; will check both entry points and report.

This is the most useful single thing I can produce for you, so I'll prioritize it.

**#4 (silent fallback swallows failures).** Conceded fully. Switching to `Result<Option<PathBuf>>` per option (1), with the helper returning `Ok(None)` only on the explicit "not a .mov" / "no stems expected" path, and `Err(...)` with a typed `MinutesError` variant for "this looked like a native-call capture but we couldn't produce a mix." On macOS 26 + native-call, falling through to the broken decode is strictly worse than failing loud.

**#5 (`amix=normalize=0` clipping).** Agree, removing `normalize=0`. Removing `duration=longest` too since it's the default. Pre-mix levels on the stems today were peak -0.1 dB and -0.0 dB on the two stems, which would have clipped on `normalize=0` exactly as you said. Today's recovery mix used the default and came out clean (peak -5.4 dB post-mix).

**#6 (background job path uncovered).** Conceded. Extracting the input-preparation logic into a shared helper at the boundary of both `process_with_progress_and_sidecar` and `transcribe_to_artifact`, so any path that ends at the transcription coordinator goes through the same input pre-check. This was the structural miss I most regret not catching.

**#7 (diarization independence).** Confirmed, and useful context for #4. I'll be looking at `diarize_with_context -> diarize_from_source_aware_stems` anyway since that's the stem-discovery path I actually want to model the transcription stem-lookup on, given the live-pipeline directory layout I described above. If that helper exposes a stable "given a .mov path, return the stem paths" function, the right move is to call it from both places rather than reimplement the discovery in `prepare_transcription_input`.

---

**On tests.** Your integration shape is what I want too. I'll add the `tempdir` fixture, the 1-byte `.mov` sniff target, and `hound`-generated 16kHz silence stems. Will assert: returns `Some(path)` on the happy path, the returned PCM exists and starts with `RIFF`/`WAVE`, drops cleanly on early return, and returns `Err` (per #4) on the should-have-but-couldn't path. Cover the `transcribe_to_artifact` entry point too (#6).

**On your offer to push #1, #2, #5 directly.** Yes, gratefully. I'd rather focus my time on (a) the #3 diagnosis with the in-prod data I just collected, (b) the stem-lookup correctness fix that today's failure surfaced, and (c) the structural #4/#6 reshape with proper error typing. If you push 1/2/5 to my branch I'll rebase and continue on top.

**Timeline:** if you can take 1/2/5 in the next day or so, I can have v2 with #3 diagnosis + corrected stem lookup + #4 hard-error + #6 shared helper + tests up within 48 hours of seeing your pushes.

Thank you again. This is a much sharper bug than the one I filed.

silverstein:
Three things, in order.

**One: thank you for the disclosure on the validation data.** Re-reading your reply twice. The fact that you traced your own "6 of 7 captures recovered" claim back through the logs, realized it was the extension short-circuit and not the patch logic, and then flagged it before I asked for clarification is exactly the kind of integrity I want in the project. Most contributors would have quietly noticed and quietly fixed. You wrote it down in plain English on the PR. That matters.

**Two: the stem-lookup directory mismatch is the bigger find.** When the live capture writes `.mov` to `~/.minutes/native-captures/` and stems to `~/meetings/<calendar-title>.voice.wav`, the sibling-existence check returns None instantly and the rescue never even gets to the failure modes I was reviewing. Your in-prod 5033.7s / 2516s evidence is the definitive 2.00x repro we needed. That is also one more reason to model the stem discovery on `diarize_with_context` / `diarize_from_source_aware_stems` rather than re-implement it in `prepare_transcription_input`. Sharing the discovery function across diarization and transcription is the right structural shape, not just convenient.

**Three: pushed #1, #2, #5 to your branch as `f843d06`.** One commit, three review items. The commit message has the rationale and explicitly notes which review items it does NOT address so you have a clean handoff back to v2 work.

Summary of what's on your branch now:

- `MixedStemTempFile` RAII handle. Dropping the value unlinks the file from /tmp, so cleanup runs on the success path, the Err path, panic, and any future call-site refactor. The explicit `drop(mixed_stem_path)` after the transcribe call makes the lifetime legible. Belt-and-suspenders Drop shape you asked for, plus the `let result; cleanup; result?` flow folds naturally into the Drop boundary.
- Nanosecond-suffix tempfile name: `minutes-stem-mix-${pid}-${nanos}-${stem}.wav`. Stayed off the direct `tempfile` dep since it is one filename. If you would rather have `NamedTempFile`, the dep is already transitive and the swap is small, your call.
- `amix=inputs=2` (no `normalize=0`, no `duration=longest`). Your in-prod -0.0 dB / -0.1 dB stem peaks would have clipped on `normalize=0` exactly as you described; default normalization is the right default for two correlated stems without measured weights.

cargo check + clippy + fmt clean across the workspace on this commit.

What I deliberately did NOT touch, all of it yours for v2:

- Item #3: decode-path diagnosis with your fresh in-prod data.
- Item #4: `Result<Option<PathBuf>>` reshape with a typed `MinutesError` variant for the hard-error case.
- Item #6: shared helper at the `transcribe_to_artifact` + `process_with_progress_and_sidecar` boundary.
- The stem-lookup correctness fix you discovered today.

When you have a v2 ready, ping me. Take whatever time you need on the diagnosis. Getting the decode path right matters more than getting it shipped fast.

One more thing on #3 that might save you a grep: when you trace which decoder fires, also check whether either decode path is gated on extension at the entry. If `transcribe_path_for_content_with_hints` dispatches by extension and `.mov` lands in a path that uses an older ffmpeg invocation with `-map 0` or `-map 0:a`, that would be the third decode path. Worth a `git log -S '-map 0:a' crates/core/src/` to see if any such call has ever existed in the repo, then a search for `.mov` extension dispatch.

Mathieu

silverstein (commented):
Thanks for tackling this. The stem-mixing approach is right, and the fact that it leverages the same survival path as #216 is a good call. I ran an adversarial code review and there are a few gaps worth closing before merge. Want to surface them with full context so you can decide whether to push more commits or whether I should land follow-up patches on top.

## Real blockers

1. **Silent fallback to the broken `.mov` path.** When `prepare_transcription_input` returns `None` because stems are missing or ffmpeg failed, the caller `.unwrap_or(audio_path)` re-enters the exact 2x-duration decode that this PR exists to prevent. For a `.mov` user whose stems didn't survive a crash, the result is still garbled. Probably want a typed result like `Result<PathBuf, PrepareError>` where the caller can distinguish "no stems needed because not a `.mov`" from "stems were expected but unavailable", and error out loudly in the second case rather than silently transcribing the broken `.mov`.

2. **`transcribe_to_artifact` is the other transcription entry point.** Background jobs and `minutes process` go through `transcribe_to_artifact` (`pipeline.rs:949-954`, `jobs.rs:1173-1181`), which still passes the raw `.mov`. The fix only wires `process_with_progress_and_sidecar`. Same prep should run there too, otherwise CLI users hit the 2x bug.

3. **Tests.** The new helper has zero coverage. Minimum useful set: happy path with both stems present, ffmpeg failure path, one stem missing, mismatched stem durations, non-`.mov` returns identity. The diarization side has prior art for stem validation in `diarize.rs:566` (`stem_has_audio`) that this can lean on.

## Smaller stuff

- `.exists()` accepts a zero-byte stem written during a partial crash. The diarization path validates via `stem_has_audio` (`diarize.rs:566-567`). Worth reusing.
- Symlinked `.mov` paths look for stems beside the symlink, not the canonical target. Recent #237 work touched this exact area, worth canonicalizing before stem lookup.
- Could reuse `discover_stem_plan` from `diarize.rs:560-610` instead of duplicating a simpler version.
- `.beads/issues.jsonl` churn in the diff is unrelated; cleaner if stripped to its own commit or dropped entirely.

## What I'd like to do

If you have bandwidth and want to keep the PR yours, push fixes for blockers 1, 2, and 3 (typed result, second entry point, basic test coverage). Happy to design-review the result shape if that helps.

If you don't have time, say the word and I'll land the missing pieces on top of your stem-mixing approach in a follow-up commit so the bug fix ships this week. You stay credited.

The stem-mixing core is solid. Just want to make sure we don't ship a fix that quietly resurfaces the bug on the unhappy paths.

silverstein:
Apologies, ignore the previous review. I lost context between sessions and posted as if this was a fresh PR. I just re-read the May 12 thread and the codex pass I ran is ~90% overlap with the changes already requested. The v2 work you outlined (typed Result for the silent-fallback case, shared helper across both transcription entry points, tests, the stem-lookup directory fix) covers it.

One small thing the prior review did not explicitly raise that came up in the rerun: if a `.mov` is reached via symlink, the stem lookup needs to canonicalize the path first or stems beside the canonical target will be missed. Recent #237 touched this exact area. Worth a one-line `fs::canonicalize` before sibling lookup in the shared helper.

Take your time on v2. Looking forward to it.

jmh1313:
No worries on the duplicate review. The new items from the rerun pass net out to three real folds I'm adding to v2:

1. **`fs::canonicalize` before stem lookup** in the shared helper, so symlinked `.mov` paths resolve to the canonical target before sibling lookup. Will check #237 to make sure I'm consistent with that path's treatment.
2. **Reuse `discover_stem_plan`** at `diarize.rs:560-610` instead of reimplementing the lookup in `prepare_transcription_input`. The existing function already returns `Option<SourceAwareDiarizationPlan>` with the three variants we need (full / system-only / silent-system), already validates via `stem_has_audio`, already has logging. Wire-up is a one-line call; transcription and diarization share one discovery path.
3. **`stem_has_audio` validation** falls in for free since `discover_stem_plan` calls it. Solves the zero-byte-stem case the original `.exists()` accepted.

For the record on the open #3 thread, since you asked it cleanly the first time:

- `.mov` always lands in `decode_with_ffmpeg` (`transcribe.rs:937` dispatches by extension).
- That function auto-selects without `-map`. `git log -S '-map 0:a' --all crates/core/` is empty; no third decode path exists historically.
- Doubling is inside `decode_with_ffmpeg` (the `audio_duration_secs` in the diagnosis line is computed from sample count after `load_audio_samples`, so the doubling is pre-whisper). `whisper_rs` is the only whisper invocation in core; no whisper-cli path.
- A synthetic two-stream `.mov` I built with `ffmpeg -map 0:a -map 1:a -c copy` decodes correctly through the production path (5.0155s for 5s input). So plain "MOV with two audio streams" is not the trigger.
- Best remaining hypothesis: something specific in how `SCRecordingOutput` writes the container (edit lists, track groups, or matched dispositions) trips ffmpeg's auto-select on Apple Silicon macOS 26. Will ffprobe a fresh capture for forensics when I next run a native-call recording and follow up. The patch sidesteps it regardless.

Working on v2 now. Will ping when ready.

jmh1313:
Pushed v2 to the branch. Three structural changes plus an important narrative correction I owe you.

## What v2 changes

**Typed error replacing silent fallback (#4).** `prepare_transcription_input` now returns `Result<Option<MixedStemTempFile>, MinutesError>` with a new `TranscribeError::NativeCaptureStemMixUnavailable { reason }` variant. The three should-have-mixed-but-couldn't cases (voice stem missing, system stem zero-byte partial-crash, ffmpeg amix failed) hard-error instead of silently re-entering the broken `.mov` decode. Non-`.mov` inputs and stemless `.mov` files still return `Ok(None)`.

**`transcribe_to_artifact` wired into the same prep (#6).** v1 only covered `process_with_progress_and_sidecar`. v2 adds the same `prepare_transcription_input` call at the start of `transcribe_to_artifact` (pipeline.rs:1373-1391), so background jobs and `minutes process` now route through the fix too. `jobs.rs:1174` is the relevant entry that v2 newly covers.

**Stem-lookup hardening.** Reuses `discover_stem_plan` (now `pub(crate)`) from diarize.rs:560 instead of duplicating the lookup in `prepare_transcription_input`. That pulls in `stem_has_audio` for free, which catches zero-byte partial-crash stems that `.exists()` would have accepted. `fs::canonicalize` runs before lookup so symlinked `.mov` files resolve to their canonical target (#237's area).

**6 integration tests** in a new `prepare_transcription_input_tests` module: returns_ok_none cases, mixes_stems_when_both_present_and_valid (asserts RIFF/WAVE magic + Drop cleanup), errors on missing voice / zero-byte system, canonicalizes symlinked `.mov`. ffmpeg-required tests skip cleanly when ffmpeg isn't on PATH.

cargo check + clippy + fmt clean. All 733 lib tests + 8 integration tests + 6 new tests pass.

## Narrative correction I owe you

In my reply two messages up I claimed the live capture writes the `.mov` to `~/.minutes/native-captures/` and the stems to `~/meetings/<calendar-derived-title>.voice.wav`, framing this as a "directory mismatch" the v1 stem-lookup got wrong.

That was wrong, and I caught it in a pre-push code review. `tauri/src-tauri/src/system_audio_record.swift:87-90` derives stem URLs from `outputURL.deletingLastPathComponent()` plus the same basename, so stems are written as **siblings** of the `.mov`. `jobs.rs::move_stems_with_audio` (305-321) preserves that invariant on move; `rescue_paired_stems` keeps stems beside the `.mov` even in failed-capture recovery.

What I actually observed (stems in `~/meetings/`, `.mov` in `~/.minutes/native-captures/`) was the post-processing state after the trio is moved together, not the during-capture layout. v1's sibling-based lookup was therefore correct for the live path; the "directory mismatch fix" framing in my prior comment was a misreading on my part.

What v2 does still fix in this area, and why the work was worth doing:

- v1's inline `.exists()` accepted zero-byte stems from a partial crash. v2's `stem_has_audio` reuse via `discover_stem_plan` rejects them and surfaces a typed error.
- v1 didn't canonicalize before lookup, so a symlinked `.mov` would still miss its stems (the #237 case).
- v1's duplicated stem discovery would have drifted from diarization's. v2 funnels both through `discover_stem_plan` so the "stems present" definition is single-source.
- v1 covered only one of the two transcription entry points (#6).
- v1 silently fell through to the broken decode on any failure (#4).

So the structural value of v2 holds. The framing in my prior comment did not. Apologies for the misdirection.

## #3 (decode-path diagnosis)

My preliminary findings from the prior reply stand: `.mov` dispatches to `decode_with_ffmpeg` at `transcribe.rs:937`, which uses auto-select without `-map`. `git log -S '-map 0:a' --all crates/core/` is empty. The doubling is inside `decode_with_ffmpeg` (the `audio_duration_secs` value is computed from `samples.len() / sample_rate` after `load_audio_samples` returns), so it's pre-whisper. A synthetic two-stream `.mov` built with `ffmpeg -map 0:a -map 1:a -c copy` decodes correctly through the production path, so the trigger is specific to SCRecordingOutput's container shape (edit lists, track groups, or matched dispositions in the `.mov` atoms). I'll ffprobe a fresh capture for forensics when I next run a native-call recording.

The fix sidesteps this regardless, but the underlying ffmpeg auto-select behavior on SCRecordingOutput output is still worth understanding. Logging-only warning in `decode_with_ffmpeg` for "multi-stream input detected, picking first" is a candidate defense-in-depth follow-up if you want it.

## Acknowledged test gaps (not in v2 scope)

The pre-push review flagged two test gaps I chose not to close in this commit because the infrastructure was heavier than the marginal coverage gain. Happy to file as follow-ups if you want them tracked:

1. **`transcribe_to_artifact` end-to-end test.** Would verify the entry-point wiring directly via a partial-stem `.mov` driven through `transcribe_to_artifact`, but requires a `BackgroundPipelineContext`, `Config`, and a way to short-circuit whisper transcription. I covered the helper level (6 tests on `prepare_transcription_input` directly) and confirmed the entry-point wiring by inspection.
2. **ffmpeg-fails branch.** Would exercise the `ffmpeg amix failed` reason string. Synthesizing a WAV that `stem_has_audio` accepts but ffmpeg rejects is genuinely hard (the helper's validation is strong by design). Possible via PATH manipulation in a test harness, but adds significant complexity.

If you want either landed before merge, say the word and I'll add them.

## Commits on the branch now

- `1ce1393` (mine) v2 typed errors, second entry point, helper reuse, tests
- `92fa5f2` (yours) mechanical cleanup, with .beads/issues.jsonl churn removed during the rebase (your authorship preserved; this was the unrelated churn you flagged in your second review)
- `73c70ec` (mine) initial stem-mix fix

Rebased onto current `main` (`3718547`). `.beads/issues.jsonl` no longer in the PR diff. force-with-lease used for the push so no chance of clobbering anything.

Ready for your re-review.

silverstein:
v2 looks great. Read every change end-to-end, ran the full gate locally on the rebased branch, all 733 lib tests pass, clippy + fmt clean. All five v1 blockers cleanly addressed and the `discover_stem_plan` reuse is exactly the shape I was hoping for.

Two small things I handled inline so we can land this:

1. Bumped `MINUTES_TEST_COUNT` 1100 to 1106 in `site/lib/release.ts` to match the six new tests in `prepare_transcription_input_tests`. The Site Release Link Consistency check is a CI gate that compares the constant to the actual count and was failing on 1100. Pushed as commit `7554718` to your branch.
2. The Vercel deploy failure is a permissions thing (external contributors do not have Vercel team access), not a code issue. Will not block the merge.

The narrative correction you posted about the stem directory layout is the kind of contributor discipline that is rare. The structural value of v2 holds regardless and the rest of the PR is solid.

On the two test gaps you flagged (end-to-end `transcribe_to_artifact` test and the ffmpeg-fails branch): both are fine as follow-ups. The helper-level coverage you added is the right shape and synthesizing a WAV that passes `stem_has_audio` but fails ffmpeg is genuinely hard. If real-world data ever surfaces a failure mode there we will revisit.

Merging once CI clears on the new commit.

silverstein:
Quick update before merge. Ran the codex adversarial pass on v2 + my test-count bump, and codex found one remaining variant of the silent-fallback bug that v2 itself did not cover.

The shape: `discover_stem_plan` returns `None` for two semantically different cases (`diarize.rs:600-606`):

1. Neither stem present on disk, ordinary non-native-call `.mov`, fall through is correct.
2. Voice stem present and audible, but system stem entirely absent from disk, partial-crash native capture, fall through silently re-enters the 2x decode bug.

v2 handled the zero-byte system stem case (via `stem_has_audio` rejecting it) but the system-stem-completely-absent case still collapsed to `Ok(None)` and fell through.

Pushed commit `7187277` to your branch with:

- Independent check for a usable sibling voice stem when `discover_stem_plan` returns `None`. Surfaces the same typed `NativeCaptureStemMixUnavailable` error.
- New regression test `errors_when_voice_present_and_system_stem_file_absent` to pin the contract.
- `MINUTES_TEST_COUNT` bumped 1106 to 1107.

Everything else in v2 verified clean by codex: typed error propagation, both call sites wired, `discover_stem_plan` reuse, RAII Drop cleanup, canonicalization, `.beads/issues.jsonl` removed, test count bookkeeping matches. cargo fmt + clippy + tests all green locally on the new commit.

Will merge once the new CI clears.

PR #250, opened 2026-05-16
docs: rewrite PARAKEET.md install guide to fix 9 verified gotchas

rymalia:
## Summary

Rewrites `docs/PARAKEET.md` (and corrects one misleading line in `README.md`) based on an end-to-end install of Parakeet + Apple Silicon Metal on an M3 Air, validated against v0.18.0 of Minutes (Homebrew + DMG channels) and parakeet.cpp HEAD as of 2026-05-15. Doc-only; no source changes.

**Validation:** install was driven start-to-finish using the new recipe, then verified with two successful test recordings through the desktop app. `parakeet` and `example-server` binaries link Metal + MPS + MPSGraph + Accelerate (confirmed via `otool -L`). `tdt-600m.safetensors` is 2.3 GB / 627M params / 723 tensors mapped.

**Context the reviewer may want:**
- No upstream parakeet.cpp binary releases exist as of 2026-05-15 — building from source is currently the only path, which is why the PR keeps a full build recipe rather than pointing at a prebuilt artifact.
- The two CMake flags the rewrite leans on (`-DAXIOM_INSTALL=OFF -DPARAKEET_INSTALL=OFF`) are the maintainer's own escape hatch added at `CMakeLists.txt:96-100` in parakeet.cpp — they short-circuit the `install(EXPORT) AxiomTargets` strictness error on any CMake version, which is a cleaner fix than the doc's previous CMake-version-downgrade workaround.
- The `.nemo` HuggingFace URL (`huggingface.co/nvidia/parakeet-tdt-0.6b-v3/resolve/main/parakeet-tdt-0.6b-v3.nemo`) is publicly curl-able — HF transparently 302s to its xet-hub CDN with no auth — so the rewrite drops the `huggingface_hub` Python dependency.

## Changelog

| # | Gotcha (PARAKEET.md before this PR) | Fix |
|---|--------------------------------------|-----|
| 1 | "Option A: Use Minutes setup (recommended)" implied `setup --parakeet` downloads + converts the model. In v0.18.0 it only installs Silero VAD weights and prints manual instructions (see `crates/cli/src/main.rs:5258-5404`). | Replaced the Option A/B split with a single 6-step manual flow. Step 1 honestly describes `setup --parakeet` as VAD-weights + binary-resolution + recipe printout. |
| 2 | Doc didn't mention that the whisper-only Homebrew Formula CLI can still run `setup --parakeet` — useful when the user's transcription happens in the DMG-installed `.app` bundle, which does include the parakeet feature. | Added a one-line note in the Fastest Path noting the warning is informational on a whisper-only CLI; the setup itself completes. |
| 3 | `brew install cmake@3` no longer works — Homebrew dropped the formula. | Replaced with Kitware's official `cmake-3.31.12-macos-universal.tar.gz` tarball, installed into `~/.local/opt/` and prepended on `PATH` only for the build shell. |
| 4 | Claimed CMake 3.x alone fixes the `install(EXPORT) AxiomTargets` error. The strictness rule actually fires on any modern CMake. | Promoted the real fix (the maintainer's `-DAXIOM_INSTALL=OFF -DPARAKEET_INSTALL=OFF` flags) to the primary recipe. The atomics-check and EXPORT-strictness errors are now two distinct troubleshooting entries. |
| 5 | `cp build/bin/parakeet ~/.local/bin/` assumes the Ninja generator. With Unix Makefiles (default when Ninja isn't installed) parakeet lands in `build/` and example-server in `build/examples/server/`. | Replaced literal paths with `find build -type f -perm -u+x \( -name parakeet -o -name example-server \) -exec cp` so the copy step works under either generator. |
| 6 | `example-server` (the live-mode warm sidecar) is off by default in parakeet.cpp's CMake. | Added `-DPARAKEET_BUILD_SERVER_EXAMPLE=ON` to every configure invocation, with an inline comment explaining the live-mode performance impact of omitting it. |
| 7 | `tar xf *.nemo --wildcards --no-anchored '*tokenizer.vocab'` is GNU-tar-only; macOS BSD tar errors out. | macOS recipe is now `tar tf` + literal-UUID extract. The GNU form is preserved as a footnote for Linux / `gtar` users. |
| 8 | `pip install safetensors torch torchaudio huggingface_hub` was missing the transitive `packaging` dep (conversion crashes at import) and included two packages `convert_nemo.py` doesn't use. | Replaced with `uv pip install torch safetensors packaging numpy` — drops `torchaudio` (unused) and `huggingface_hub` (the `.nemo` is curl-able). |
| 9 | `.nemo` download instruction used `hf download` (requires `huggingface_hub`). | Replaced with `curl -L -O https://huggingface.co/nvidia/parakeet-tdt-0.6b-v3/resolve/main/parakeet-tdt-0.6b-v3.nemo`. |

## Files changed

- `docs/PARAKEET.md` — Fastest Path, Prerequisites/macOS, Build parakeet.cpp, Install Models, and Troubleshooting sections rewritten.
- `README.md` — Parakeet install block: the `# Multilingual v3 (tdt-600m, ~1.2GB)` comment on `minutes setup --parakeet` was misleading on two counts (the command doesn't auto-download the model; the converted safetensors is ~2.3 GB, not 1.2 GB). Replaced with an explicit note about what setup does vs. what's manual.

## Test plan

- [ ] `grep -nE "Option A|Option B|cmake@3|torchaudio|huggingface_hub" docs/PARAKEET.md` — every match should be either explanatory ("no longer works", "not required") or a GNU-tar footnote; no recipe steps should reference them.
- [ ] Section anchors used in cross-references resolve: `#troubleshooting`, `#cmake-4x-atomics-error-build`, `#prerequisites`.
- [ ] On a fresh Apple Silicon machine, run the Fastest Path end-to-end and confirm each shell block executes cleanly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -940,9 +940,13 @@ minutes setup --diarization
 # DMG and tagged CLI release binaries include the feature; the Homebrew Formula
 # CLI (`brew install silverstein/tap/minutes`) and bare `cargo install minutes-cli`
 # do not. See docs/PARAKEET.md for the source-build walkthrough.
-minutes setup --parakeet                          # Multilingual v3 (tdt-600m, ~1.2GB)
-minutes setup --parakeet --parakeet-model tdt-ctc-110m  # English-only compact model (~220MB)
-# Also installs native Silero VAD weights for the parakeet.cpp --vad path
+#
+# Note: `minutes setup --parakeet` installs the bundled Silero VAD weights
+# (~1.2 MB) and prints a manual recipe for downloading + converting the
+# tdt-600m .nemo from HuggingFace (the resulting safetensors file is ~2.3 GB).
+# The `.nemo` download and `convert_nemo.py` step are still manual.
+minutes setup --parakeet                          # Multilingual v3 setup (tdt-600m)
+minutes setup --parakeet --parakeet-model tdt-ctc-110m  # English-only compact (tdt-ctc-110m)
 
 # Enroll your voice for automatic speaker identification
 minutes enroll              # Records 10s of your voice
--- a/docs/PARAKEET.md
+++ b/docs/PARAKEET.md
@@ -73,30 +73,60 @@ feature, so this only matters for unusual build configurations.
 ## Fastest Path on Apple Silicon
 
 If you want the shortest path from "I have a Mac" to "Minutes is using
-Parakeet locally on Metal," do this. Each step is a separate command — do
-not paste TOML into the shell; step 4 writes a file.
+Parakeet locally on Metal," do this. Steps 1–3 are shell commands; step 4
+is a manual download + conversion (the CLI prints these instructions but
+does not run them automatically); step 5 writes a config file — do not paste
+the TOML block into the shell.
 
 Prerequisites first (see [Prerequisites](#prerequisites) for details):
 - Full Xcode installed (Command Line Tools alone is not enough — Metal needs
   the full Xcode shader compiler)
-- CMake 3.x. CMake 4.x trips both an atomics check and a new
-  `install(EXPORT)` strictness rule in parakeet.cpp's `axiom` submodule
-  (`target "axiom" ... requires target "hwy" that is not in any export set`).
-  Easiest fix: `brew install cmake@3` and prepend it on `PATH` for the
-  parakeet.cpp build. See [Troubleshooting](#troubleshooting).
+- CMake 3.31.x. CMake 4.x trips an atomics check in parakeet.cpp's bundled
+  Google Highway dependency on Apple Silicon. Homebrew no longer ships
+  `cmake@3`, so grab the official Kitware universal tarball:
+
+  ```bash
+  mkdir -p ~/.local/opt && cd ~/.local/opt
+  curl -L -O https://github.com/Kitware/CMake/releases/download/v3.31.12/cmake-3.31.12-macos-universal.tar.gz
+  tar xf cmake-3.31.12-macos-universal.tar.gz
+  export PATH="$HOME/.local/opt/cmake-3.31.12-macos-universal/CMake.app/Contents/bin:$PATH"
+  cmake --version    # expect 3.31.12
+  ```
+
+  Only export this `PATH` in the build shell — your system CMake stays
+  untouched everywhere else. The separate `install(EXPORT) ... AxiomTargets`
+  error is handled by build flags rather than CMake version; see step 1
+  below and [Troubleshooting](#troubleshooting).
 
 ```bash
 # 1. Build and install parakeet.cpp — clone OUTSIDE the Minutes repo
 mkdir -p ~/src && cd ~/src
 git clone --recursive https://github.com/Frikallo/parakeet.cpp
 cd parakeet.cpp
-make build
+
+# Configure with the maintainer's two opt-out flags (AXIOM_INSTALL=OFF,
+# PARAKEET_INSTALL=OFF). They short-circuit the install(EXPORT) export-set
+# strictness error and work on any CMake version. PARAKEET_BUILD_SERVER_EXAMPLE
+# turns on the warm-sidecar binary that live mode reuses across utterances.
+cmake -S . -B build -DCMAKE_BUILD_TYPE=Release \
+  -DAXIOM_INSTALL=OFF \
+  -DPARAKEET_INSTALL=OFF \
+  -DPARAKEET_BUILD_SERVER_EXAMPLE=ON \
+  -DAXIOM_BUILD_TESTS=OFF \
+  -DAXIOM_BUILD_EXAMPLES=OFF
+cmake --build build -j
+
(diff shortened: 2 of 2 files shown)


silverstein:
Thank you for this, @rymalia. The validation notes in the PR body are exactly the kind of detail that makes install docs trustworthy, especially for the Parakeet path where the source-build steps and Apple Silicon linking are easy to get subtly wrong.

I see this is still marked draft, so I am treating it as not ready for final review unless you say otherwise. Two bits from the maintainer side:

- The Vercel failure is just a preview-deploy authorization gate, not a docs problem.
- I see the MCP Server check failed and will take a closer look when this comes out of draft / after the native-capture patch release I am currently finishing.

Appreciate the careful field testing here. This is useful work, and I do not want it to look ignored just because the current release-blocker took the front of the queue.

rymalia:
Thanks for the feedback, and the reminder! It's in draft because I want to remove some of the extraneous changes that got included, and then got pulled into other stuff. I'll get this out of draft and submitted soon.

silverstein (approved):
Reviewed in depth, and verified your claims against the codebase and live endpoints rather than taking them on faith. Everything checks out:

- **`minutes setup --parakeet` behavior**: confirmed in `crates/cli/src/main.rs` (`cmd_setup_parakeet`). It installs the bundled Silero VAD weights, resolves the binary, and prints the manual recipe. It never downloads or converts the `.nemo`. So the old doc (and the old README line) describing an automatic ~1.2 GB model download was wrong, and your rewrite is the accurate one. Good catch.
- **`parakeet_fp16`**: real key. One note: it already defaults to `true` (`crates/core/src/config.rs`), so the added line in the config block is documentation of the default rather than a behavior change. Fine to keep for explicitness; just flagging so you know it's not load-bearing.
- **Homebrew dropping `cmake@3`**: confirmed (`brew info cmake@3` returns no formula). The Kitware tarball pivot is the right call, and the URL resolves.
- **Public `.nemo` URL**: confirmed curl-able with no auth (HF 302s to the CDN as you described).
- **BSD tar `--wildcards` gotcha and the generator-dependent binary layout `find`**: both correct and the kind of thing that only surfaces from actually running the install, which is exactly why this rewrite is valuable.

The maintainer-blessed `-DAXIOM_INSTALL=OFF -DPARAKEET_INSTALL=OFF` flags are a much better story than the old CMake-downgrade-only workaround, and keeping the 3.31.x path for the separate atomics error is the right structure.

Happy to merge as soon as you mark it ready for review. Thanks for the end-to-end validation work on this, it shows.

rymalia:
Thanks for the feedback. I'm cleaning up a few things and will have the update posted today for review.

silverstein:
Hey @rymalia, no pressure at all, just checking in since this has been sitting. The content here is genuinely valuable (the CMake 4.x atomics gotcha and the Kitware tarball workaround especially), and I would like to get it landed.

Two things have shifted since you opened it: main has moved several releases (it now conflicts and needs a rebase), and some of the v0.18.0-specific notes may need a quick freshness check against current behavior, since there has been work toward automating parts of the Parakeet setup. If you are still up for the cleanup you mentioned, that would be great. If you have moved on to other things, totally fine, just say the word and I am happy to rebase it, verify the steps against the current release, and land it with credit to you. Either way I do not want this good work to go stale.

silverstein:
I went ahead and took this over so it does not go stale, keeping it as your PR with your authorship intact (the commit is still yours). What I did:

- Rebased onto latest main (the branch had drifted several releases behind).
- Resolved one conflict where main had already updated the sidecar section: the warm example-server sidecar now auto-enables when the example-server copy resolves, so I kept that current wording and your valid `parakeet_fp16 = true` addition, and dropped the older manual `parakeet_sidecar_enabled = true` line.
- Freshness-checked the core claims against current behavior: the `.nemo` download + `convert_nemo.py` conversion is still manual (confirmed in the setup code), so your install steps are accurate. I just made the two "in v0.18.0" version qualifiers version-agnostic since the manual flow persists.

Everything else is your work, unchanged. Thanks for the careful gotcha documentation, the CMake 4.x atomics workaround especially is the kind of thing that saves people hours. Merging once CI is green.

rymalia:
That's cool. Sorry for sitting on that so long - I ended up scope-creeping it and want to shift some of the sections around in PARAKEET.md -- so it's probably for the best to pull it in this state to take care of the gotchas, and then I'll submit another that cleans up the duplication within that doc and some out of sync instructions within 'crates/cli/src/main.rs' too.

silverstein:
Perfect, that works. No apology needed, scope creep on docs is a sign you care about getting it right. Landing this now for the gotchas (it is queued to merge as soon as CI goes green), and I will look out for your follow-up on the in-doc duplication and the CLI's printed setup instructions in main.rs. Good call flagging that main.rs path, since the CLI's eprintln recipe is exactly the kind of thing that drifts out of sync with the doc. Thanks again for the careful work here.

PR #262, opened 2026-05-29
fix(windows): bound PTY reader-thread join so the app can't hang on quit

### LINKED ISSUE #261: Windows: app hangs on quit — PtyManager::kill_session blocks forever joining a ConPTY reader thread that never sees EOF
**Title:** Windows: app hangs on quit — `PtyManager::kill_session` blocks forever joining a ConPTY reader thread that never sees EOF

## Summary

On Windows 11, quitting Minutes (tray → Quit, window close, or any
`RunEvent::Exit`) hangs the process indefinitely — window hides but the
process never terminates, tray icon lingers, and the app shows the
grey "Not Responding" overlay. Task Manager → End Task is the only way
out. macOS is unaffected.

Root cause is a single unconditional `JoinHandle::join()` in
`tauri/src-tauri/src/pty.rs` that blocks forever because, on Windows
ConPTY, killing the PTY child does not deliver EOF to the master, so
the reader thread never exits.

## Environment

- Windows 11, WebView2 Runtime 148.0.3967.83
- Reproduced on stock release binaries **v0.18.0, v0.18.2, v0.18.3**
  (`minutes-desktop-windows-x64-setup.exe` from Releases)
- Not reproducible on macOS (any version)

## Repro

1. Install Minutes on Windows 11.
2. Launch from Start Menu; wait for the main window (the embedded
   assistant panel spawns a PTY session on startup).
3. Right-click the tray icon → Quit Minutes (or close the window, or
   any path that reaches `RunEvent::Exit`).

**Expected:** process terminates within ~1s.
**Actual:** window hides, process stays alive indefinitely, tray icon
remains, subsequent window interaction shows "Not Responding". Only
Task Manager kills it.

## Root cause

`tauri/src-tauri/src/pty.rs`, `kill_session` (current `main`):

```rust
pub fn kill_session(&mut self, session_id: &str) -> Option<PathBuf> {
    if let Some(mut session) = self.sessions.remove(session_id) {
        session.child.kill().ok();
        if let Some(handle) = session.reader_handle.take() {
            // Don't block forever waiting for reader thread
            let _ = handle.join();   // <-- blocks forever on Windows
        }
        Some(session.context_dir)
    } else {
        None
    }
}
```

The reader thread (same file, in `spawn`) loops on
`reader.read(&mut buf)` and only breaks on `Ok(0)` (EOF) or an error:

```rust
let reader_handle = std::thread::spawn(move || {
    let mut buf = [0u8; 4096];
    loop {
        match reader.read(&mut buf) {
            Ok(0) => { /* EOF → break */ }
            Ok(n) => { /* emit → continue */ }
            Err(_) => { /* break */ }
        }
    }
});
```

On **Windows ConPTY**, calling `child.kill()` terminates the child
process but the pseudoconsole keeps the master pipe open — the reader's
`read()` does **not** return `Ok(0)` or an error, it just blocks
waiting for more data. The reader thread never exits, so the
unconditional `handle.join()` in `kill_session` blocks forever.

On **macOS/Unix**, killing the child closes the slave PTY, the master
read returns `Ok(0)`, the reader thread breaks and exits, and `join()`
returns immediately — which is why this never reproduces there. The
comment ("Don't block forever waiting for reader thread") describes the
intended behavior but the code does the opposite on Windows.

Because every quit path runs `cleanup_before_process_exit` →
`PtyManager::kill_all` → `kill_session` (`main.rs` lines ~44, ~63,
`RunEvent::Exit`), the join wedges shutdown before the process-exit call
is ever reached.

## Diagnostic evidence

With per-step shutdown tracing added, the breadcrumbs stop cold at the
PTY kill and a 15s watchdog (added for the investigation) then
force-kills:

```
cleanup_before_process_exit: starting
cleanup_before_process_exit: killing PTY sessions    <-- last line
[ 15.0s of total silence ]
shutdown watchdog fired: forcing TerminateProcess
```

`shutting down parakeet sidecar` / `done` are never reached. An
independent-thread `TerminateProcess` does kill the process — confirming
the exit primitive is fine and the blocker is strictly the join.

## Proposed fix

Bound the reader-thread join so it can never wedge shutdown. The reader
holds no lock the app needs and dies with the process on exit:

```rust
if let Some(handle) = session.reader_handle.take() {
    // The reader blocks on a PTY master read. On Windows/ConPTY,
    // killing the child does NOT deliver EOF, so an unconditional
    // join() hangs forever. Wait briefly for a clean exit (macOS gets
    // EOF and finishes instantly), then detach.
    let deadline = std::time::Instant::now()
        + std::time::Duration::from_millis(300);
    while !handle.is_finished() && std::time::Instant::now() < deadline {
        std::thread::sleep(std::time::Duration::from_millis(5));
    }
    if handle.is_finished() {
        let _ = handle.join();
    }
    // else: detach — never block shutdown on a wedged ConPTY reader.
}
```

`JoinHandle::is_finished()` is stable since Rust 1.61. This preserves
the clean-join behavior on macOS (thread finishes immediately) and
caps Windows shutdown at the 300ms deadline. Verified on Windows 11:
the full shutdown sequence completes and the process exits in well
under a second, every time (5/5 clean quits; per-step shutdown trace
confirms cleanup runs to completion instead of wedging at the PTY
kill).

I've opened #262 with this fix against `main`.

mquinn614:
Fixes #261.

## Problem

On Windows 11, quitting Minutes (tray → Quit, window close, or any
`RunEvent::Exit`) hangs the process indefinitely — the window hides but
the process never terminates, the tray icon lingers, and the app shows
the "Not Responding" overlay. Task Manager → End Task is the only way
out. Reproduced on stock release binaries **v0.18.0, v0.18.2, v0.18.3**.
macOS is unaffected.

## Root cause

The unconditional `handle.join()` in `PtyManager::kill_session`
(`tauri/src-tauri/src/pty.rs`). The PTY reader thread loops on
`reader.read()` of the master and only breaks on EOF/error. On Windows
ConPTY, killing the child process does **not** deliver EOF to the master
— the pseudoconsole keeps the pipe open — so the read never returns, the
thread never exits, and the join blocks forever. Every quit path runs
`cleanup_before_process_exit` → `kill_all` → `kill_session`, so the join
wedges shutdown before the process-exit call is ever reached.

On macOS/Unix killing the child closes the slave PTY, the read returns
`Ok(0)`, the thread exits immediately, and the join is instant — which
is why this is Windows-only. The existing comment ("Don't block forever
waiting for reader thread") described the intent; the code did the
opposite.

## Fix

Bound the join: give the reader a 300 ms window to exit cleanly (it does
wherever EOF is delivered), then detach. The reader holds no lock the
app needs and dies with the process on exit, so detaching is safe and
shutdown can never wedge on it. macOS behavior is unchanged (the thread
finishes immediately, so the join stays instant).

## Verification

Windows 11, WebView2 Runtime 148.0.3967.83: quit now completes in well
under a second, **5/5 clean** across repeated cycles. A per-step
shutdown trace confirms cleanup runs to completion — the ~300 ms
reader-detach gap is visible between "killing PTY sessions" and the next
step — and the process exits normally instead of hanging at the PTY kill.

No change to behavior on macOS/Linux.

## Credit

Root cause isolated with a per-step shutdown-trace harness (the
breadcrumbs dead-ended at the PTY kill; a temporary watchdog
force-killed 15 s later, proving the blocker was the join, not the exit
primitive). That diagnostic harness was built with Codex/Cursor.


silverstein:
Textbook diagnosis, Quinn. The ConPTY no-EOF behavior is exactly it, and bounded-join-then-detach is the right shape. The only thing CI flagged was a one-line rustfmt nit on the `deadline` binding, so I pushed that fix straight to your branch (logic untouched) to get the Format check green. Merging as soon as CI finishes. Thanks again, this is your second sharp Windows fix and you're credited in the release notes.

PR #263, opened 2026-05-30
fix(ui): make the AI assistant optional — degrade gracefully when no agent CLI is installed

mquinn614:
## Problem

The Recall assistant panel spawns an external agent CLI (Claude Code by
default). On a machine that doesn't have one installed, two rough edges show up:

1. **On startup**, if the saved workspace state has the assistant panel
   expanded, the app auto-spawns the agent and the panel renders a
   `Could not start: 'claude' not found on PATH` error — even though the
   assistant is optional and the rest of the app (recording, etc.) works fine
   without it.
2. **Opening the assistant manually** surfaces the same raw spawn error.

For someone who just wants to record/transcribe and never set up an agent, this
reads as "the app is broken."

## Fix

Make the assistant degrade gracefully when no agent CLI is installed:

- **Startup** skips the assistant auto-open entirely when `cmd_list_agents`
  reports none installed — a fresh install gets a clean main window instead of
  an auto-expanded panel it can't use.
- **Manual open** shows a calm note —
  `AI assistant is optional; install Claude Code to enable it.` — instead of the
  raw `Could not start` error.

No behavior change for anyone who has an agent installed: the availability check
falls through to the normal path, and the new message only replaces the error
text on the specific "agent binary not found" failure.

## Scope

Frontend-only (`tauri/src/index.html`), two small hunks. No backend or config
changes.

## Testing

- With an agent installed: assistant auto-opens and spawns exactly as before.
- Without an agent: startup leaves the panel collapsed; opening the assistant
  manually shows the calm optional note instead of the error.

### DIFF
--- a/tauri/src/index.html
+++ b/tauri/src/index.html
@@ -10360,6 +10360,14 @@
               }
             });
           }
+        } else if (/not found on PATH|not found in PATH|not found in common install/i.test(msg)) {
+          // The assistant needs an external agent CLI (Claude Code by
+          // default). It's OPTIONAL — the rest of the app works without it —
+          // so degrade with a calm note instead of a scary "Could not start"
+          // error for users who don't have an agent installed.
+          console.info('Recall: assistant agent not installed (optional):', msg);
+          recallLoading.querySelector('span').textContent =
+            'AI assistant is optional; install Claude Code to enable it.';
         } else {
           console.error('Recall spawn:', e);
           recallLoading.querySelector('span').textContent = 'Could not start: ' + msg;
@@ -12421,6 +12429,18 @@
 
     // Restore panel state from previous session
     (async () => {
+      // The assistant auto-open below spawns the agent CLI (Claude Code by
+      // default). It's optional — skip the auto-open entirely when no agent
+      // is installed so a fresh install gets a clean main window instead of
+      // an auto-expanded panel it can't use. The assistant is still reachable
+      // manually (it shows the calm "optional" note then).
+      try {
+        const agents = await invoke('cmd_list_agents');
+        if (Array.isArray(agents) && agents.length === 0) {
+          return;
+        }
+      } catch (_) { /* if the check fails, fall through to normal restore */ }
+
       try {
         recallWorkspaceState = await invoke('cmd_get_recall_workspace_state');
       } catch (e) {
--- a/tauri/src/index.html
+++ b/tauri/src/index.html
@@ -12429,17 +12429,26 @@
 
     // Restore panel state from previous session
     (async () => {
-      // The assistant auto-open below spawns the agent CLI (Claude Code by
-      // default). It's optional — skip the auto-open entirely when no agent
-      // is installed so a fresh install gets a clean main window instead of
-      // an auto-expanded panel it can't use. The assistant is still reachable
-      // manually (it shows the calm "optional" note then).
+      // The assistant auto-open below spawns the *configured* agent CLI
+      // (settings.assistant.agent, default "claude"). It's optional — when
+      // that agent isn't installed, skip ONLY the auto-open so the panel
+      // doesn't pop up and immediately fail to spawn. Everything else in the
+      // restore (recallRatio, last open document, last open meeting) is
+      // untouched. The assistant is still reachable manually, where it shows
+      // the calm "optional" note.
+      //
+      // NOTE: gate on the *configured* agent specifically — cmd_list_agents
+      // also enumerates shells (bash/zsh) that exist on every machine, so a
+      // bare "any agent installed?" check would never report unavailable.
+      let assistantAvailable = true;
       try {
-        const agents = await invoke('cmd_list_agents');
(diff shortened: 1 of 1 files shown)


tauri/src/index.html
```diff
@@ -10360,6 +10360,14 @@ <h2>Welcome to Minutes</h2>
               }
             });
           }
+        } else if (/not found on PATH|not found in PATH|not found in common install/i.test(msg)) {
```
silverstein, line 10363: This branch is solid. The regex matches the real backend string (`commands.rs:7562` emits `'{}' not found on PATH or in common install dirs.`), `msg` is safely stringified just above, and `recallLoading.querySelector('span')` reuses the exact pattern the adjacent `else` branch already uses, so it adds no new null risk. This is the hunk that actually delivers the graceful degrade, and it's fine to ship on its own.

tauri/src/index.html
```diff
@@ -12421,6 +12429,18 @@ <h2>Welcome to Minutes</h2>
...
     // Restore panel state from previous session
     (async () => {
+      // The assistant auto-open below spawns the agent CLI (Claude Code by
+      // default). It's optional — skip the auto-open entirely when no agent
+      // is installed so a fresh install gets a clean main window instead of
+      // an auto-expanded panel it can't use. The assistant is still reachable
+      // manually (it shows the calm "optional" note then).
+      try {
+        const agents = await invoke('cmd_list_agents');
+        if (Array.isArray(agents) && agents.length === 0) {
```
silverstein: Two issues with this guard:

**1. It almost never fires.** `cmd_list_agents` enumerates `WELL_KNOWN_AGENTS = ["claude","codex","gemini","opencode","pi","bash","zsh"]` (`commands.rs:7654`). `bash` and `zsh` exist on basically every macOS/Linux box (zsh is the macOS default shell), so `agents.length === 0` is unreachable in practice. On a machine without Claude Code, this early return never happens, the panel still auto-opens, and it still tries to spawn `claude`. So the "fresh install gets a clean main window" goal isn't actually delivered here. What rescues the UX is the hunk 1 calm message, shown inside an auto-expanded panel.

**2. When it does fire, the `return` is too broad.** It sits before `cmd_get_recall_workspace_state` and skips restoration of `recallRatio`, `openArtifactPath` (last open document) and `currentMeetingPath` (last open meeting), not just the assistant auto-open. The comment says "skip the auto-open entirely," but it short-circuits the whole restore. That's a latent regression for exactly the agent-less user this PR targets (it only stays hidden today because of issue 1).

**Suggested fix:** the repo already has the right predicate at `index.html:8504` and `:9122` (`agents.find(a => a.name === currentCmd)`). Check the *configured* agent (`settings.assistant.agent`, default `"claude"` per `config.rs:695`) and gate only the two `openRecall('assistant')` call sites rather than returning early. Something like:

```js
let assistantAvailable = true;
try {
  const [agents, settings] = await Promise.all([
    invoke('cmd_list_agents'),
    invoke('cmd_get_settings'),
  ]);
  const configured = settings?.assistant?.agent || 'claude';
  assistantAvailable = Array.isArray(agents) && agents.some(a => a.name === configured);
} catch (_) { /* assume available, fall through */ }
```

Then guard the two auto-open sites with `assistantAvailable &&` and leave the ratio/meeting/artifact restore untouched. That delivers the actual clean-main-window behavior AND keeps last-meeting/last-document restore working for agent-less users.

silverstein (changes requested):
Thanks Mike, this is the right problem to fix and the graceful-degrade instinct is good. I reviewed both hunks against the source.

**Hunk 1** (the calm "AI assistant is optional" message): solid and ship-worthy on its own. Note inline.

**Hunk 2** (skip auto-open on a fresh install): it doesn't actually fire on a normal machine, and the early `return` skips more than the assistant. Details inline. Short version: `cmd_list_agents` counts `bash`/`zsh` as "agents," so the empty-list check is essentially unreachable, and the `return` is placed before the rest of session restore (last meeting / last open document). Suggested fix is to check the configured agent and gate only the auto-open call sites instead of returning early.

Requesting changes on hunk 2 only. Hunk 1 is good to go. Happy to push a patch for hunk 2 if it saves you a round trip.

mquinn614:
Thanks Mat, both spot on. Pushed a revision:

- Hunk 2 now gates on the **configured** agent (`settings.assistant.agent`, default `claude`) via `cmd_get_settings` + `agents.some(...)`, matching the existing predicate at `index.html:8508/8530`, so it actually fires when the configured agent isn't installed (the bash/zsh trap you flagged).
- Dropped the early `return`. Only the two `openRecall('assistant')` auto-open sites are gated now, so `recallRatio` / last-document / last-meeting restore stay intact.
- Hunk 1 unchanged.

Thanks for the careful review.

silverstein (approved):
Thanks @mquinn614. Clean fix, and gating on the configured agent (rather than a naive 'any agent on PATH' check, which cmd_list_agents would always satisfy via shells) is exactly the right call. Addresses the earlier feedback. Merging for v0.18.5.

PR #272, opened 2026-06-02
Fix Recovery Center dates and retry all

### LINKED ISSUE #271: Feature Request: Show meeting date in Recovery Center and add "Retry all"
## Summary

The **Recovery Center** in the desktop app is hard to scan and tedious to act on:

1. Each failed item is labeled with a long, confusing identifier (effectively the preserved file's name) instead of a human-readable **meeting/capture date**.
2. There's no way to **retry everything at once** — you have to click retry on each item individually.

Reported by the maintainer while dogfooding.

## Current behavior

Recovery Center items render `item.title`, which the backend derives from the failed file's stem (dashes → spaces) in `recovery_title()`:

```rust
// tauri/src-tauri/src/commands.rs:4017
fn recovery_title(path: &std::path::Path, fallback: &str) -> String {
    path.file_stem()
        .and_then(|stem| stem.to_str())
        .map(|stem| stem.replace('-', " "))
        ...
}
```

Preserved captures land in `failed-captures/` (and watched failures in `<watch>/failed/`) with timestamp/ID-style filenames, so the title shows up as a long opaque string rather than a date. The `RecoveryItem` struct carries no date field today, even though the scan already computes each file's modified time for sorting and then discards it:

```rust
// tauri/src-tauri/src/commands.rs:1395
pub struct RecoveryItem {
    pub kind: String,
    pub title: String,
    pub path: String,
    pub detail: String,
    pub retry_type: String,
}

// tauri/src-tauri/src/commands.rs:4051 — `modified` is computed for sort order, never surfaced to the UI
```

Retry is per-item only — the UI wires one button per card to `cmd_retry_recovery`:

```js
// tauri/src/index.html:6889
retry.textContent = item.retry_type === 'memo' ? 'Retry as Memo' : 'Retry as Meeting';
retry.addEventListener('click', async () => {
  await invoke('cmd_retry_recovery', { path: item.path, contentType: item.retry_type });
  ...
});
```

There is no header action in the Recovery Center overlay (`tauri/src/index.html:4686`) beyond **Close**.

## Proposed change

**1. Show the capture date instead of (or alongside) the cryptic title**
- Add a date/timestamp to `RecoveryItem` (surface the `modified` SystemTime that `scan_recovery_items` already reads), formatted to a friendly local date/time in the UI (reuse the existing `formatJobTimestamp()` helper used by the Jobs Center).
- Use the date as the primary label, keeping the underlying filename available as secondary detail (it's already shown via `item.path`).

**2. Add a "Retry all" action**
- Add a **Retry all** button to the Recovery Center header (next to **Close**).
- Back it with a bulk command that iterates the current recovery items and requeues each via the existing `cmd_retry_recovery` path (respecting each item's `retry_type`), then refreshes the panel and resumes status polling.
- Disable/hide the button when there are no recovery items.

## Acceptance criteria

- [ ] Each Recovery Center item displays a readable capture/meeting date (e.g. `Jun 2, 2026, 3:14 PM`) as its primary label instead of a long filename-derived string.
- [ ] The original file path remains visible as secondary detail for disambiguation.
- [ ] A **Retry all** control appears in the Recovery Center header and retries every listed item in one click, honoring each item's `retry_type` (meeting vs memo).
- [ ] After "Retry all", the panel refreshes and status polling reflects the requeued work.
- [ ] The control is hidden or disabled when the list is empty.
- [ ] `cargo fmt`, `cargo clippy --all --no-default-features -- -D warnings`, and `cargo test -p minutes-core --no-default-features --lib` pass.
- [ ] UI verified by click-test in `~/Applications/Minutes Dev.app` (built via `./scripts/install-dev-app.sh`).

## Implementation notes (non-binding)

- The Jobs Center (`jobs-overlay`) is a **separate** overlay from the Recovery Center (`recovery-overlay`); this request targets the Recovery Center. A prior plan (`PLAN.retry-failed-jobs.md`) deferred a "Retry All Failed" for the Jobs Center citing a missing header — the Recovery Center already has a header row to drop a button into.
- Date formatting helper `formatJobTimestamp()` already exists in `tauri/src/index.html` and can be reused.
- A bulk retry can be a thin Tauri command that loops over `scan_recovery_items` results server-side, or a client-side loop over the already-fetched `recoveryItems` calling `cmd_retry_recovery` per item — prefer whichever keeps error handling per-item so one failure doesn't abort the rest.

## References

- `tauri/src/index.html:4686` — Recovery Center overlay markup (header has only a Close button)
- `tauri/src/index.html:6837` — `loadRecoveryItems()` renders items; title is `item.title || 'Recovery item'`
- `tauri/src/index.html:6889` — per-item retry button wiring (`cmd_retry_recovery`)
- `tauri/src/index.html:6104` — `formatJobTimestamp()` (reusable date formatter)
- `tauri/src-tauri/src/commands.rs:1395` — `RecoveryItem` struct (no date field today)
- `tauri/src-tauri/src/commands.rs:4017` — `recovery_title()` (filename-stem-derived title)
- `tauri/src-tauri/src/commands.rs:4026` — `scan_recovery_items()` (computes `modified` time for sorting)
- `tauri/src-tauri/src/commands.rs:5962` — `cmd_recovery_items` command
- `tauri/src-tauri/src/commands.rs` — `cmd_retry_recovery` (per-item retry backend)

cathrynlavery:
## Summary
- show each Recovery Center item's modified/capture timestamp as the primary label while keeping the path visible
- add a Recovery Center `Retry all` action backed by a new Tauri command
- queue bulk recovery items via the background job system and move them out of recovery folders while preserving file extensions

Fixes #271.

## Verification
- `cargo fmt`
- `cargo test -p minutes-core --no-default-features --lib`
- `cargo test -p minutes-app --no-default-features recovery`
- `cargo clippy --all --no-default-features -- -D warnings`
- `git diff --check`
- Installed `~/Applications/Minutes Dev.app` with `MINUTES_BUILD_FEATURES=parakeet ./scripts/install-dev-app.sh --no-open`
- Click-tested Recovery Center in `Minutes Dev.app`: date-first item labels, visible file paths, and `Retry all (11)` rendered in the header

## Notes
- Direct push to `silverstein/minutes` was not available for the current GitHub account, so this PR is opened from the fork branch.
- The configured Developer ID signing identity was not present locally, so the dev app install used the script's ad-hoc signing fallback.
- The default `parakeet,metal` dev-app build failed locally on an unrelated `whisper_rs_sys` Metal link error; the successful UI verification build used `MINUTES_BUILD_FEATURES=parakeet`.


silverstein:
Thanks @cathrynlavery, this is a really nice Recovery Center upgrade. The date-first labels, Retry all, and the queue progress all land well, and the click-testing notes are appreciated.

Before merging I ran the Retry-all path through an adversarial review, since it moves people's failed-recording audio around, and it surfaced a data-safety risk worth fixing first. The version that moved each file into the jobs dir before enqueuing could strand a recording: a failed job leaves the file in the jobs dir, recovery scanning never looks there, so a crash mid-queue or a failed job could make a recoverable recording invisible. The move plus stem handling plus best-effort rollback also had a few silent-failure paths.

I pushed a hardening commit on top of yours that keeps your UI exactly as-is but changes the backend to enqueue each file IN PLACE, with no move:

- On success the worker already moves the source out next to the saved meeting (preserve_audio_alongside_output), so it stops being a recovery item.
- On failure the worker leaves it where it is, so it stays recoverable and reappears in the list.
- scan_recovery_items now hides items that already have an active job, so queued items still disappear from the list without the risky move; a failed job is terminal, so the file correctly comes back.
- It also stops Retry all from enqueuing dual-source .voice/.system sidecars as standalone jobs that race the primary, and blocks a double-process race with single retry.

Net effect: same UX, about 110 fewer lines, and no path that can lose a recording. Merging for the next release with your authorship intact. Thanks again for the feature.

Issue #274, opened 2026-06-05
Recall "End Session" button freezes app — requires force-kill (Windows 11)


JstG3tnbi:
## Bug Report

**Version:** 0.18.5
**OS:** Windows 11 Enterprise 25H2
**CPU:** Intel Core Ultra 7 258V (2.20 GHz)
**RAM:** 32.0 GB (31.6 GB usable)

## Steps to Reproduce

1. Open Minutes Recall
2. Click the **End Session** button

## Expected Behavior

Session closes cleanly.

## Actual Behavior

The application freezes and becomes unresponsive. The process must be force-killed — there is no recovery without terminating it.

## Additional Notes

No errors appear in `minutes logs --errors`. The `minutes events` log does not show any session-end event, suggesting the freeze occurs in the UI layer before any shutdown event is written.

silverstein:
Thanks for the detailed report, and for checking the logs and events first. That already points us at the desktop session-teardown path rather than the recording pipeline.

A few questions to pin it down, since we develop on macOS and this looks specific to Windows ConPTY teardown:

1. When it freezes, does it stay frozen until you force-kill it, or does it recover after ~10 to 20 seconds?
2. Which agent is running in the Recall panel when you hit End Session (Claude, Codex, Gemini, OpenCode, or a plain shell)?
3. Does the whole window go unresponsive, or just the Recall panel on the right?
4. Every time, or only sometimes?

We have a candidate fix that hardens how the pseudo-terminal is torn down on Windows (it currently holds an internal lock across the kill, and may not terminate the whole agent process tree). I am preparing a Windows test build and will follow up here with a download link so you can confirm it resolves this on your machine before the next release.

silverstein:
The Windows test build is ready, if you are able to try it before the next release:

https://github.com/silverstein/minutes/actions/runs/27032771726

Download the `minutes-desktop-windows-x64` artifact (you will need to be signed in to GitHub), unzip it, and run the setup .exe. It is unsigned, so Windows shows a SmartScreen warning: click More info, then Run anyway. Then open Recall and hit End Session: it should close cleanly with no freeze.

The questions above still help even if you cannot test the build. Thanks again.

silverstein:
Fixed and merged in #275 (will ship in the next release). A Windows contributor verified the End Session freeze is gone across a full test matrix: clean teardown, no force-kill, no orphaned processes, and rapid start/stop cycles with no leak.

[...]

Issue #258, opened 2026-05-27
Live transcript status reports 0 lines / 0:00 on Windows during a healthy in-app (Tauri) Live session


mquinn614:
# Bug: live transcript status reports 0 lines / 0:00 on Windows during a healthy in-app (Tauri) Live session

## Summary
On Windows, `live_transcript::session_status()` returns `line_count: 0` and
`duration_secs: 0.0` throughout an otherwise-healthy in-app (Tauri) live
transcript session — even while utterances are being captured, transcribed, and
written to `live-transcript.jsonl`. Any UI that drives a timer or "is it
hearing audio?" indicator off those fields shows a frozen 0:00 and can trip a
false "no audio" state. macOS is unaffected.

## Environment
- OS: Windows 10/11 (x64), WebView2
- Build: `cargo tauri build --features parakeet` (release desktop installer)
- Surface: in-app Tauri live transcript (the session runs on a background
  thread inside the app process), `TranscriptSource::Standalone`

## Symptoms
- The session works end to end: `events.jsonl` shows `recording.started`
  (`source:"live"`) followed by multiple `live.utterance.final` events, and
  `live-transcript.jsonl` contains the transcribed lines.
- But `cmd_live_transcript_status` / `session_status()` returns
  `line_count: 0`, `duration_secs: 0.0` for the whole session.
- (The UI still shows "recording" because that flag is tracked separately by
  the app, which is what makes the contradiction visible: clearly recording,
  yet 0 lines / 0:00.)

## Root cause (Windows file-locking)
`session_status()` → `derive_session_status()` gates the stats on
`standalone_active = lt_process_pid.is_some()`, where
`lt_process_pid = check_pid_file(&live_transcript_pid_path()).ok().flatten()`:

```rust
let should_report_stats = standalone_active || recording_pid.is_some();
let (line_count, duration_secs) = if should_report_stats {
    status_metrics(live_status.as_ref(), jsonl_path, now)
} else {
    (0, 0.0)   // <-- Windows lands here
};
```

`check_pid_file()` reads the PID file with `fs::read_to_string(path)?`:

```rust
let pid_str = fs::read_to_string(path)?;  // fails on Windows, see below
```

But the live session holds that same PID file under an **exclusive `fs2`
lock** for its entire lifetime (`create_pid_guard` → `file.try_lock_exclusive()`).

- On **Unix**, `fs2` uses `flock`, which is *advisory* — a separate
[...]

PR #275, opened 2026-06-05
fix(windows): release pty_manager lock before PTY teardown + kill process tree on End Session (#274)

silverstein:
## What
Recall "End Session" (`cmd_pty_kill`) freezes the desktop app on Windows: force-kill required, no session-end event written. Reported on v0.18.5 (#274).

## Why
`cmd_pty_kill` held the `pty_manager` mutex across `child.kill()` and the bounded reader wait. If ConPTY teardown blocks on Windows, the lock wedges every other PTY command (the xterm UI issues writes/resizes constantly) and the panel freezes. The #262 fix bounded the reader-thread join for the quit path, but did not address the lock-held-across-kill, nor the agent process tree.

## Changes (Windows behavior; Unix unchanged)
- **Drop the lock before teardown.** Lock only long enough to `take_session` (map removal), drop it, then run `child.kill()` plus the existing 300ms bounded-wait-then-detach outside the lock. Same for shutdown (`take_all_sessions`, drop, `kill_all`).
- **Kill the full process tree on Windows.** A `#[cfg(windows)]` best-effort, non-blocking `taskkill /T /F /PID` after `child.kill()`, since terminating the ConPTY leader leaves agent grandchildren holding the pipe open with no EOF.

## Testing
- macOS: `cargo fmt`, clippy (default and `--no-default-features`), `cargo build -p minutes-app` all pass.
- This path is Windows-ConPTY-specific and cannot be reproduced on macOS. Relying on CI's Windows build for the compile gate and on a Windows tester to confirm the runtime fix before closing #274.

Refs #274.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -7884,8 +7884,13 @@ pub fn cmd_pty_resize(
 
 #[tauri::command]
 pub fn cmd_pty_kill(state: tauri::State<AppState>, session_id: String) -> Result<(), String> {
-    let mut manager = state.pty_manager.lock().map_err(|_| "Lock failed")?;
-    manager.kill_session(&session_id);
+    let session = {
+        let mut manager = state.pty_manager.lock().map_err(|_| "Lock failed")?;
+        manager.take_session(&session_id)
+    };
+    if let Some(session) = session {
+        crate::pty::kill_session(session);
+    }
     Ok(())
 }
 
--- a/tauri/src-tauri/src/main.rs
+++ b/tauri/src-tauri/src/main.rs
@@ -39,11 +39,17 @@ static CLEAN_EXIT_STARTED: AtomicBool = AtomicBool::new(false);
 static MACOS_TERMINATE_APP_HANDLE: OnceLock<tauri::AppHandle> = OnceLock::new();
 
 fn cleanup_before_process_exit(app: &tauri::AppHandle) {
-    if let Some(state) = app.try_state::<commands::AppState>() {
-        if let Ok(mut mgr) = state.pty_manager.lock() {
-            mgr.kill_all();
-        }
-    }
+    let sessions = app
+        .try_state::<commands::AppState>()
+        .and_then(|state| {
+            state
+                .pty_manager
+                .lock()
+                .ok()
+                .map(|mut mgr| mgr.take_all_sessions())
+        })
+        .unwrap_or_default();
+    crate::pty::kill_all(sessions);
     minutes_core::parakeet_sidecar::shutdown_global_parakeet_sidecar();
 }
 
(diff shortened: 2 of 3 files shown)


silverstein:
@mquinn614, could use your Windows eyes here. This PR fixes #274: Recall "End Session" freezing the app on Windows (force-kill required). You're our most reliable Windows tester and you've already hit this corner (#258).

The fix hardens PTY teardown: it stops holding an internal lock across the kill, and on Windows it now kills the full agent process tree (taskkill /T), since killing the ConPTY leader alone leaves grandchildren holding the pipe open.

Unsigned Windows test build (download the `minutes-desktop-windows-x64` artifact, unzip, run the setup .exe; Windows shows a SmartScreen warning, click More info then Run anyway):
https://github.com/silverstein/minutes/actions/runs/27032771726

If End Session closes cleanly with no freeze, that is the green light to land this before the next release. Thanks!

mquinn614:
Thanks @silverstein. Took a close look at the diff and ran the Windows build, and this is a clean fix. The diagnosis matches what we ran into on the Madness side: [#274](https://github.com/mquinn614/minutes/issues/274) is two Windows failure modes stacking. The pty_manager mutex being held across kill_session (the child.kill() plus reader join) meant one blocked teardown froze every other command that needs the lock, so the app hangs before any shutdown event is written, which is exactly the "no error logs, no session-end event" symptom. And on Windows child.kill() only takes out the ConPTY leader, so the grandchildren (node plus the npx minutes-mcp processes) keep the pipe open and the reader never EOFs.

The take_session/take_all_sessions then kill-outside-the-lock split is the right fix for the freeze, and the bounded 300ms reader-join (detach if it doesn't finish) is a nice touch, since there's no hang even if the pipe stays open. 👍

Tested on Windows 11 against the minutes-desktop-windows-x64 setup artifact:

✅ End Session closes the embedded Claude session cleanly, no freeze, no force-kill
✅ App stays fully responsive after End Session (settings, recording, tray all work)
✅ Can start a fresh Claude session after ending one (session slot freed correctly)
✅ No orphaned node.exe / conhost.exe / minutes-mcp processes in Task Manager afterward
✅ End Session while Claude is mid-task ends cleanly
✅ 10 rapid start/stop cycles, no issues, no leak or eventual deadlock
✅ Quit Minutes from the taskbar right-click closes the app successfully
That's a green light from me to land it.

One minor, non-blocking note for the future: in kill_session, child.kill() runs before terminate_process_tree(). In theory, killing the ConPTY leader first can leave taskkill /T walking a stale PID, which would orphan the node/MCP grandchildren. My Task Manager check came back clean across every run, so it isn't biting in practice, but if orphaned processes ever show up on a different setup, swapping to run terminate_process_tree (it's /T /F, so it kills the leader too) before child.kill() would close that window.

silverstein:
Thanks for the thorough review and the full Windows test matrix, @mquinn614. The orphaned-process check across rapid cycles is exactly the signal we couldn't get from macOS.

Applied your ordering note: `terminate_process_tree` (taskkill /T /F) now runs before `child.kill()`, so the tree is walked from the live leader PID instead of a potentially reaped one (af8652e). Landing this for the next release. Credited you in the release notes.

PR #277, opened 2026-06-08
chore(deps): move cpal from git pin to crates.io 0.18.1 (#79)

silverstein:
## What
Moves cpal from a git rev pin to the crates.io 0.18.1 release. This removes the **last git dependency** in the workspace, which is the structural blocker for `cargo publish` (#79). pyannote-rs already moved to crates.io 0.3.4 (43bbef3), so after this, nothing blocks publishing.

## Why it's a clean swap
cpal 0.18.1 is the published release of the same PipeWire + realtime backend work we pinned to git rev `33e05ec`. The published feature surface matches our pin exactly (`default-features = false`, `features = ["pipewire", "realtime"]`, `default = []`). Lockfile now resolves cpal from `registry+...` with no git source, and pulls the matching transitive bumps (pipewire 0.9.2 to 0.10.0, libspa 0.9.2 to 0.10.0, nix dropped).

## Testing
- macOS (local): core / CLI / app build, 745 lib tests, 8 integration tests, `clippy --all --no-default-features` all green.
- Linux PipeWire backend and Windows WASAPI backend are exercised by CI on this PR; macOS does not compile those paths.

## Scope
Cargo.toml + Cargo.lock only. No code changes.

Refs #79.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -833,15 +833,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6245d59a3e82a7fc217c5828a6692dbc6dfb63a0c8c90495621f7b9d79704a0e"
 
 [[package]]
-name = "convert_case"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "baaaa0ecca5b51987b9423ccdc971514dd8b0bb7b4060b983d3664dad3f1f89f"
-dependencies = [
- "unicode-segmentation",
-]
-
-[[package]]
 name = "cookie"
 version = "0.18.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


silverstein:
Updated: pinned to crates.io **0.18.0**, not 0.18.1. The 0.18.1 release regressed the WASAPI/Windows backend (fails to compile against windows-core 0.61.x with sealed-trait `IUnknownImpl` errors). 0.18.0 is the version-matched release our git rev already was, so it keeps the known-good Windows build while still removing the git dependency. Re-running CI on 0.18.0.

silverstein:
Closing: pinning to crates.io cpal does not work right now. Both 0.18.0 and 0.18.1 fail the Windows build (WASAPI sealed-trait `IUnknownImpl` error) because the published manifest ranges `windows-core` separately from `windows`, so in this tree `windows` resolves to 0.61.3 while `windows-core` floats to 0.62.2. The git rev we pin does not have this problem. Keeping the git pin (works on all platforms); filing an upstream cpal issue. crates.io publishing of minutes-core/cli stays retired per docs/PRE-RELEASE-CHECKLIST.md, so this is not a release blocker.

PR #288, opened 2026-06-10
fix(summarize): use Stdio::null() for agents that use --file args (pi, opencode)

gtheys:
## Problem

Agents like `pi` and `opencode` use a prompt file (`--file` / `@file`) rather than reading from stdin. The summarization pipeline opened a piped stdin unconditionally and never wrote to or closed it, causing those agents to block indefinitely waiting for EOF — manifesting as a consistent 300s timeout on every summarization attempt.

## Root cause

In `summarize_with_agent_impl_timeout` (`crates/core/src/summarize.rs`), `Stdio::piped()` was used regardless of whether a `stdin_payload` existed. For `pi` and `opencode`, `stdin_payload` is `None` (they use `--file` args), so the child process sat waiting for input that never came.

## Fix

Use `Stdio::null()` when `stdin_payload` is `None`, so file-based agents receive a closed stdin immediately and proceed normally. Agents that do use stdin (`claude -p -`, `codex`, etc.) are unaffected.

## Testing

Verified with `pi` as `agent_command`: summarization completes in ~19s on a 21-minute meeting transcript instead of timing out at 300s.

### DIFF
--- a/crates/core/src/summarize.rs
+++ b/crates/core/src/summarize.rs
@@ -1083,9 +1083,17 @@ fn summarize_with_agent_impl_timeout(
     let invocation = prepare_agent_invocation(&agent_cmd, &prompt)?;
     let cleanup_path = invocation.cleanup_path.clone();
 
+    // AIDEV-NOTE: Use Stdio::null() when no stdin payload is needed (e.g. pi, opencode
+    // which use --file args). Keeping a piped stdin open without writing/closing it causes
+    // agents that read stdin to block indefinitely waiting for EOF.
+    let stdin_stdio = if invocation.stdin_payload.is_some() {
+        std::process::Stdio::piped()
+    } else {
+        std::process::Stdio::null()
+    };
     let mut child = std::process::Command::new(&invocation.cmd)
         .args(&invocation.args)
-        .stdin(std::process::Stdio::piped())
+        .stdin(stdin_stdio)
         .stdout(std::process::Stdio::piped())
         .stderr(std::process::Stdio::piped())
         .spawn()


silverstein (commented):
Excellent diagnosis, and verified on our side: `prepare_agent_invocation` returns `stdin_payload: None` for the `pi` and `opencode` arms, and in `summarize_with_agent_impl_timeout` the parent's write end of the pipe is only taken inside the `if let Some(prompt_bytes)` block, so with no payload it sits open in `child` for the entire wait loop. Any agent that reads stdin to EOF deadlocks until the timeout kills it. `Stdio::null()` is exactly the right idiom, and thanks for the clean AIDEV-NOTE (we have precedent for those in `capture.rs`).

One thing we spotted while verifying: the same pattern exists in two sibling functions in the same file, and both also route through `prepare_agent_invocation`, so `pi`/`opencode` users would hit the identical hang there:

- `run_title_refinement_via_agent` (~line 1926)
- `run_speaker_mapping_via_agent` (~line 2238, with a 120s timeout)

Would you like to extend the same one-line change to both in this PR so the fix lands coherently in one commit? Happy to merge as soon as that's pushed. If you'd rather keep this PR as-is, that's fine too, and we'll do the siblings in a follow-up with a pointer back here. Your call.

CI note: everything real is green; the red "Vercel" check is just fork-deploy authorization noise and not your problem.

silverstein (approved):
Verified root cause and fix (details in earlier comment). Merging as-is; we'll extend the same fix to the two sibling call sites in an immediate follow-up so you don't have to wait on us. Thank you for the excellent diagnosis!

silverstein:
Merged, thank you again! Follow-up is up at #289: same one-line fix applied to the two sibling spawn sites (`run_title_refinement_via_agent` and `run_speaker_mapping_via_agent`), plus Unix regression tests built on your diagnosis (a fake stdin-draining `pi` agent; verified the tests hang against the pre-fix code and complete in ~0.2s with it). Your fix ships in the next release, and you'll be credited in the release notes.

Issue #318, opened 2026-06-12
feat(import): native Granola import via the API (drop the fragile pre-export step)


silverstein:
## Problem

`minutes import granola` reads pre-exported markdown from `~/.granola-archivist/output/`. That directory has to be populated by a separate Granola export tool first, and the common approach (scraping Granola's local cache) broke when recent Granola versions encrypted the cache (`cache-v6.json.enc`). So the import depends on a fragile external step that no longer reliably works.

(The `import_granola` code itself is fine. It parses exported markdown, not the cache. This is about removing the brittle pre-export dependency.)

## Proposal

Add a native import that reads Granola's data via its API, the same source the community [granola-cli](https://github.com/magarcia/granola-cli) / [granola-to-minutes](https://github.com/calvindotsg/granola-to-minutes) tools use, so `minutes import granola` works without a manual pre-export and is resilient to Granola's local cache format/encryption changes.

Sketch:
- `minutes import granola --via-api` (or auto-detect when no export dir is present): pull meetings from Granola's API and reuse the existing markdown writer in `import_granola` (`crates/cli/src/main.rs`).
- Keep the directory-based import as a fallback for offline archives.
- Document clearly that Granola's API is unofficial and may change without notice (as the README already notes for the granola-cli route).

## Context

Surfaced while clarifying the import docs. Worth scoping before building.

PR #319, opened 2026-06-12
docs(import): clarify how to populate the Granola export dir

silverstein:
## What

Clarifies the **Switching from Granola?** import flow. Docs and error message only, no behavior change to the importer.

- **README**: adds a note that `~/.granola-archivist/output/` must be populated by a separate Granola export tool first, and that recent Granola versions encrypt their local cache (`cache-v6.json.enc`), which breaks cache-scraping exporters. Points users to the API-based [`granola-to-minutes`](https://github.com/calvindotsg/granola-to-minutes) route already documented just below.
- **Error message**: the "directory not found" error from `minutes import granola` now names the working granola-cli route instead of a vague "export first".

## Why

`import_granola` is fine. It parses exported markdown, not Granola's cache, so it is unaffected by the encryption change. But the README never explained how to produce the export directory, and the common cache-scraping approach silently stopped working when Granola encrypted its cache. Users who hit `Granola export directory not found` had no actionable next step.

## Follow-up

A native Granola-API import path (so import does not depend on a fragile pre-export step) is filed separately: #318.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,8 @@ minutes import granola              # Import all meetings to ~/meetings/
 
 Reads from `~/.granola-archivist/output/`. Meetings are converted to Minutes' markdown format with YAML frontmatter. Duplicates are skipped automatically. All your data stays local — no cloud, no $18/mo.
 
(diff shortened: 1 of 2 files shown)


silverstein:
Heads up: an earlier draft of this change accidentally rode into main via #316 (squash 0b5e604), so the README note and the import_granola error message are already on main. The only remaining delta in this PR is the improved wording of one README line (period instead of the dash). Folding that one-line fix into the upcoming 0.18.10 release PR, after which this can close as superseded. Sorry for the churn.

silverstein:
Merging worked out cleaner than the supersede route: verified main now carries exactly one copy of the note with the improved wording. Nothing further needed.

Issue #312, opened 2026-06-12
Windows 11: app menu bar renders transparent — menu labels invisible over light backgrounds


mquinn614:
## Summary
On Windows 11 the in-window menu bar (File / Edit / Window / Help) renders with
an unpainted, transparent background and white item text. Whatever is behind
the Minutes window shows *through* the menu strip — desktop wallpaper, or
another app's window. Over dark content the labels happen to be readable; move
the window in front of anything light (e.g. Gmail in a browser) and the menu
becomes literally invisible. The menus still work when clicked — this is purely
a rendering problem, but it makes Check for Updates / Help / etc.
undiscoverable.

Screenshots will be attached in the comments: one over a desktop image
(wallpaper visible through the menu strip), one over a white Gmail window
(labels fully invisible, the other window's content bleeding through the
strip).

## Environment
- Windows 11, system dark theme, WebView2 (NSIS desktop build, 0.18.x)
- macOS unaffected (the same menu lives in the system menu bar there)
- Referenced against `main` @ `37669b9`

## Where it comes from
The app menu is built in `tauri/src-tauri/src/main.rs` (`build_app_menu`,
~L715) and attached app-wide via `.menu(build_app_menu)` (~L1314). On macOS
that menu goes to the system menu bar; on Windows, Tauri/muda renders it as a
native Win32 menu bar inside the window — and Win32 menu bars can't follow the
app's dark theme. This is a known toolkit limitation:
[tauri-apps/muda#97 "Menu bar doesn't follow dark/light theme on Windows"](https://github.com/tauri-apps/muda/issues/97).
In Minutes' case the strip ends up not just wrong-themed but unpainted, so it
composites transparent under DWM with white menu text on top.

## Repro
1. Install the Windows desktop build on Windows 11 with dark theme.
2. Open the main window — File / Edit / Window / Help show as white text on a
   see-through strip.
3. Drag the window so the menu row overlaps any white window behind it
   (browser, Notepad). The labels disappear entirely.

## Suggested fixes (options, cheapest first)
1. **Drop the native menu on Windows/Linux** and keep the in-app UI as the
   control surface:
   ```rust
   #[cfg(not(target_os = "macos"))]
   let _ = window.remove_menu();
   ```
   Items without an in-app equivalent (Check for Updates, the Help links)
   would need a small home in the UI — e.g. the existing settings/gear surface.
   (We run this in a fork's single-window mode and it sidesteps the issue
   completely.)
[...]

PR #328, opened 2026-06-15
fix: avoid dev app WebKit frame crash

calebchongc:
## Summary
- Keep the long-lived main Tauri WebView opaque and skip macOS vibrancy to reduce WebKit frame-update crashes when the app is hidden/shown from the tray.
- Stop upcoming-meeting calendar polls from mutating hidden main-window UI, and gate Recall auto-expand on the main window being visible and focused before it resizes the WebView.
- Add regression guards for both the main-window opacity/vibrancy settings and the upcoming-meeting auto-expand visibility/focus contract.
- Make the markdown rename collision test timezone-safe after it failed under Asia/Singapore local time.

## Testing
- cargo fmt --all -- --check
- cargo test -p minutes-core --no-default-features --lib
- cargo test -p minutes-app --no-default-features
- ./scripts/install-dev-app.sh
- Relaunched ~/Applications/Minutes Dev.app, hid the main window, waited 75s for the upcoming-meeting poll interval, and confirmed PID stayed alive with no new minutes-app .ips after the 2026-06-15 14:27:23 pre-fix crash report.

### DIFF
--- a/crates/core/src/markdown.rs
+++ b/crates/core/src/markdown.rs
@@ -1772,23 +1772,28 @@ mod tests {
     #[test]
     fn rename_meeting_resolves_slug_collision() {
         let dir = TempDir::new().unwrap();
+        let frontmatter =
+            "title: \"Call\"\ntype: meeting\ndate: 2026-04-07T10:00:00-07:00\nduration: 0\n";
         let path = write_meeting(
             &dir,
             "2026-04-07-call.md",
-            "title: \"Call\"\ntype: meeting\ndate: 2026-04-07T10:00:00-07:00\nduration: 0\n",
+            frontmatter,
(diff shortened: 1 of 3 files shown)


silverstein:
Merged. This is your second strong contribution this week, and exactly the kind I want to see: a real crash, a minimal fix, a regression guard that documents the why, and a bonus timezone-safe test fix I didn't even ask for.

One thing worth flagging: making the main window opaque is also the change I'd suspected would fix #312 (Windows 11 menu bar rendering transparent). That one is Windows-specific and I can't verify it from here, so I've asked the reporter to confirm on his machine rather than auto-closing it on a guess. If it's confirmed fixed, your PR gets the credit there too.

Thanks Caleb.

PR #329, opened 2026-06-15
fix: avoid upcoming meeting hidden WebView crash

calebchongc:
## Summary
- Skip upcoming-meeting UI refreshes while the main window is hidden, so the calendar poll does not mutate a tray-hidden WebView.
- Gate Recall auto-expand on the main window being both visible and focused before it resizes/repositions the WebView.
- Add a frontend contract test for the upcoming-meeting prep auto-expand guard.

## Testing
- cargo fmt --all -- --check
- cargo test -p minutes-app --no-default-features
- ./scripts/install-dev-app.sh
- Relaunched ~/Applications/Minutes Dev.app, hid the main window, waited 75s for the upcoming-meeting poll interval, and confirmed PID stayed alive with no new minutes-app .ips after the 2026-06-15 14:27:23 pre-fix crash report.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -9308,6 +9308,34 @@ mod tests {
     }
 
     #[test]
+    fn upcoming_meeting_auto_expand_requires_visible_focused_main_window() {
+        let manifest = env!("CARGO_MANIFEST_DIR"); // .../tauri/src-tauri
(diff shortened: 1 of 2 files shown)


calebchongc:
Closing while I investigate additional local crash reports. I will follow up with a cleaner fix from my fork once the dismiss/popup crash is understood.

Issue #244, opened 2026-05-13
call_detection: false positives when dictation tools (superwhisper, etc.) are the active mic consumer


silverstein:
## Summary

The call_detection loop currently fires whenever a configured app is running AND the mic is active. It does not distinguish between a real call (Zoom/Teams/etc. holding the mic) and a short-burst dictation tool (superwhisper, Wisprflow, etc.) that grabs the mic for a few seconds at a time during normal workflow.

For users who run dictation tools throughout the day, this causes the call-detect banner to fire on every utterance, which is interruptive.

## Reported by

@jacobSingh in #224:

> It decides you're in a meeting and steals focus every time I speak. Since I use superwhisper all day, this doesn't work obviously.

## Current workaround

\`\`\`toml
[call_detection]
enabled = false
\`\`\`

Silences the loop entirely, but the user loses the actual feature for real calls.

## Why this happens

\`tauri/src-tauri/src/call_detect.rs\` polls for two signals together: (a) a configured call app is in the running-process list, (b) mic is in use. Both signals together fire the detection. But on macOS the mic-in-use signal is process-agnostic. A short-lived dictation grab from another app reads as "mic active" regardless of who is actually consuming it.

## Possible fixes

1. **Process-specific mic check**: tie the mic-in-use signal to the configured call app, not generic mic activity. Requires walking per-process audio input state (HAL or coreaudio API).
2. **Mic-active duration gate**: require mic to be continuously active for N seconds before firing the banner. Filters out short dictation bursts.
3. **Configurable allowlist of "ignore mic activity if these apps are foreground"**: explicit user opt-out for dictation tools.

(2) is the simplest and cheapest. (1) is the correct fix.

## Acceptance

When superwhisper is running and the user dictates throughout the day, the call_detect banner should not fire. When the same user starts a real Zoom call, the banner should fire as today.

adriannoes:
Came back to this one after the Safari probe fix (#858), since it is the same detector. Rechecking the current code, half of this issue is already fixed, and what remains is scoped to the browser path.

[...]

Issue #324, opened 2026-06-12
macOS app bundles ship a placeholder CLI sidecar; in-app Set up CLI installs a dead stub


silverstein:
Since v0.16.2, every CI-built macOS bundle (DMG and updater artifact) contains a 52-byte shell placeholder at `Contents/MacOS/minutes` instead of the real CLI. Any user who runs the in-app **Set up CLI** flow gets a symlink to that stub, and every `minutes` invocation prints `minutes CLI placeholder` and exits 1.

**Root cause.** `tauri/src-tauri/build.rs` stages the CLI sidecar only if a release binary already exists at `target/release/minutes`; otherwise it writes a placeholder and emits a cargo *warning*. The placeholder exists so plain `cargo check`/`clippy` work without a CLI build, on the assumption that real bundling always goes through `scripts/build.sh` (which builds the CLI first). The release workflow calls `cargo tauri build` directly, so the assumption silently fails on every release and the warning disappears into CI logs.

**Why it was never caught.** Local builds always go through `build.sh` or `install-dev-app.sh`, which build the CLI first, so every bundle a maintainer ever launched had the real binary. Only CI-built artifacts are affected. Confirmed by inspecting the v0.18.10 production bundle delivered by the auto-updater.

**Scope.** macOS only: the sidecar staging early-returns on other platforms, and `externalBin` is declared only in `tauri.macos.conf.json`. Homebrew and `cargo install` users are unaffected. Audited the rest of the packaging surface for the same class: the Swift helpers (`system_audio_record`, `calendar-events`) and the skill-bundle staging all hard-fail on error, and `pack_mcpb.sh` validates its inputs; this was the only silent fallback.

**Fix.**
1. Build the release CLI in `release-macos.yml` before `cargo tauri build`.
2. Make `build.rs` hard-fail instead of writing a placeholder when `MINUTES_REQUIRE_REAL_CLI_SIDECAR=1` is set, and set it in the release workflow, so this class of bug fails the build instead of shipping.
3. Assert the bundled sidecar is a real multi-MB binary after the build as a belt-and-braces check.

Shipping as 0.18.11.

silverstein:
[...]

PR #331, opened 2026-06-15
fix: detect Zoom calls using non-default input devices

### LINKED ISSUE #330: macOS call detection misses Zoom when Zoom uses a non-default input device
## Problem

On macOS, the desktop call detector can miss an active Zoom call when Zoom is using an input route that is not the current system default input device.

The current detector requires two signals before showing the **Call detected** banner:

1. a configured call app process is running (`zoom.us`, Teams, Webex, etc.)
2. the mic gate reports active input

The mic gate was effectively default-input-only: `mic_check` queried CoreAudio activity on `kAudioHardwarePropertyDefaultInputDevice`. If Zoom captured from a different input route, macOS still showed Zoom using the microphone, but Minutes logged `mic_gate_inactive` and never showed the banner.

## Repro

1. Start the macOS desktop app with call detection enabled and `zoom.us` in `[call_detection].apps`.
2. Start a Zoom call.
3. Set Zoom's microphone/input to a device that is not the current macOS default input.
4. Confirm macOS shows Zoom is using the microphone.
5. Observe Minutes does not show the **Call detected** banner and logs `mic_gate_inactive`.
6. Switch the macOS default input to the same input Zoom is using.
7. The detector flips to active and the Zoom banner appears.

## Expected

A Zoom call should be detected when Zoom is actively capturing input, regardless of whether Zoom's selected input is the current macOS default input.

## Important false-positive constraint

This overlaps with #244. A naive fix that changes the gate from "default input is active" to "any input device is active" can make #244 worse:

- idle Zoom is running in the background;
- a dictation app such as superwhisper/Wisprflow grabs a non-default mic;
- generic all-device mic activity becomes true;
- the detector incorrectly fires because `zoom.us` is also running.

So the fix should not simply make generic mic activity broader.

## Appropriate fix

Use CoreAudio process-level input activity where available:

- have the helper expose active input PIDs using `kAudioHardwarePropertyProcessObjectList`, `kAudioProcessPropertyIsRunningInput`, and `kAudioProcessPropertyPID`;
- have the detector preserve PID/PPID/name in the running-process snapshot;
- treat a native call app as active only when the active input PID belongs to that configured app's process family;
- for Zoom, include known active meeting/helper processes such as `ZoomHybridConf` in that process family;
- keep all-device input scanning only as a helper fallback for systems/API states where process attribution is unavailable.

## Acceptance

- Zoom call using a non-default input still shows the **Call detected** banner.
- Idle Zoom plus another app using the mic does not trigger the Zoom banner.
- Existing Google Meet / Teams-web browser detection behavior is unchanged.
- The bundled `mic_check` helper is rebuilt from source during app builds so packaged apps do not ship a stale helper binary.

calebchongc:
## Summary

Fixes #330.

This makes macOS native call detection work when Zoom is actively using a non-default input device, without broadening the generic mic gate in a way that would worsen #244.

Changes:
- rebuild and bundle `mic_check` from `mic_check.swift` during Tauri builds;
- update `mic_check` to expose CoreAudio active-input PIDs via `--active-input-pids`;
- keep the all-input-device scan only as a helper fallback when process attribution is unavailable;
- preserve PID/PPID/name in the detector's process snapshot;
- require native call-app detection to match active input to the configured app's process family;
- include Zoom meeting/helper processes such as `ZoomHybridConf` in Zoom's attributed process family.

## Why this is not a #244 regression

#244 happens because generic mic activity can be caused by another app, such as a dictation tool, while an idle call app is merely running.

This PR avoids using broadened generic mic activity as the native app trigger when process attribution is available. The detector now checks whether the active input PID belongs to the configured native call app process family. A regression test covers idle `zoom.us` plus active `superwhisper` and expects no detection.

## Tests

- `cargo test -p minutes-app -- --test-threads=1` — 211 passed
- `xcrun swiftc tauri/src-tauri/src/mic_check.swift -o /tmp/minutes-mic-check-test && /tmp/minutes-mic-check-test && /tmp/minutes-mic-check-test --active-input-pids`
- `INSTALL_DIR=/Applications ./scripts/install-dev-app.sh --no-open`
- `codesign --verify --strict --verbose=2 /Applications/Minutes\ Dev.app`

## Scope

This branch is intentionally limited to macOS call detection and the bundled `mic_check` helper. It does not include MLX, Sherpa ONNX, transcription backend, or dependency changes.


silverstein:
This is excellent work, Caleb. Filing #330 with a clean repro and then fixing it with CoreAudio active-input-PID attribution is exactly the right approach, and the descendant-expansion handling for Zoom helpers is thoughtful. I want to merge it. One blocker and two smaller things first.

**Blocker: the attribution-unavailable fallback reopens #244.**

In `detect_active_call` (call_detect.rs:766), `mic_live` is computed as:

```rust
let mic_live = active_input_pids.as_ref().map(|p| !p.is_empty()).unwrap_or_else(is_mic_in_use);
```

and the native-app loops fall back with:

```rust
let native_active = match (processes, active_input_pids) {
    (Some(processes), Some(active_input_pids)) => native_app_has_active_input(...),
    _ => native_app_matches_running_process(config_app, running),
};
```

When `active_input_process_pids()` returns `None`, and `mic_check.swift` returns exit 2 whenever CoreAudio process attribution is unavailable (older macOS, a permission/sandbox failure, or a missing helper), both lines collapse back to the pre-#244 behavior: generic mic-in-use plus 'zoom.us is running' attributes the call to Zoom. So an idle Zoom plus an active dictation tool re-triggers the exact false positive #330's fix is meant to prevent, on any machine where attribution is not available.

Suggested fix: once attribution is unavailable, native-app detection should not fall back to 'process exists plus generic mic active.' Either return no detection for native apps until PIDs are available (degraded, logged), or keep the all-device mic scan strictly as a liveness gate for the browser-tab probe path, never as the native trigger. Concretely, the `_ =>` arms at lines 909 and 925 should not detect without `Some(active_input_pids)`.

**Two smaller items:**

- **Test gap:** `idle_zoom_does_not_detect_when_another_app_has_input` only covers `active_input_pids = Some({superwhisper})`. The path that actually reopens #244 is `active_input_pids = None`, which is not tested. A case with idle zoom.us plus active superwhisper plus `active_input_pids = None` expecting no detection would have caught the above.
- **Helper scoping (minor):** the Zoom helper names (ZoomHybridConf/CptHost/caphost/aomhost) are matched by exact name without requiring a zoom.us root or a descendant link, so a stale or identically-named helper holding the mic is accepted as Zoom. Rooting helpers in a matched zoom.us process (or bundle path) would tighten it.
- **CI is red** only on a clippy `too_many_arguments` for `detect_active_call_from_snapshot_with_processes` (call_detect.rs:836), a one-liner once the rest settles (`#[allow]` or bundle the args into a struct).

Take whatever you have learned from the local crash reports into account too. Happy to talk through the fallback approach if useful. I think closing that one path is all this needs.

calebchongc:
I'll honestly attribute 100% of the credit for the fixes to codex and GPT, I'm just trying to use the app and throwing codex at any bugs I find while I'm at it to try and use the great framework you've built here 😂.

silverstein:
Ha, all good. Credit can land wherever it likes. The only bar here is that the code is correct, and the review exists exactly so it doesn't matter whether a fix came from you, Codex, GPT, or a Ouija board. Funny enough the blocker above came out of a Codex adversarial pass on my end, so it was robots checking robots, and that is fine by me. You are finding real bugs in a tool you actually use and reporting them with clean repros, which is the most useful kind of contribution there is. Keep them coming.

calebchongc:
Added two follow-up commits to address the review feedback:

- `fe63224` removes the unsafe fallback where native detection could use “process is running + generic mic active” when input attribution is unavailable. Native apps now require attributed input PIDs, which should close the #244/#330 false-positive path.
- The same commit also tightens Zoom helper handling by relying on descendants of the matched Zoom process instead of accepting helper names standalone, and fixes the clippy failure.
- `b3b49d8` adds the missing regression coverage for `active_input_pids = None` after a browser probe returns no match, expecting no native detection.

So I believe the blocker and the test gap from the review are now covered.

silverstein:
This nails it. I verified the fix end to end:

- The fail-open is gone: the native-app fallback is now `_ => false` instead of "process running + generic mic active," so when input attribution is unavailable, native detection simply does not fire. Fail-closed is the right call here (missing a detection beats a false one), and it pairs cleanly with the record-start permission gate that just landed on main.
- The unrooted-helper case is tightened and has its own test (`unrooted_zoom_helper_input_does_not_detect_zoom_call`), and the `active_input_pids = None` regression test (`idle_zoom_does_not_detect_when_active_input_attribution_is_unavailable`) covers exactly the path that reopened #244.
- Clippy and the full test matrix are green.

One housekeeping thing before I merge: main moved a fair bit today (a couple of releases plus a microphone permission gate), and your branch now shows conflicts. The likely spot is `build.rs`, since your mic_check compile step overlaps with the CLI-sidecar build logic that landed in #324. Could you rebase on latest main and resolve? Once it is green again I will merge it. Thanks for turning this around so fast.

calebchongc:
Rebased this onto latest `main` and resolved the only conflict, which was the test-module overlap from the meeting-prompt fix already merged upstream.

Local checks are green:

- `cargo fmt --check`
- `cargo clippy -p minutes-app --all-targets -- -D warnings`
- `cargo test -p minutes-app` (`214 passed`)

GitHub now reports the branch as mergeable. CI has restarted; Vercel still appears to be the external authorization gate.

silverstein:
Merged, and #330 closed with it. Nice work, Caleb. This one had the most moving parts of anything you have sent: a real bug you hit and reproduced, a CoreAudio attribution fix, a regression it almost reintroduced, and then a clean fail-closed correction with tests covering both the attribution-unavailable and unrooted-helper paths. That is exactly how a fix in a tricky area should land. It will go out in the next release. Thanks for the fast turnarounds.

Issue #332, opened 2026-06-15
Question: direction for trying alternative transcription backends


calebchongc:
@silverstein 
I’ve been experimenting with a couple of alternative local transcription backends, mostly as POCs to see whether newer/different model runtimes could work with Minutes. I’m not a developer looking to maintain these long term, but I wanted to share the experiments and ask what direction would be useful upstream.

POC branches:

mlx
- [audio-batch](https://github.com/calebchongc/minutes/tree/mlx-audio-batch)
- [live-dictation](https://github.com/calebchongc/minutes/tree/mlx-audio-live-dictation)
sherpa-onnx
[- upstream/sherpa-onnx-engine](https://github.com/calebchongc/minutes/tree/upstream/sherpa-onnx-engine)

The broader question: does it make sense for Minutes to have a clearer backend/adapter layer for trying engines like Sherpa ONNX or MLX-based audio tooling, or would you prefer model integrations to stay direct and tightly controlled in the core pipeline?

I imagine there are tradeoffs around packaging, timestamp quality, live/dictation support, VAD/diarization integration, cleanup, and maintenance burden. Hopefully the branches can serve as exploratory references for future work.

silverstein:
Thanks for doing this exploration, Caleb, and for being upfront that you are not looking to maintain these long term. That actually clarifies the direction rather than complicating it.

Here is where I land: of the two, **Sherpa ONNX is the one worth pursuing as a real engine**, and MLX stays a valued reference POC.

The reason is the thing that has shaped every engine decision in Minutes so far. The pipeline deliberately avoids Python at runtime: pyannote-rs instead of pyannote, parakeet.cpp instead of Python parakeet. It was never about not being able to write Python, it is about not putting a venv plus pip on every user's machine, because that becomes the dominant support surface (broken interpreter, pip resolution failures, upstream package drift) and it lands at runtime on the user, not in our code. Your MLX branch is clean work, but it reintroduces exactly that.

Your Sherpa branch sidesteps all of it. It is a native Cargo feature (`--features sherpa-onnx`) loading ONNX model files directly, no Python anywhere, the same shape as the existing `diarize` and `parakeet` features. And it reaches the same destination MLX was aiming at: the default `parakeet-tdt-0.6b-v3-int8` model is the multilingual lane we actually have demand for, and unlike the batch-only MLX work it also covers live and dictation. That is the engine I would carry.

[...]

PR #334, opened 2026-06-15
Fix meeting prompt dismissal crash

### LINKED ISSUE #333: Dev app can crash when upcoming meeting prompt is dismissed
## Problem

The dev app can crash when the floating upcoming-meeting prompt is dismissed. This is a separate crash path from the already-merged WebKit frame lifecycle fix in #328 and from the call-detection work in #331.

## Repro / observed behavior

1. Launch the installed dev app.
2. Let Calendar surface an upcoming meeting prompt.
3. Dismiss the prompt via Dismiss, the close button, Escape, or the success-close path after starting a recording.
4. The app can crash around the prompt webview close lifecycle.

## Expected behavior

Dismissing the upcoming-meeting prompt should tear down the prompt window without taking down the dev app.

## Proposed fix

Route prompt dismissal through a native Tauri command that destroys the `meeting-prompt` window, and make the prompt close path idempotent so repeated/near-simultaneous close triggers do not race each other.

## Verification notes

- Built and installed `Minutes Dev.app` locally.
- Exercised the meeting-prompt dismissal paths and confirmed dismissing no longer crashes the app.
- Ran Rust formatting, clippy, and the full `minutes-app` test package.
- Could not fully exercise the `Join & Record` calendar URL path on this managed Mac because the bundled `calendar-events` helper continued to report write-only Calendar access even though System Settings showed Full Access, so direct helper reads failed with `access not granted`.

calebchongc:
## Summary

- Close the `meeting-prompt` window through a native `cmd_close_meeting_prompt` command that destroys the webview window.
- Make prompt dismissal idempotent so Dismiss, X, Escape, success auto-close, and auto-dismiss all share the same close path.
- Add a regression test covering that the meeting prompt dismissal flow uses the native destroy path.

Fixes #333.

## Scope

This branch is intentionally limited to the meeting-prompt crash fix and excludes the call-detection changes from #331.

## Tests

- `cargo fmt --check`
- `cargo clippy -p minutes-app --all-targets -- -D warnings`
- `cargo test -p minutes-app` (`207 passed`)
- `./scripts/install-dev-app.sh --no-open`
- Manual: built/installed `Minutes Dev.app`, exercised meeting-prompt dismissal paths, and confirmed dismissing no longer crashes the app.

## Manual test caveat

I could not fully exercise the `Join & Record` calendar URL path on this managed Mac. The bundled `calendar-events` helper continued to report write-only Calendar access even though System Settings showed Full Access, so direct helper reads failed with `access not granted`.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -1016,6 +1016,14 @@ pub fn cmd_get_meeting_prompt(
     }
 }
 
+#[tauri::command]
+pub fn cmd_close_meeting_prompt(app: tauri::AppHandle) -> Result<(), String> {
+    if let Some(win) = app.get_webview_window("meeting-prompt") {
+        win.destroy().map_err(|e| e.to_string())?;
+    }
+    Ok(())
+}
+
 /// Surface a deferred update notification if one is pending and no session is active.
 /// Call this after recording/live/dictation stops.
 pub fn surface_deferred_update(app: &tauri::AppHandle) {
--- a/tauri/src-tauri/src/main.rs
+++ b/tauri/src-tauri/src/main.rs
@@ -932,7 +932,7 @@ fn show_meeting_prompt(app: &tauri::AppHandle, event: &minutes_core::calendar::C
 
     // Close any existing prompt window
     if let Some(win) = app.get_webview_window("meeting-prompt") {
-        win.close().ok();
+        win.destroy().ok();
     }
 
     // Stage the payload keyed by a monotonic token. The overlay reads its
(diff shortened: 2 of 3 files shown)


silverstein:
Merged. This is the cleaner version you promised when you closed #329, and it shows: routing every dismissal path through one native destroy, deferred out of the JS frame, with the closeScheduled guard on the JS side and a None-check on the native side, closes the WebKit self-close crash without the loose ends. I verified there are no orphaned getCurrentWebviewWindow references left, so no path can hit a missing symbol.

On your manual-test caveat: no worries about the Join & Record calendar path. The code that changed is the close path, which the dismiss/X/Escape/auto-close paths you did exercise all share, so it is well covered. The calendar write-only permission thing you hit is a separate known macOS quirk on this kind of managed Mac, not something your change touches.

That is three merged now (#328, #334) plus the engine direction on #332. Thanks Caleb.

PR #338, opened 2026-06-17
fix: centralize main window reveal

calebchongc:
## Summary
- make `show_main_window` idempotent for existing main windows
- expose a native `cmd_show_main_window` for frontend-triggered reveals
- route call detection, Recall expansion, and consent prompt reveals through the same guarded path

## Why
The latest Minutes Dev crash still points at WebKit frame updates on the main thread. The prior opacity/vibrancy fix is already present in the crashed binary, so this removes remaining duplicate show/focus triggers that bypassed the native main-window helper.

## Tests
- `cargo fmt --check`
- `cargo check -p minutes-app`
- `cargo test -p minutes-app tray_activity_tests -- --nocapture`
- `cargo test -p minutes-app`

## Dogfood
Rebuilt and installed `/Applications/Minutes Dev.app` on 0.18.12. Bundle seal verifies; local install is ad-hoc signed because no Developer ID identity is available on this machine.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -5545,10 +5545,7 @@ pub fn launch_recording(
             // the in-window button instead of erroring or bypassing
             // (spec Part A; review F1/F2). The frontend listener stashes the
             // pending args and re-invokes with consent confirmed.
-            if let Some(window) = app.get_webview_window("main") {
-                window.show().ok();
-                window.set_focus().ok();
-            }
+            crate::show_main_window(&app);
             let _ = app.emit(
                 "minutes://recording-consent-required",
                 serde_json::json!({ "disclosure": disclosure }),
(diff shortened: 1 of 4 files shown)


silverstein:
Reviewed. Making show_main_window idempotent (only show if not visible, only focus if not focused) and routing every duplicate show/focus through it, including replacing the frontend window.show()/setFocus() with the native cmd_show_main_window, is exactly the right move against the redundant-frame-mutation crash. Low risk, correct direction, approved as the base of the stack.

It predates 0.18.13, so it needs a rebase on latest main. One note: the reveal primitive here should also unminimize a minimized window (details in my #339 review, since that is where the full Bring All To Front routing lands). Happy to merge this first once rebased, then #339 on top.

calebchongc:
Rebased this onto latest `main` / `v0.18.13` and addressed the reveal follow-up from review.

Changes since the last review:
- `show_main_window` now restores a minimized main window before focusing it.
- Added a guardrail test to keep the unminimize-before-focus ordering in place.

PR #337, opened 2026-06-16
Detect Zoom vanity meeting URLs

calebchongc:
## Summary
- detect Zoom meeting links on vanity subdomains like `company.zoom.com` and dashed `*.zoom.us` hosts
- support Zoom personal links and Google Calendar redirect wrappers
- stop markdown link text at `]` so visible calendar URLs win over wrapped hrefs

## Tests
- `cargo test -p minutes-core calendar::tests::extract_ --no-default-features`
- `cargo test -p minutes-core --no-default-features`

PR #339, opened 2026-06-17
fix: guard main window frame mutations

calebchongc:
## Summary
- move Recall main-window size/position changes behind a native guarded command
- remove frontend permissions for set-focus/set-size/set-position/outer-frame APIs
- destroy dictation overlays synchronously before same-label rebuild/teardown
- route Bring All to Front through the guarded main-window reveal helper
- add guardrail tests for the window-lifecycle crash class

## Stack
Stacked on #338 (`fix: centralize main window reveal`). Until #338 merges, this PR includes that first commit plus the follow-up commit. Review focus for this PR is the second commit: `fix: guard main window frame mutations`.

## Why
The crash class points at WebKit/AppKit frame reconciliation, so after centralizing reveal/focus, the remaining risk is direct frontend frame mutation and async close/rebuild of transparent overlay windows. This pushes those operations into Rust-owned, visibility-guarded paths and narrows frontend capability surface.

## Tests
- `cargo fmt --check`
- `cargo check -p minutes-app`
- `cargo test -p minutes-app tray_activity_tests -- --nocapture`
- `cargo test -p minutes-app`

## Dogfood
Rebuilt and installed `/Applications/Minutes Dev.app` from this stacked branch on 0.18.12. Bundle seal verifies; local install is ad-hoc signed because no Developer ID identity is available on this machine.


silverstein:
Reviewed this closely (it is the riskiest of the three, since it narrows the capability surface in the crash area). The core is sound: I verified there are no orphaned frontend calls to the five removed capabilities, every site is migrated to a native command, and the synchronous dictation-overlay destroy is race-safe (the take() guard makes a double-dismiss a no-op rather than destroying a new session's window). Good, on-thesis work.

Two things to fix before merge, then it is good to go:

1. **The reveal primitive should unminimize.** `show_main_window` does `if !is_visible { show }` then focus, but never `unminimize()`. A minimized main window counts as visible, so revealing it (Bring All To Front, call-detected reveal) leaves it minimized. To be fair this predates your change, the old code did not unminimize either, but since you are consolidating this into the one reveal primitive it is the right moment to make it a true reveal: `if win.is_minimized().unwrap_or(false) { win.unminimize().ok(); }` before the show/focus.

2. **The guarded Recall layout command still fires no-op frame mutations.** `cmd_apply_recall_window_layout` calls `set_size(...)` unconditionally at the end, and sets position without checking the target differs. The whole point of this PR is to stop redundant frame mutations from hitting AppKit during reconciliation, so moving them into Rust but still firing them when nothing changed only relocates the problem. Read the current size/position, compare with a small tolerance, and skip the call when it is already there.

Optional/minor: the two outer-geometry read perms (allow-outer-position / allow-outer-size) are still granted transitively via `core:window:default`, so they are not actually removed at runtime. The mutators (set-focus/size/position) are genuinely gone, which is the part that matters for the crash, so this is just a tidy-up if you want the surface as narrow as the description implies.

Two logistics: this stack and #337/#338 all predate the 0.18.13 release, so please rebase on latest main (that also clears the test-count sync check). And because this changes window behavior in the crash area on my daily-driver machine, I am going to build it and exercise the reveal/Recall/dictation paths myself before merging. Really nice instinct on making Rust own all the frame mutations.

calebchongc:
Rebased this follow-up on top of the updated #338 branch and addressed the remaining window-lifecycle review notes.

Changes since the last review:
- Recall layout now compares the current frame before calling `set_position` / `set_size`, with a small tolerance to skip no-op frame mutations.
- Bring All to Front continues to route the main window through the shared reveal primitive from #338.
- Confirmed the frontend no longer calls the removed mutating window APIs directly.

Note: `core:window:default` still includes read-only outer geometry permissions transitively, as called out in review, but the crash-relevant mutators are removed from the frontend capability surface.

silverstein:
#337 and #338 are merged, thanks. I verified both fixes you pushed: show_main_window now unminimizes before focus (with the guardrail test), and the Recall layout command compares the current frame against the target with FRAME_EPSILON and skips the call when nothing changed, so it no longer fires no-op set_size / set_position into AppKit. That fully delivers the anti-redundant-mutation goal. Nice.

This now shows conflicts because #338 merged as a squash commit, so the copy of #338 inside this stacked branch collides. Could you rebase this on latest main? That drops the now-merged base commit and leaves just the frame-guard delta.

Once it is rebased and green, I will build it on my machine and exercise the reveal, Recall expand/collapse, and dictation-overlay paths before merging, since this is the capability-surface change in the crash area. That dev test is the last gate; everything in the diff itself checks out.

silverstein:
I went ahead and did the rebase for you (no need to redo it) since it was purely the mechanical stacked-squash conflict from #338 landing, not anything in your code. Replanted your frame-guard commit straight onto main, dropped the now-merged base commit, and confirmed fmt / clippy / the window-lifecycle guardrail tests all pass on the new base. Your commit and authorship are unchanged. It is mergeable again now.

Last gate is the dev-app test on my side (the reveal, Recall expand/collapse, and dictation-overlay paths, since this narrows the window capability surface in the crash area). I will run that and merge.

silverstein:
Merged. Full window-lifecycle stack is in now (#338 + #339). Verified the rebased commit (fmt/clippy/guardrail tests green) and eyeballed the dev app on my machine: window reveal renders, no crash. Your instinct to push every frame mutation into Rust and strip the frontend's window-mutation capabilities was the right architecture, and it should finally close the WebKit frame-update crash class that the earlier opacity fix only dented. Thanks Caleb, this was a strong stack.

PR #351, opened 2026-06-25
fix(tauri): stop Windows recall panel from resizing window and leaving a phantom column

maosuarez:
## Problem

On Windows, the embedded AI-assistant (Recall) panel caused two layout issues on the home view that don't occur on macOS (I am a win user):

1. **Window resizes when toggling the panel.** Opening/closing the panel drives an OS-level window resize (expand → 960px, collapse → 520px) through `cmd_apply_recall_window_layout`. On macOS this "grow to fit, snap back" is intentional, but on Windows it snaps the window to a fixed 520px on close regardless of the size the user chose, and the timing leaves transient whitespace.

2. **Phantom empty column when the panel is collapsed.** The top-level `.recall-panel { padding: 16px }` rule, combined with `width: 0` and the global `box-sizing: border-box`, forces the collapsed panel to render ~32px wide — a permanent empty column on the right edge of the home view. macOS never shows it because the window shrinks below the panel's footprint on collapse.

## Fix

Both changes are **platform-gated; macOS behavior is byte-for-byte unchanged.**

1. Skip the OS window resize on Windows (`IS_WINDOWS` guard, early-return in `applyRecallWindowLayout`). The CSS flex layout (`.recall-panel` `width: 0` ↔ `expanded`, `.app-left { flex: 1 }`) reflows on its own, so the window stays at whatever size the user set.

2. Zero the padding on the **collapsed** panel for non-macOS only:
   ```css
   html:not([data-platform="macos"]) .recall-panel:not(.expanded) { padding: 0; }
   ```
   The expanded panel keeps its 16px padding (the terminal needs it).

Linux also benefits from fix #2 (it had the same phantom column); the rule is gated away from macOS, not toward a single platform, and removes a gap that shouldn't exist anywhere.

## Scope

- 1 file changed: `tauri/src/index.html` (+21 lines)
- No Rust changes
- No macOS/Linux regressions — all changes guarded by platform detection

## Testing

Verified visually on Windows 11 with `cargo tauri dev`:
- Toggling the AI assistant panel no longer resizes/snaps the window
- Collapsed panel no longer leaves an empty column on the right
- Expanded panel still renders with correct internal padding

### DIFF
--- a/tauri/src/index.html
+++ b/tauri/src/index.html
@@ -4271,6 +4271,15 @@
       gap: 12px;
     }
 
+    /* On non-macOS (Windows/Linux) the OS window stays wide when the panel is
+       collapsed, so the `padding: 16px` above — combined with `width: 0` and
+       `box-sizing: border-box` — leaves a ~32px phantom column on the right.
+       On macOS the window shrinks to 520px on collapse (below any threshold),
+       so this was never visible there. Zero the padding only when collapsed. */
+    html:not([data-platform="macos"]) .recall-panel:not(.expanded) {
+      padding: 0;
+    }
+
     .recall-header {
       border-radius: 12px;
     }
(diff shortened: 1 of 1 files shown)


silverstein:
Merged, thank you. This is exactly the kind of Windows coverage the project needs most, and the platform gating is clean. I especially appreciate the comments explaining why the macOS resize is intentional and why the collapsed padding only matters off-mac. That context makes this easy to maintain later. Hope to see more from you.

PR #367, opened 2026-06-26
feat(minutes-prep): add Microsoft 365 / Outlook calendar source

gtheys:
## What

Add `m365` CLI as a 4th calendar source in the `minutes-prep` skill, alongside the existing `gog`, `icalbuddy`, and `osascript` sources.

## Why

Users on Microsoft 365 / Outlook have no way to pull today's calendar events into the meeting prep flow. The `m365` CLI (CLI for Microsoft 365) is a widely-used cross-platform tool that can fetch Outlook events.

## How

- Detects if `m365` is installed (`command -v m365`)
- Checks login status via `m365 status`
- If not logged in, prompts user via `AskUserQuestion` to optionally start browser-based login (`m365 login --authType browser`)
- If logged in, fetches today's remaining events via `m365 outlook event list`
- Filters out all-day events, cancelled events, and solo events (no attendees)
- Converts UTC `start.dateTime` / `end.dateTime` to local time for display
- Falls through silently if m365 not installed or user declines login

## Files changed

- `tooling/skills/sources/minutes-prep/skill.md` — canonical source
- `.agents/skills/minutes/minutes-prep/SKILL.md` — agents mirror
- `.claude/plugins/minutes/skills/minutes-prep/SKILL.md` — Claude plugin mirror
- `.opencode/skills/minutes-prep/SKILL.md` — OpenCode mirror

### DIFF
--- a/.agents/skills/minutes/minutes-prep/SKILL.md
+++ b/.agents/skills/minutes/minutes-prep/SKILL.md
@@ -45,7 +45,33 @@ gog calendar list --today --json -a <account> 2>/dev/null
 osascript -e 'tell application "Calendar" to get {summary, start date} of (every event of every calendar whose start date >= (current date) and start date < ((current date) + 1 * days))'
 ```
 
-**4. None available** — skip to Phase 1 and ask manually.
+**4. Microsoft 365 / Outlook (`m365` CLI)** (if installed):
+
+First check if m365 is installed and whether it's logged in:
+```bash
+command -v m365 2>/dev/null && m365 status --output json 2>/dev/null
+```
+
+- **If not installed**: skip silently to the next source.
+- **If installed but not logged in** (`status` returns `"Logged out"`): ask the user via AskUserQuestion:
+  > "I found the Microsoft 365 CLI (`m365`) but you're not logged in. Want me to start the login flow so I can pull your Outlook calendar?"
+  - If **yes**: run `m365 login --authType browser` and wait. Once done, proceed to fetch events below.
+  - If **no**: skip to the next source silently.
+- **If installed and logged in**: fetch today's remaining events:
+```bash
+m365 outlook event list \
+  --userId "@meId" \
+  --startDateTime "$(date -u +%Y-%m-%dT%H:%M:%S)" \
+  --endDateTime "$(date -u +%Y-%m-%d)T23:59:59" \
+  --output json 2>/dev/null
+```
+Only use the result if the command exits with code 0 and returns a non-empty JSON array. Parse the response:
+- **Title**: `event.subject`
+- **Start time**: `event.start.dateTime` (UTC — convert to local time for display)
+- **Attendees**: `event.attendees[].emailAddress.name` (extract first names)
+- **Skip if**: `event.isAllDay === true`, `event.isCancelled === true`, or `event.attendees` is empty
+
+**5. None available** — skip to Phase 1 and ask manually.
 
 **If upcoming meetings are found:**
 Present via AskUserQuestion: "I see you have these meetings coming up today:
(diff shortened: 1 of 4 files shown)


silverstein:
Merged, thank you. Three in the bank now, after the PipeWire segfault fix (#160) and the Stdio::null fix (#288), and this carries the same instinct: you found the four-way skill sync on your own (the sync guard is green), kept the whole thing best-effort with silent fall-through, and gated the login behind an explicit prompt. Outlook/M365 users now have a path into prep that didn't exist before.

One small, non-blocking follow-up whenever you feel like it: the event window uses UTC day boundaries (`date -u +%Y-%m-%d` ... `T23:59:59`) while `startDateTime` is the current UTC time. For users far from UTC (US Pacific in the evening, say, where local "today" is already tomorrow in UTC), that window can drift off their local day. The other sources lean on local-day semantics (`gog --today`). It degrades to "shows some events," so it's minor, but widening the fetch to now..+24h UTC and filtering to local-today would tighten it. We may just push a small follow-up for that. Either way this was a pleasure to review.

silverstein:
Small follow-up: I pushed a refinement on top of your m365 source in #368 (just merged). It scopes the event fetch to the user's local day (over-fetch a 24h UTC window, then filter to local-today) so it stays correct at any UTC offset, not just near UTC. Your feature, tightened. Nothing for you to do.

Also, a fun one you might enjoy: your name is quietly load-bearing in our test suite. The name-correction feature has a canonical eval case literally labeled `dutch-geert`: pool `["Geert", "Sanne"]`, input `"thanks bert for the notes"`, expected `"thanks Geert for the notes"`. It's the headline acceptance criterion for the whole feature (the "Geert from Bert" class, a different-first-letter correction that only fires once Geert is a confirmed participant). So every time that corrector ships, it's because "Geert" passed. Seemed right to let you know.

gtheys:
@silverstein That is awesome I was still struggling with names. This is 1 person:

image

silverstein:
Love it, and that screenshot is a perfect bug report. What you're seeing is two different problems stacked, and only one is the name-corrector's job:

1. **Spelling drift** (`junlei` / `junrei` / `junwei` / `jun-rei`). The transcriber spells the name a little differently each time, and the knowledge graph keys people by an exact slug, so each spelling becomes its own person. The graph deliberately does not fuzzy-merge on its own, because a wrong merge (collapsing two real people) is worse than a split.
2. **Role-suffix contamination** (`junlei-tech-lead`, `junrei-core-team`). Here a title leaked into the name during entity extraction. That's a separate bug from spelling, and a real one. Filed it as #370.

The lever that fixes this today is the local vocabulary, which feeds the name-correction pool you just read about. Add the canonical spelling once with the bad variants as aliases:

```
minutes vocabulary add --kind person "Junrei" \
  --alias junlei --alias junwei --alias jun-rei \
  --alias junlei-tech-lead --alias junrei-core-team
```

(Use whichever spelling is actually correct as the canonical.) After that, the corrector normalizes those variants to "Junrei" before they reach the graph, so future meetings converge on one person instead of six. One caveat relevant to your case: the relaxed, different-first-letter corrections (junwei to Junrei) only fire once Junrei is a confirmed participant in the meeting, the same gating as that dutch-geert test, so having the person in attendees/identity for that meeting matters.

`minutes graph` also prints alias suggestions when it detects likely-same-person pairs, so it may already be flagging some of these for you.

Bigger picture: this whole area (person entity resolution across spelling drift, role contamination, and diarization) is worth a real upgrade, and we're researching the June 2026 state of the art plus a proper eval for it so the fix is measured, not guessed. Your example is exactly the kind of real-world mess that sharpens that. Thank you.

PR #374, opened 2026-06-27
fix(graph): strip role/title suffixes from person entities (#370)

### LINKED ISSUE #370: Knowledge graph: role/title suffixes contaminate person entity names (one person split into many)
## Reported by

@gtheys in #367 (screenshot: one real person showing up as six separate knowledge-graph entities).

## Problem

The conversation knowledge graph creates separate person entities when a role or title descriptor leaks into the extracted name. All six of these are one person:

```
jun-rei
junlei
junlei-tech-lead
junrei
junrei-core-team
junwei
```

Two failure modes are visible, stacked:

1. **Role/title contamination (the focus of this issue).** `junlei-tech-lead` and `junrei-core-team` fuse a role ("tech lead", "core team") into the person name, producing a distinct slug from the bare name. A title should never become part of identity.
2. **Spelling drift (related, mitigation exists today).** `junlei` / `junrei` / `junwei` / `jun-rei` are transcriber spelling variants. The graph keys people by exact slug and deliberately does not fuzzy-merge (a wrong merge is worse than a split). Current lever: `minutes vocabulary add --kind person "Junrei" --alias junlei ...` feeds the name-correction pool so variants normalize before reaching the graph.

## Expected

Role/title descriptors must not become part of a person's canonical name or slug. "Junrei, the tech lead" and "Junrei from the core team" should resolve to the person `Junrei`, with the role captured as a role attribute, not baked into the identity/slug.

## Pointers

- `crates/core/src/knowledge_extract.rs` (structured entity extraction from frontmatter)
- `crates/core/src/graph.rs` (`push_file_person`, slug keying, `person_role_priority`, `merge_person_aliases`, `detect_aliases`)
- `crates/core/src/person_identity.rs` (`PersonCanonicalizer`)

## Direction

Auto-merging distinct people is intentionally avoided, so the fix should target extraction/canonicalization (strip role/title suffixes when constructing the entity name) rather than post-hoc fuzzy graph merging. A broader entity-resolution + eval improvement for this whole area is being researched separately and will land as its own proposal.

maosuarez:
## Summary

Fixes #370 — implements **Slice 1** of the entity-resolution plan in #371.

The LLM occasionally appends role/title context to extracted entity names (`"Junlei, tech lead"`, `"Junrei (core team)"`, `"Sam - engineering lead"`). That descriptor leaked into the slug (`junlei-tech-lead`, `junrei-core-team`, `junrei-tech-lead`…), so a single person fanned out into multiple graph nodes — the exact symptom reported in #370.

## What changed

Two layers, by design:

**Primary — `pipeline.rs::add_person_entity()`** (the correct fix per the plan)
Strips the role suffix at the extraction point, *before* the slug is written to the `.md` frontmatter. `"Junlei, tech lead"` → slug `junlei`. New meetings never persist a contaminated slug.

**Secondary — `person_identity.rs`** (retroactive defense)
New `pub(crate) strip_role_suffix()` with 17 role/title patterns. Handles separator forms (`, `, ` (`, ` — `, ` from `, ` the `) and a vocabulary fallback that is hyphen/space-agnostic, so both label form (`"Junlei Tech Lead"`) and slug form (`"junlei-tech-lead"`) collapse to `junlei`. `normalize_entity_identity()` plus the exact/slug indexing paths apply the strip too, so meetings **already on disk** with contaminated slugs still resolve to the clean identity on the next index rebuild — without requiring a re-extraction.

## Why two layers

The pipeline fix stops the bleeding for all future extractions. But existing vaults already contain contaminated frontmatter; the `person_identity.rs` layer makes lookups/canonicalization resolve those to the clean slug so users don't have to reprocess history.

## Tests

- `pipeline.rs`: +1 — `add_person_entity_strips_role_suffix_before_slug`
- `person_identity.rs`: +13 — separator/parenthetical/dash/connector/vocab forms, label vs slug form, clean-name-untouched, `normalize_entity_identity` variants, canonicalizer resolution
- `graph.rs`: +1 integration — `role_suffix_on_entity_does_not_create_spurious_graph_node` asserts contaminated slugs never reach the `people` table and action items link to the clean slug

Full suite green locally (845 passing, 0 failing). The 2 clippy findings on this branch are pre-existing on `main` in files this PR doesn't touch (confirmed via `git stash`).

Closes #370

### DIFF
--- a/crates/core/src/graph.rs
+++ b/crates/core/src/graph.rs
@@ -2111,6 +2111,103 @@ attendees: [Sarah Miller]
     }
 
     #[test]
+    fn role_suffix_on_entity_does_not_create_spurious_graph_node() {
+        // Regression for issue #370: the LLM sometimes appends role context to
+        // entity labels and slugs ("Junlei, tech lead" → slug "junlei-tech-lead").
+        // After the fix, only the clean slug "junlei" should appear in the graph.
+        let tmp = TempDir::new().unwrap();
+        let meetings = tmp.path().join("meetings");
+        fs::create_dir_all(&meetings).unwrap();
+        let meeting = r#"---
+title: Architecture Review
+type: meeting
+date: 2026-03-20T14:00:00-07:00
+duration: 30m
+attendees: [Junlei, Junrei]
+entities:
+  people:
+    - slug: junlei-tech-lead
+      label: "Junlei, tech lead"
+      aliases: []
+    - slug: junrei-core-team
+      label: "Junrei (core team)"
+      aliases: []
+action_items:
+  - assignee: Junlei
+    task: Review the architecture doc
+    due: "2026-03-25"
+    status: open
+---
+
+## Transcript
+[JUNLEI 0:00] Let's review the design.
+[JUNREI 0:30] I agree with the approach.
+"#;
+        write_meeting(&meetings, "arch-review.md", meeting);
+        let config = test_config(&meetings);
+        let db = tmp.path().join("graph.db");
+        rebuild_index_at(&config, &db).unwrap();
+        let conn = open_db(&db).unwrap();
+
+        // Contaminated slugs must not appear in the people table
+        let contaminated_junlei: i64 = conn
+            .query_row(
+                "SELECT COUNT(*) FROM people WHERE slug = 'junlei-tech-lead'",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+        assert_eq!(
+            contaminated_junlei, 0,
+            "role-contaminated slug 'junlei-tech-lead' must not exist in the graph"
+        );
+        let contaminated_junrei: i64 = conn
+            .query_row(
+                "SELECT COUNT(*) FROM people WHERE slug = 'junrei-core-team'",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+        assert_eq!(
+            contaminated_junrei, 0,
+            "role-contaminated slug 'junrei-core-team' must not exist in the graph"
+        );
+
+        // Clean slugs must be present
+        let junlei: i64 = conn
+            .query_row(
+                "SELECT COUNT(*) FROM people WHERE slug = 'junlei'",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+        assert_eq!(junlei, 1, "clean slug 'junlei' must exist in the graph");
+        let junrei: i64 = conn
+            .query_row(
+                "SELECT COUNT(*) FROM people WHERE slug = 'junrei'",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+        assert_eq!(junrei, 1, "clean slug 'junrei' must exist in the graph");
+
+        // The action item assigned to "Junlei" must resolve to the clean slug
+        let commitment_person_count: i64 = conn
+            .query_row(
+                "SELECT COUNT(*) FROM commitments c
+                 JOIN people p ON c.person_id = p.id
+                 WHERE c.commitment_type = 'action_item' AND p.slug = 'junlei'",
+                [],
+                |r| r.get(0),
+            )
+            .unwrap();
+        assert_eq!(
+            commitment_person_count, 1,
+            "action item must be linked to clean slug 'junlei'"
+        );
+    }
+
+    #[test]
     fn test_corrupted_db_auto_rebuild() {
         let tmp = TempDir::new().unwrap();
         let meetings = tmp.path().join("meetings");
--- a/crates/core/src/person_identity.rs
+++ b/crates/core/src/person_identity.rs
@@ -303,8 +303,7 @@ pub(crate) fn strip_role_suffix(name: &str) -> &str {
         if !name.is_char_boundary(split) {
             continue;
         }
-        let candidate =
-            name[..split].trim_end_matches(|c: char| c == '-' || c.is_whitespace());
+        let candidate = name[..split].trim_end_matches(|c: char| c == '-' || c.is_whitespace());
         if !candidate.is_empty() {
             return candidate;
         }
@@ -567,7 +566,10 @@ mod tests {
             strip_role_suffix("Robert (Bob) Smith"),
             "Robert (Bob) Smith"
         );
-        assert_eq!(strip_role_suffix("Mike (Michael) Johnson"), "Mike (Michael) Johnson");
+        assert_eq!(
+            strip_role_suffix("Mike (Michael) Johnson"),
+            "Mike (Michael) Johnson"
+        );
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


silverstein:
Thanks for this, the two-layer approach (strip at extraction in `pipeline.rs` plus retroactive resolution in `person_identity.rs`) is exactly right. Before a logic review, here is a triage of the red CI so you are not left guessing: none of the three failures are problems with your actual change.

**1. Site Release Link Consistency** is the only one on your side, and it is a one-command fix. That check runs `scripts/sync_site_release_version.mjs --check`, which regenerates `site/lib/release.ts` from the source tree and diffs it against what is committed. One of the baked-in constants is `MINUTES_TEST_COUNT`, computed by counting `#[test]` / `#[tokio::test]` across the repo. Your PR adds new tests, so the computed count no longer matches the committed `1248` and the check fails. Fix:

```
node scripts/sync_site_release_version.mjs
git add site/lib/release.ts && git commit -m "chore: sync site release test count"
```

That will clear it. Worth knowing this gate trips on essentially any PR that adds or removes a test.

**2. Test (macos-latest)** is our CI runner, not your code. The macOS image migrated to Xcode 26.5 and the linker is now looking for `clang_rt.osx` in a stale Xcode 16.4 path (`ld: library 'clang_rt.osx' not found`). It is hitting every Rust PR right now and we will fix it at the workflow level. Note that the Ubuntu and Windows test jobs both pass on your branch.

**3. Vercel** ("Authorization required to deploy") is expected for fork PRs, which do not get preview deploys. Safe to ignore.

So the only action on your end is the `release.ts` regen. I will follow up with notes on the logic itself shortly. Thanks again.

maosuarez:
Thanks for the detailed triage. Pushed the `release.ts` regen — that should clear Site Release Link Consistency. Standing by for your logic notes.

silverstein (commented):
CI is sorted: your regen cleared Site Release Link Consistency, and Ubuntu plus Windows tests are green. The only remaining red is `Test (macos-latest)`, which is the Xcode 26.5 runner issue on our side, not your code. So this is purely a logic review.

Overall this is a clean, well-tested fix. Reusing the single `strip_role_suffix` in both `pipeline.rs` (extraction) and `person_identity.rs` (indexing and lookup) keeps the logic in one place, the deliberately multi-word role list avoids most vocabulary false positives, the `is_char_boundary` guard for non-ASCII lowercasing is a nice touch, and the `graph.rs` integration test (contaminated slug never reaches the `people` table, action item links to the clean slug) is exactly the assertion I would want.

My one concern before merge is over-stripping legitimate names, and it lives in the structural branches (comma, parenthetical, dash, connectors), which fire unconditionally without checking that a role actually follows. The precise vocabulary branch is undercut by them.

**1. Parenthetical nickname (the one I would want fixed).** The `" ("` branch truncates at the first parenthesis with no role check. LLMs often emit names like `"Mike (Michael) Johnson"` or `"Robert (Bob) Smith"`. Those become `"Mike"` / `"Robert"`, dropping the surname. That loses data and over-merges different people who share a first name, which is the same fan-out bug in reverse. `strip_role_suffix("Robert (Bob) Smith")` returns `"Robert"`.

**2. Comma branch is role-agnostic (lower priority).** The `", "` truncation also eats credential and generational suffixes: `"Sammy Davis, Jr."`, `"Jane Doe, PhD"`. Mostly harmless, but `", Jr."` versus `", Sr."` can over-merge a parent and child pair.

**3. `" the "` / `" from "` connectors are unbounded (low).** They match anywhere and do not require a known role to follow, so `"Winnie the Pooh"` becomes `"Winnie"`. Rare for real participants, but a latent footgun.

The cleanest fix for all three is to gate the structural branches on the trailing fragment actually matching (or containing) a `ROLE_TITLE_SUFFIXES` token, which would make them as precise as the vocabulary branch. At minimum I would add false-positive tests for `"Robert (Bob) Smith"`, `"Sammy Davis, Jr."`, and a `" the "` name, asserting they are left intact.

Since this is Slice 1 of #371, I am happy to land it once the parenthetical case is handled (or explicitly scoped out with a test documenting it as a known limitation). The rest can be follow-ups. Really nice work overall.

maosuarez:
Thanks for the detailed review. Pushed a fix for the structural branches:

- Added `trailing_fragment_is_role()` helper that normalizes and checks the fragment against `ROLE_TITLE_SUFFIXES` before any structural branch fires.
- The parenthetical branch also gains an `ends_with(')')` guard, which rules out `"Robert (Bob) Smith"` without needing to inspect the content at all.
- Updated two existing tests that were asserting the old over-broad behavior (`"Alex - PM"` and `"Sam from engineering"`) to use concrete role tokens instead.
- Added the three false-positive guard tests you called out: `"Robert (Bob) Smith"`, `"Sammy Davis, Jr."`, `"Winnie the Pooh"` — all assert unchanged.

848 tests pass.

silverstein:
Heads up: the `Test (macos-latest)` failure you hit was never your code. I just merged #375, which fixes it. The cargo `target/` cache was being restored across a runner-image rotation (macOS 15 to macOS 26) and pulling a stale `Xcode_16.4` linker path that no longer exists on the new image, so the link failed with `clang_rt.osx not found`. #375 scopes the cache key by the actual runner image (ImageOS + ImageVersion).

I have updated this branch with latest `main` so it re-runs against the fix. Once the new run finishes, `Test (macos-latest)` should be green. Combined with your `release.ts` regen, that clears all the CI blockers here. The only thing left from my earlier review is the parenthetical-nickname over-strip (`"Robert (Bob) Smith"` -> `"Robert"`); handle that or scope it out with a test and this is good to go. Thanks for your patience.

silverstein:
Quick correction to my last comment: the macOS infra fix landed and worked. On the new run, `Build full workspace`, all the tests, and `Clippy` all pass on `macos-latest`. The only remaining red there is the `Format check` step: `cargo fmt --all -- --check` flags `crates/core/src/person_identity.rs` at lines 303, 567, and 581. This was hidden before because the job used to die at the build step and never reached fmt.

Fix is one command:

```
cargo fmt --all
git commit -am "style: cargo fmt"
```

Heads up that `fmt` (and `clippy`) only run on the macOS job in our matrix, which is why the Ubuntu and Windows checks were green. After the fmt commit, `macos-latest` should be fully green and, with the parenthetical-nickname item from the review, this is ready.

silverstein:
Caught up on your 01:42 push (my macOS comments crossed it, so ignore the "parenthetical still pending" line above). I reviewed the structural-branch fix and it's right: `trailing_fragment_is_role()` gating plus the `ends_with(')')` guard handle all three cases correctly. `"Robert (Bob) Smith"`, `"Sammy Davis, Jr."`, and `"Winnie the Pooh"` all fall through unchanged, while real role forms (`"Junrei (core team)"`, `"Sam the tech lead"`) still strip. That closes the logic concern from my review.

So the only thing left is the `cargo fmt` check (person_identity.rs:303/567/581). Run `cargo fmt --all`, push, and once `macos-latest` is green this is good to merge. Nice work.

maosuarez:
Applied `cargo fmt` — format check should be green now. Thanks for the thorough review!

Issue #372, opened 2026-06-27
feat: UI font scaling for accessibility (⌘+/⌘−/⌘0)


SherriAlexander:
## Problem

The Minutes desktop app uses fixed pixel font sizes throughout — dominant tiers at 11px and 12px — with no way to adjust them. For users with lower vision, aging eyes, or high-DPI displays at large physical distances, these sizes are too small to read comfortably. There is no system-level workaround: macOS Display Zoom changes the entire OS, not just the app, and Minutes has no zoom support of its own.

## Proposed solution

- Replace every hardcoded `font-size: Xpx` rule (~175 rules across 13 size tiers) with CSS custom property tokens (`--text-xs`, `--text-sm`, `--text-body`, `--text-meta`, etc.) that scale relative to a single `--text-scale-basis` variable
- Add `⌘+` / `⌘−` to increase/decrease the base rem font size in 2px steps (16px floor, 26px ceiling), and `⌘0` to reset
- Persist the chosen zoom level in `localStorage` so it survives app restarts
- Sync zoom across all open windows (main, dictation overlay, note, palette, etc.) so every surface stays consistent
- Resize secondary windows proportionally when zoom changes so nothing clips or overflows

## Who benefits

- Users with reduced vision or sensitivity to small text
- Users on large or high-resolution external displays where 11px–12px text is physically tiny
- Anyone who finds the transcript-heavy UI fatiguing at default size

## Non-goals

Not a replacement for OS-level accessibility. Users who need system-wide zoom should use macOS Accessibility settings. This targets the broad middle range of users whose vision is fine but find small labels tiring over a full workday.

## How to verify (once implemented)

- Build `~/Applications/Minutes Dev.app` via `./scripts/install-dev-app.sh`.
- Press `⌘+` several times, navigate between views, open About/What's New modals, open New Note window.
- Text should scale consistently, windows resize proportionally, zoom should persist after quit and relaunch.
- Press `⌘-` to make sure text size decreases
- Press `⌘0` to reset to defaults

I'd actually like to take this on myself -- I've gotten quite far, and have it working pretty well across most surfaces of the app.  Working on finding all the different displays and testing edge cases now.  

Let me know ASAP if there are any concerns about the approach!

silverstein:
Love this. In-app text scaling is a real accessibility gap, and the token-based approach (one `--text-scale-basis`, everything else relative) is exactly right. Green light to keep going. A few notes from a pass over the code so you don't hit surprises:

[...]

PR #381, opened 2026-06-29
feat(ui): Persistent UI font scaling with keyboard shortcuts

SherriAlexander:
## Summary

- Adds persistent UI zoom with `Cmd++` / `Cmd+-` / `Cmd+0` keyboard shortcuts, stored in Tauri store so it survives restarts
- Extracts shared zoom logic into `ui-zoom.js` + `ui-zoom-tokens.css` and syncs the zoom factor across all windows (dictation overlay, palette, meeting prompt, note) via Tauri events
- Scales window *dimensions* (both main and non-main windows) proportionally with the zoom factor so layouts don't clip

## Still to do
- Adjust layout of meeting detail (column width is way off)
- Scale the About modal width more effectively (more like "What's New")
- Scale the Settings modal width
- Test Artifact template overlay
- Test CLI setup scrim
- Add behavior to terminal.html if needed (doesn't seem to need it?)
- Palette window has a thin square-cornered outline showing outside the curved border radius -- may be pre-existing, but verify
- Naming adjustments for consistency
- Test thoroughly using the demo data
- Make some spacing adjustments where needed
- Discuss spacing tokens vs. ems when it comes to declaring heights

## What changed

| File | Change |
|---|---|
| `tauri/src-tauri/src/main.rs` | Zoom Tauri commands (`cmd_get_zoom`, `cmd_set_zoom`), cross-window sync via `zoom-changed` event, window resize logic |
| `tauri/src/scripts/ui-zoom.js` | Shared zoom bootstrap: reads stored factor, applies CSS `font-size` on `<html>`, listens for sync events |
| `tauri/src/styles/ui-zoom-tokens.css` | CSS custom props for all font-size tiers (9px–48px) as `rem` multiples of the zoom base |
| `tauri/src/index.html` | Imports shared zoom script/styles; `About` and `What's New` modal widths scale with zoom |
| `tauri/src/dictation-overlay.html` | Scalable overlay — imports shared zoom bootstrap |
| `tauri/src/meeting-prompt.html` | Imports shared zoom bootstrap |
| `tauri/src/note.html` | Imports shared zoom bootstrap |
| `tauri/src/palette/index.html` | Imports shared zoom bootstrap |

## Test plan

- [ ] `Cmd++` / `Cmd+-` increase/decrease font size across main window
- [ ] `Cmd+0` resets to default
- [ ] Zoom persists after app restart
- [ ] Dictation overlay, palette, note, and meeting-prompt windows match zoom level
- [ ] Window dimensions scale proportionally — no clipping at min/max zoom
- [ ] About and What's New modal widths scale correctly

### DIFF
--- a/tauri/src-tauri/src/main.rs
+++ b/tauri/src-tauri/src/main.rs
@@ -382,6 +382,35 @@ fn cmd_apply_recall_window_layout(
     Ok(())
 }
 
+#[tauri::command]
+fn cmd_scale_window(app: tauri::AppHandle, label: String, zoom: f64) -> Result<(), String> {
+    const BASE: f64 = 16.0;
+    let Some(win) = app.get_webview_window(&label) else {
+        return Ok(());
+    };
+    let scale = win
+        .current_monitor()
+        .ok()
+        .flatten()
+        .map(|m| m.scale_factor())
+        .unwrap_or(1.0);
+    let size = win.inner_size().map_err(|e| e.to_string())?;
+    let ratio = zoom / BASE;
+    let new_w = ((size.width as f64 / scale) * ratio).round();
+    let new_h = ((size.height as f64 / scale) * ratio).round();
+    let was_resizable = win.is_resizable().unwrap_or(false);
+    if !was_resizable {
+        win.set_resizable(true).ok();
+    }
+    let result = win
+        .set_size(LogicalSize::new(new_w, new_h))
+        .map_err(|e| e.to_string());
+    if !was_resizable {
+        win.set_resizable(false).ok();
+    }
+    result
+}
+
 fn show_note_window(app: &tauri::AppHandle) {
     if let Some(win) = app.get_webview_window("note") {
         win.show().ok();
@@ -2504,6 +2533,7 @@ fn main() {
             commands::cmd_mark_activation_nudge_shown,
             cmd_show_main_window,
             cmd_apply_recall_window_layout,
+            cmd_scale_window,
             commands::cmd_upcoming_meetings,
             commands::cmd_spawn_terminal,
             commands::cmd_pty_input,
--- a/tauri/src-tauri/src/main.rs
+++ b/tauri/src-tauri/src/main.rs
@@ -410,19 +410,9 @@ fn cmd_scale_window(app: tauri::AppHandle, label: String, zoom: f64) -> Result<(
     let height = (base_h * ratio).round();
 
     // setFrame from Rust works on non-resizable windows; no resizable toggle is
-    // needed (toggling it makes macOS draw a square standard frame on the
-    // transparent rounded window).
+    // needed
     win.set_size(LogicalSize::new(width, height))
         .map_err(|e| e.to_string())?;
-
-    // Transparent + shadowed windows (palette) keep a stale square shadow around
-    // the full frame after a programmatic resize. Toggling the shadow forces
-    // AppKit to recompute it from the new rounded content mask. Decorated windows
-    // (note, meeting-prompt) and shadowless ones (dictation-overlay) don't need it.
-    if label == "palette" {
-        win.set_shadow(false).ok();
-        win.set_shadow(true).ok();
-    }
 
     // Clamp position so bottom-right anchored windows (e.g. dictation overlay)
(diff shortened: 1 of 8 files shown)


SherriAlexander:
@silverstein This is still a work in progress, but I think it's already pretty aligned to our discussion in issue #372!  Let me know if you see anything concerning!

(I'd also like to chat about the spacing rules you mentioned -- totally agree that using a `px` token-based spacing system makes sense in general, but there are cases where using `em`-based spacing can work better.  For example, to declare padding in buttons, `em` spacing will scale where `px` spacing won't.)

Here is the screen recording (link valid until 7/3): https://ln5.sync.com/dl/6e934f670#2zxkhrfu-9uxpguah-fcusshtb-cf34p79k

silverstein:
This is really nicely done, and the architecture is exactly the shape I was hoping for. Highlights from a close read:

- rem-token scaling that's **size-only**: `:root { font-size: calc(var(--text-scale-basis) * 1px) }` with tokens in rem. Fonts, colors, and radii are untouched, so it stays within DESIGN.md. The token comments show you did the real px-tier audit (11px = 52 rules, 13px = 28 rules, folding 15/17/22 into neighbors). Clean.
- All five surfaces wired (index, dictation-overlay, meeting-prompt, note, palette), with `terminal.html` correctly left out since xterm owns its own font sizing. Good call.
- FOUC handled by restoring `--text-scale-basis` before render, and no echo loop on the receive path.
- `cmd_scale_window` is thoughtful: scaling from a fixed `window_base_size` (so reopens never compound), recomputing the palette's transparent shadow, and clamping position so a scaled-up overlay can't run off-screen. Those are exactly the macOS quirks that bite here.

Two things before merge:

**1. Cross-window sync uses `BroadcastChannel`, not Tauri events (the PR text says "Tauri events").** This is the one I'd want to nail down: `BroadcastChannel` isn't guaranteed to propagate across separate WKWebView windows on macOS, which is why I steered toward Tauri `emit`/`listen` on the issue (it's what the rest of the app uses for cross-window signals, e.g. `permissions:changed`, `shortcut:status`). If it doesn't propagate, live sync silently fails: a zoom change in the main window won't update an already-open palette, though newly-opened windows still pick up the persisted value from shared `localStorage`, so it degrades rather than fully breaks. Could you either confirm it works across windows on macOS or switch the cross-window signal to a Tauri event? I'm building this onto a macOS dev app now to test it directly and will report what I see.

**2. `Cmd++` is effectively dead on US keyboards.** The keydown handler returns early on `e.shiftKey`, but typing `+` requires Shift, so only `Cmd+=` actually zooms in. Functionality isn't lost (Cmd+= is the common muscle memory), but the advertised `Cmd++` never fires. One-line fix: allow Shift through for the `=`/`+` branch.

Neither is a big lift, and the foundation is strong. Thanks for the careful work and for following the cross-window-sync guidance from the issue (the intent is right; it's just the transport to verify).

silverstein:
Tested on a macOS dev build: `Cmd+=` / `Cmd+-` / `Cmd+0` resize text cleanly across the surfaces — feels great. Whenever you've handled the two review notes (the `Cmd++` keydown guard, and confirming or switching the cross-window sync from `BroadcastChannel` to a Tauri event), mark it out of draft and I'll merge. No rush — nice work on this.

silverstein:
Pushed one tiny commit (4a9b607) for the `Cmd++` guard we discussed: dropped `e.shiftKey` from the keydown bail so `Cmd++` (which is `Cmd+Shift+=` on US layouts) zooms in, not just `Cmd+=`. The shifted forms of `-`/`0` produce different keys (`_`/`)`) that don't match, so no false triggers. Hope that's alright — your cross-window-sync switch to Tauri events is exactly right. Merging once CI is green. Thanks for the great work on this.

SherriAlexander:
@silverstein No problem at all, thank you for picking that up!

I was thinking I'd test a bit more, but if you're okay with it, I don't mind if you merge it in!

I redefined the widths of the windows/overlay cards using `ch` units so they'd scale nicely with font size.  Left some comments in the CSS so it would be easier to see how they translate to pixels down the line.  :) 

Let me know if you run into any issues, or have any questions -- and thanks for being so open to this idea! 🙏

Issue #382, opened 2026-06-29
speaker_mapping agent times out (120s, 0 output) → meetings ship with only the recorder named


silverstein:
## Summary

The `speaker_mapping` step delegates to a tool-enabled LLM agent
(`model: agent:claude`) hard-capped at **120s**. On real multi-party calls it is
**consistently hitting the cap and returning zero output**
(`outcome: error, reason: "Agent timed out", output_chars: 0`). When that
happens, no speaker gets a name except the recorder — and only if the
deterministic stem self-attribution path also fires. The user sees a clean,
well-separated transcript with every voice but their own left anonymous, and
**no surfaced warning** that naming was attempted and failed.

This is a **distinct failure mode** from:
- the silent-system-stem collapse (diarization itself returns 1 speaker), and
- prompt-truncation (the LLM returns a *wrong/partial* map).

Here diarization is healthy and the prompt is delivered — the agent simply
**never returns**.

## Evidence — two independent meetings, same day, same timeout

From `~/.minutes/logs/minutes.log`, 2026-06-29.

**Meeting 1 — 7-person team call (`job-20260629122506729`, from `.mov`)**
```json
diarize:         {"segments":1180,"speakers":6}
speaker_mapping: {"attendees":17,"input_chars":4230,"model":"agent:claude",
                  "outcome":"error","output_chars":0,
                  "reason":"Agent timed out","speaker_labels":5,"duration_ms":120060}
attribution:     {"raw_diarization_num_speakers":6,
                  "speaker_map":[{"name":"<recorder>","source":"deterministic",
                                  "speaker_label":"SPEAKER_0"}]}
```
6 speakers diarized; mapping timed out; final map = **1 entry**, and only because
deterministic stem self-attribution named the recorder independently of the agent.

**Meeting 2 — family meeting (`job-20260629110843838`, from `.wav`)**
```json
diarize:         {"segments":56,"speakers":4}
speaker_mapping: {"attendees":5,"input_chars":3610,"model":"agent:claude",
                  "outcome":"error","output_chars":0,
                  "reason":"Agent timed out","speaker_labels":3,"duration_ms":120106}
attribution:     {"diarization_from_stems":false,
                  "self_attribution":{"returned_some":false,
                                      "skipped_reason":"diarization-not-from-stems"},
                  "speaker_map":[]}
```
Not from stems, so the deterministic fallback was skipped too → `speaker_map: []`,
**zero** names.

[...]

Issue #362, opened 2026-06-25
Native in-app chat panel for Recall (replace embedded terminal for non-developer users)


maosuarez:
## Context

First, a bit of background: I'm a Windows contributor who's been using
Minutes as an end user. The app is genuinely excellent — the pipeline
from recording to structured markdown is smooth, and the MCP integration
is thoughtful. This is a "thinking out loud" proposal, not a complaint.
Happy to hear why the current approach is the right one.

---

## The current experience for non-developers

When a user opens the Recall panel, they're greeted by a full xterm.js
terminal running Claude Code. As a developer I can appreciate the
elegance — the assistant workspace is a real directory with
`CLAUDE.md` + `CURRENT_MEETING.md`, the MCP tools feed live context,
skills auto-sync. It's a well-engineered system.

But for a non-technical user (or even a technical user who just wants
to ask "what did we decide about the budget?"), the experience is jarring:

- The terminal appearance signals "developer tool" rather than "meeting assistant"
- The user needs to know what to type and how to interact with a CLI
- There's no visible conversation history in a natural format
- On Windows in particular, the PTY layer adds friction and the terminal
  aesthetic feels out of place inside a Tauri productivity app

The core intelligence is solid. The presentation is where non-developers get lost.

---

## The proposal: a lightweight native chat surface

What if the Recall panel rendered as a proper chat widget instead of an
embedded terminal — while keeping everything else the same?

Concretely, the panel would show:

- A message thread (user bubbles + assistant bubbles), styled in Geist Mono per the design system
- A plain text input at the bottom (no CLI knowledge required)
- Conversation history scoped to the current session

Under the hood, the implementation could work in two ways:

**Option A — thin wrapper over the existing stack**: keep spawning the
agent process, but pipe its output through a structured protocol
(e.g. streaming JSON or SSE) and render it as chat bubbles in the
Tauri webview. The assistant workspace, MCP tools, and skill mirrors stay exactly as-is.

**Option B — direct API calls from the Tauri backend**: the Rust backend
calls the LLM API directly (Claude, Gemini, or Codex via an API key the
user configures in Settings), passing transcript context as system prompt.
No agent process needed; the MCP tools that matter most (transcript read,
action item query, speaker lookup) get inlined as structured context injection.

Option A is lower risk — it's essentially a rendering change, not an
architecture change. Option B is more self-contained but requires
maintaining an HTTP client path alongside the existing agent-based one.

Either way, the change is **additive**: existing developer users who prefer
[...]

PR #404, opened 2026-07-03
feat(recall): native chat panel for Recall, replacing the embedded terminal by default (#362)

maosuarez:
# feat(recall): native chat panel for Recall, replacing the embedded terminal by default (#362)

*Draft PR body — not yet pushed or opened. For review only.*

---

## Summary

Replaces the xterm.js/PTY terminal as the default Recall panel UI with a native chat surface: message bubbles, a plain-text input, and no CLI knowledge required. The terminal is preserved as an opt-in "Developer mode" toggle in Settings (and a matching in-panel toggle) so existing power-user workflows don't regress.

Addresses #362.

## Motivation

Per the discussion on #362: the Recall panel was the one place in Minutes where the developer-tool plumbing was fully exposed to end users — a full xterm.js terminal running Claude Code, with no visible conversation history in a natural format. @silverstein confirmed Option A (thin presentation-layer wrapper, agent stack unchanged) as the direction, and specifically flagged that scraping the interactive TUI would be fragile — the non-fragile path is driving the agent headless with structured streaming JSON and rendering that as bubbles. This PR is that: `claude --output-format stream-json` piped and rendered incrementally, rather than attaching to the PTY.

## Changes

**Core (`crates/core/src/summarize.rs`)** — `e69ade8`
- New `ChatInvocation` + `build_chat_invocation()`, wrapping the existing `prepare_agent_invocation()` (the same lean invocation path #382 built for speaker mapping) instead of hand-rolling a separate claude argument list for chat.
- For claude, upgrades `--output-format text` to `stream-json` (+ `--verbose`) so the panel can render tokens incrementally; other CLIs are unaffected and keep single-shot stdout capture.
- Fixes a real bug along the way: the original hand-rolled invocation used `--no-interactive`, which doesn't exist on the claude CLI and made every chat message error.
- Prompt now always travels on stdin instead of as a `-p` argument, which also sidesteps the Windows command-line length limit on long transcripts.

**Backend (`tauri/src-tauri/src/commands.rs`, `main.rs`)** — `3ab5a0c`
- `cmd_recall_chat_send` now does provider-priority dispatch: Ollama (if configured) → Anthropic API (see "Open for discussion" below) → CLI agent via `detect_agent_cli()` → descriptive error if nothing is available.
- Context injection (capped at 12,000 chars): the full content of the meeting currently focused in the panel, plus up to 5 keyword-matched snippets from other meetings via the existing search index, so referential questions like "what did we discuss in this meeting?" work even without content-bearing keywords in the question itself.
- Conversation history (last 6 turns) is tracked in a new `AppState.recall_chat_history` and prepended to the prompt; `cmd_recall_chat_clear` resets it.
- stderr from the agent child process is now captured and surfaced as `recall-chat-error` instead of silently discarded, so auth failures / missing-binary errors are visible in the chat UI instead of an empty response.

**Frontend (`tauri/src/index.html`)** — `d00e860`
- Chat bubble UI in Geist Mono using existing CSS vars, plain-text input, welcome message.
- In-panel dev-mode toggle synced with the Settings toggle.
- Fixed 3 bugs found while testing on Windows:
  1. Welcome message rendered vertically centered instead of top-anchored (terminal container wasn't reliably hidden in chat mode).
  2. Terminal/chat split-view: a background PTY session's `pty:data` events were unconditionally re-revealing the terminal container even while chat mode was active — now gated on `recallDevMode`.
  3. Cross-meeting chat memory bleed: switching "Discuss with Agent" to a different meeting kept showing the previous meeting's bubbles and backend history. Now tracked via `recallChatMeetingPath` + `resetRecallChat()` — switching meetings resets the conversation, re-focusing the *same* meeting preserves it.

## Open for discussion — two deviations from the #362 plan, done deliberately, happy to rework

I want to flag these explicitly rather than let them slide by in the diff, since your comment on #362 was specific about both.

**1. Anthropic-direct-API fallback.** `cmd_recall_chat_send` includes an optional third path: if `ANTHROPIC_API_KEY` is set in the environment, it'll call the Messages API directly. I know you called this pattern "a trap for us specifically" and that it works against the no-API-key privacy story — I'm not trying to sneak Option B back in as the primary design. The reasoning for including it at all: without any agent CLI installed (no Claude Code, Codex, or Gemini CLI on PATH), a user currently gets a hard "no provider found" error and the chat feature is simply unusable. This fallback only activates if the user has explicitly set the env var themselves — it's meant as an accessibility path of last resort, not a redesign. The **primary, intended path remains the headless CLI-agent invocation** you asked for. If you'd rather this not exist at all in the first PR, say the word and I'll drop it — it's a self-contained ~40-line block and easy to remove without touching anything else.

**2. Lean/tool-less chat instead of the shared workspace with MCP/tools intact.** During testing, running the chat against the shared `~/.minutes/assistant/` workspace with full tools/MCP access caused a specific, visible bug: the model read that workspace's `CLAUDE.md` (written for the full-tool PTY assistant), tried to use tools it didn't actually have access to in this stdin/`--print` context, and narrated the failed attempts straight into the chat bubble — e.g. garbled "tool exited with code 1"-style text where a clean answer should be. The fix I landed is to run the chat lean (`--strict-mcp-config`, `--tools ""`, no MCP servers) from an isolated `~/.minutes/chat/` cwd, and instead ground every answer via explicit context injection (the focused meeting's content + keyword-matched snippets, pulled straight into the prompt) rather than letting the model reach for tools/files itself. This is a pragmatic stability tradeoff to ship something that works reliably for a first pass — it is not a rejection of keeping MCP/tools live, which I know was the point of the headless-over-TUI-scraping approach in the first place. If you'd rather solve the workspace-contamination problem a different way — e.g. a chat-specific `CLAUDE.md` that's honest about the lack of an interactive tool loop, instead of disabling tools outright — that seems like a very reasonable alternative and I'm glad to rework it in that direction.

## Cross-platform compatibility

Frontend changes are pure HTML/CSS/JS in the Tauri WebView (no platform-specific styling). Backend changes reuse the existing per-platform `.cmd`/`.bat` spawn handling already in place for Windows (npm-installed CLIs). No macOS/Linux-specific code paths were touched; the terminal/PTY path is unchanged for anyone using Developer mode.

## Testing

Manually verified on Windows 11 with `cargo tauri dev`:
- Opening Recall on a meeting shows the chat panel by default with a welcome bubble
- Sending a message streams tokens incrementally into an assistant bubble (claude CLI path)
- Asking a referential question ("what did we decide in this meeting?") correctly pulls in the focused meeting's content
- Switching "Discuss with Agent" to a different meeting resets the conversation; reopening the same meeting preserves it
- Toggling Developer mode (Settings and in-panel) switches to the terminal and back without a black screen or split view
- Simulated a missing-binary case (renamed `claude` off PATH) — chat surfaces a clear error instead of hanging

Not yet verified: Ollama path (don't have it running locally) and the Anthropic-API fallback path specifically (have a key configured but would appreciate a second pair of eyes given the concern above) — codex/gemini/opencode capture paths are exercised via the shared invocation builder's existing test coverage but not manually end-to-end.

## Checklist

- [x] `cargo test -p minutes-core --no-default-features` (893 unit + 7 integration passing, 0 failed)
- [ ] Maintainer sign-off on the two tradeoffs above before this is considered final


silverstein:
This is excellent work, and thank you for the care — especially preserving the terminal as an opt-in Developer mode. That was the thing I was most worried about losing, and you handled it exactly right. The headless `stream-json` approach (rather than scraping the TUI) is the direction I wanted, and catching the `--no-interactive` bug along the way is a nice bonus.

On your two flagged deviations — I really appreciate you surfacing them instead of letting them ride in the diff:

**1. Anthropic-direct-API fallback — please drop it from this PR.** You read my #362 comment right: this is the trap for us specifically. The no-API-key story is a core part of what Minutes is, and I'd rather the "no agent CLI installed" case be a clear, friendly "install Claude Code to enable chat" pointer than a quiet API path that erodes the promise. Since it's a self-contained ~40 lines as you noted, dropping it keeps this PR clean and on-message. We can revisit a truly-local fallback (Ollama is already in your priority chain) separately.

**2. Tools-less chat — let's go with the chat-specific CLAUDE.md instead of disabling tools outright.** Your instinct in the "reasonable alternative" is the one I'd take. The contamination bug you hit is real, but the fix of stripping tools entirely gives up part of what makes the Minutes assistant valuable (it can actually reach the graph, search, MCP). A chat-scoped CLAUDE.md that's honest about the non-interactive context — "you're answering in a single-shot chat, no interactive tool loop, ground answers in the injected context" — keeps the door open to progressively re-enabling the tools that *do* work in a stream-json context. I don't think you need to fully solve that in this PR; landing with the honest chat-CLAUDE.md (even if tools stay conservative for now) is a better foundation than a hard `--tools ""`.

**One logistics note, no action needed from you yet:** this branch was cut before a couple of things landed on main today — the Documents pane / calm viewer (#402) and dictation v1 (#403, merging now), both of which touch `index.html` and `commands.rs` heavily. So there'll be a rebase with real conflicts before this merges. Don't rush to do it right now — let #403 settle first, then I'll help coordinate the rebase so you're not fighting a moving target. Flagging it only so the eventual conflicts aren't a surprise.

Really strong contribution. Drop the API block, swing the tools deviation toward the honest chat-CLAUDE.md, and I think this is close.

silverstein:
Re-reviewed the revision — this is excellent. You addressed both of my points and better than I suggested: the Anthropic-API fallback is gone, and instead of the hard `--tools ""` lockout you scoped claude to the Minutes MCP server via `--strict-mcp-config` (read-only access to the user's own meeting data, nothing else). That's a genuinely nice resolution.

One maintainer decision on the **default**, and I want to flag it before I touch the branch rather than surprise you. Rather than chat-default-for-everyone (which changes the panel for existing power users) or demoting your chat to a hidden opt-in (which would bury the work), we're going with a **capability-based default**:

- **No agent CLI detected → default to your chat panel.** The terminal Recall panel needs an agent CLI to do anything, so for a user without one it's useless — chat is the only thing that works. This is exactly the #362 audience, and your panel becomes their default.
- **Agent CLI present → default to the terminal.** Power users (who have a CLI) keep the terminal; both remain switchable via the toggle you built.

Net effect: your chat stays the default *for the users it was built for*, the terminal is preserved for those who can use it, and nobody lands on a surface that doesn't function for them. It reuses the CLI detection your provider dispatch already does.

Because this touches the default behavior + needs a rebase against a heavy week of `index.html`/`commands.rs` changes, I'll **rebase onto main and wire the capability-based default myself** so you're not chasing a moving target, then merge. I'll keep your chat/terminal toggle and MCP-scoping exactly as-is. Shout if you'd rather drive it — otherwise I've got it from here. Thanks for a really solid contribution (and four this week).

silverstein:
Reviewed this end-to-end and it's genuinely strong work — thank you. The headless `stream-json` approach over TUI-scraping is exactly right, and you resolved both of your flagged deviations in the direction we'd want before I even got here: the Anthropic-direct fallback is gone (no-API-key stays core), and the tool-less lockout evolved into the **curated read-only MCP scope** — that's the better design. The trust-boundary reasoning on `CHAT_ALLOWED_MCP_TOOLS` (excluding `open_dashboard` for its HTTP-server/browser side effect, all write tools out) is careful and correct.

I pushed three commits to bring it current and fix one latent bug — all mechanical/safety, nothing touching your design:

1. **Merge with `main`** (`7d0db36`). Two conflicts, both clean add/add: your `build_chat_invocation` tests vs. #421's `prepare_agent_invocation_rejects_lean_with_screens` (kept both), and your Recall chat state vs. main's new sidebar-collapse logic in `index.html` (kept both).
2. **Signature fix.** #421 (merged today) added a `screen_files` param to `prepare_agent_invocation`; your non-claude chat branch now passes `&[]` (chat has no screenshots). Compiles + your 3 `build_chat_invocation` tests pass.
3. **`fix(recall): floor chat context truncations to UTF-8 char boundaries`** (`9d9c888`). The `&body[..6000]` / `raw[..12000]` context slices in `cmd_recall_chat_send` panic when a multi-byte char (accents/CJK/emoji) straddles the cut. Floored to a char boundary, matching `truncate_transcript` in core. This is the one real bug I found.

Verified: `cargo check -p minutes-app` green **on macOS** (your testing was Windows-only), core tests pass. (Our usual codex adversarial pass was down on my end — tool-host was timing out — so this was a careful manual review; flagging that honestly.)

**Non-blocking follow-ups** (fine as-is for this PR, noting for later): benign claude stderr can surface as a `recall-chat-error` even on success; `recall_chat_history` grows unbounded in memory until cleared (only the last 6 turns hit the prompt); and a very long prompt could in theory fill the stdin pipe buffer before the reader threads spawn — none are crashes and all are edge cases.

Last gate before merge is a **macOS click-test** of the chat panel (new `cmd_*` + `index.html` changes require it per our checklist). Building the dev app now. Once it passes, this merges. Really nice contribution. 🙌

silverstein:
Click-tested the chat panel live — it's genuinely great. One UX gap surfaced in testing: on the claude path there's a multi-second cold start (spawn `claude -p` → Minutes MCP server init → first inference) during which the empty streaming bubble only showed the tiny blink caret (`▮`), which reads as a stray cursor — so it felt like hitting enter did nothing.

Pushed a small CSS-only fix (`a76e561`): an empty streaming bubble now shows quiet pulsing dots (`:empty::before`) so the cold start reads as *working*, replaced by streamed text the instant the first chunk lands; the caret is gated to `:not(:empty)`. No behavior change. Stays within DESIGN.md's interaction doctrine ("prove the system heard the user", mono, whisper-don't-announce).

A natural follow-up (not in this change): during the MCP-grounding phase the dots could read "searching your meetings…" by parsing the `tool_use` stream events — but the dots already cover that window fine for v1.

silverstein:
Took the thinking indicator one step further into live grounding narration (`9fbe918`): the frontend already receives the full claude `stream-json`, so the empty streaming bubble now reads the `tool_use` events and shows *which part of the user's memory the agent is touching* — "searching your meetings", "checking commitments", "reading a meeting" — via a `data-status` attribute + CSS `attr()`. Starts at "thinking", falls back to "consulting your memory" for unmapped tools, and vanishes the instant the answer text streams in. Pure frontend (~40 lines JS + CSS), no backend change. Turns the cold-start dead air into a window onto the read-only MCP grounding — which also happens to showcase exactly what the local memory layer is doing.

Issue #415, opened 2026-07-06
process_audio hangs / times out unconditionally on Windows - independent of audio duration, no error, no log trace


Cedric-Duchene:
Summary

On Windows, every call to the MCP tool process_audio fails with a request timeout (MCP error -32001: Request timed out), regardless of the input audio file's duration. The failure is identical whether the file is 41 minutes or 10 seconds long, which rules out a duration-based timeout. No error, job record, or diagnostic trace of any kind is produced anywhere in Minutes' own state.

Environment

OS: Windows 11
Install: Claude Desktop MCP extension (ant.dir.gh.silverstein.minutes), version 0.19.0 (confirmed via manifest.json — already the latest published version, so this is not fixed by upgrading)
Whisper model: ggml-small.bin, 465MB, present at ~/.minutes/models/ and correctly sized (not truncated)
ffmpeg: installed, added to PATH, confirmed working standalone (used to split/trim test clips)
No config.toml exists at either ~/.config/minutes/config.toml or %APPDATA%\minutes\config.toml — so [summarization] engine = "agent" is not configured. This rules out an external LLM agent subprocess (e.g. shelling out to claude/pi/codex for summarization) as the cause.

Steps to reproduce

Call process_audio with a .mp3 file path, language: "fr".
Observe the call times out after some interval with no result.

Tested with:

Original 41-minute recording
A 6-minute segment (ffmpeg -f segment -segment_time 360 -c copy)
A 90-second clip (ffmpeg -t 90 -c copy)
A 10-second clip (ffmpeg -t 10 -c copy)

All four attempts timed out identically. Duration has no effect on the outcome.

Expected behavior

Transcription completes in a time roughly proportional to audio duration (per the project's own benchmarks), and/or a clear job record / error appears if something goes wrong.

Actual behavior

The MCP call itself returns MCP error -32001: Request timed out every time.
list_processing_jobs (with include_completed: true) returns {"jobs":[]} — no job is ever recorded, before, during, or after the failed call.
list_meetings returns {"meetings":[],"actions":[],"view":"dashboard"}.
~/.minutes/inbox/failed and ~/.minutes/inbox/processed are both empty.
~/.minutes/jobs contains only an archive folder and a .archive-initialized-v1 marker — no active/pending/failed job files from any of the attempts.
~/.minutes/logs/mcp-crash.log contains only server startup/lifecycle events (module-load-start, main-start, transport-created, transport-connected) — nothing at all related to process_audio invocations, successful or failed.

[...]

PR #416, opened 2026-07-06
feat(diag): stage breadcrumbs for the process pipeline to localize the Windows hang (#415)

silverstein:
## What

Adds immediate, flushed stage breadcrumbs to the `minutes process` / `process_audio` pipeline, written to `~/.minutes/logs/process-audio-trace.jsonl`. Each marker is `flush()` + `sync_all()`'d to disk the instant it's written, so if a stage hangs, the **last-completed marker stays on disk** — pinpointing exactly where.

Diagnostic-only: no change to decode/whisper/timeout behavior.

## Why (#415)

On Windows, `process_audio` hangs unconditionally with **zero trace** — no job record, no log line, no error. Root cause (investigated): the in-process native whisper call (model load / integrated VAD / inference) wedges on Windows where the parent can't observe or kill it, and MCP's `execFile` buffers stderr until exit, so nothing surfaces. The reporter (Cedric-Duchene) explicitly asked for Windows diagnostic logging to localize it — this is that instrument.

Stages emitted: `process.start → decode.start/done → whisper.model_load.start/done → whisper.vad_setup.start/done → whisper.full.start/done → transcribe.done → pipeline.write.start/done → pipeline.done`. Each carries input path, language, content_type, model_path, vad_enabled, audio samples/duration; `whisper.full.start` also records the internal `timeout_secs` (which confirms the ~331s internal deadline exceeds MCP's fixed 300s — why even a 10s file times out).

## Safety

Best-effort by construction: the writer opens with `let Ok(file) else { return }` (no panic), swallows flush/sync errors, and a trace failure never affects processing. Gated by `is_active()` so it only traces the process path; `MINUTES_TRACE=0` disables. Cross-platform, negligible overhead (a dozen lines per run).

## Verified

Real macOS run produced the full ordered trace (attached in the dev log). `cargo test -p minutes-core --no-default-features` green (incl. a new test asserting ordered markers), fmt + clippy clean.

## Next

Once this ships, the reporter runs a build and the trace names the wedge stage (model-load vs VAD vs inference) — then the targeted fix (likely gating VAD on Windows or moving whisper behind a killable worker with an honest timeout+error) is small. Tracked on #415.


Cedric-Duchene:
Thanks for the fast turnaround — ran both checks:

1. **`.wav` also hangs.** Converted the 10s test clip to `.wav` (bypassing ffmpeg entirely) and called `process_audio` on it directly. Same result: `MCP error -32001: Request timed out`, and `list_processing_jobs` / `list_meetings` both still show nothing afterward. This matches your read — the decode path looks innocent, the hang is downstream in whisper.

2. **`~/.minutes/logs/minutes.log` doesn't exist.** Only `mcp-crash.log` is present in that folder — no `minutes.log` at all on this machine.

Happy to test the stage-breadcrumb tracing once it's in a published release — building from source isn't something I can easily do on this machine, so I'll wait for the next version rather than build main myself. Let me know if there's anything else useful to check in the meantime.

PR #419, opened 2026-07-07
fix(desktop): speaker Confirm buttons do nothing — replace unsupported prompt() with in-app modal (#391)

### LINKED ISSUE #391: Clicking speaker 'confirm' buttons in meeting review window doesn't do anything
When reviewing in a meeting summary screen I see the identified speakers listed with medium confidence and `confirm` buttons. Clicking the `confirm` buttons appears to do nothing.

I only noticed these confirm speaker buttons this week, in v0.18.14 and now v0.19.0 -- can't confirm whether this issue has been present before that.

Image

And TIL that the `hide_from_screen_share` setting applies to taking screenshots too.

silverstein:
Fixes #391.

## Bug

Clicking the speaker "Confirm" buttons in the meeting detail view does nothing (reported by @rymalia, present in 0.18.14 and 0.19.0).

## Root cause

The Confirm handler's *only* input path was `window.prompt()` (`tauri/src/index.html`). Tauri's WKWebView returns `null` from `prompt()`, so `if (newName)` was always false and the handler silently no-op'd — exactly "clicking does nothing."

## Fix

Replaced the native `prompt()` with a reusable in-app input modal:
- `showInputModal({ title, label, value, confirmText }) → Promise<string|null>` — reuses the existing `about-overlay`/`about-card` pattern, DESIGN.md tokens only (accent used only for the input's focus ring, matching the app's global `:focus-visible`), Geist Mono input. Focuses+selects on open, Enter submits, Escape/backdrop/Cancel dismiss, trims, resolves `null` on empty, restores prior focus. `role="dialog"` + `aria-modal` + `aria-labelledby`.
- The Confirm button now `await showInputModal(...)`; the existing disabled/`Saving...`/`cmd_confirm_speaker` invoke/feedback/reload flow is untouched.

Also verified the invoke arg-names match the Rust command (`cmd_confirm_speaker(meeting_path, speaker_label, name)` ↔ JS `{ meetingPath, speakerLabel, name }`) — a param mismatch would have been a second silent-failure cause; it's correct.

## Testing

Inline JS parses (`node --check`); markup/wiring asserted. **Needs a maintainer click-test in the dev app** (per PRE-COMMIT: any index.html change): open a meeting with a medium-confidence speaker → click Confirm → modal opens → enter a name → confirms and saves. Dev app built for that.

## Follow-up (not in this PR)

`tauri/src/index.html` also has a `window.confirm()` for the restart flow — same class of WKWebView unreliability; worth a separate check.

### DIFF
--- a/tauri/src/index.html
+++ b/tauri/src/index.html
@@ -2060,6 +2060,52 @@
       gap: 8px;
     }
 
+    .input-modal-card {
+      display: flex;
+      flex-direction: column;
+      gap: 14px;
+      width: min(42ch, 100vw - 32px);
+      padding-top: 18px;
+    }
+
+    .input-modal-title {
+      font-family: var(--font-display);
+      font-size: var(--text-xl);
+      color: var(--text);
+    }
+
+    .input-modal-field {
+      display: grid;
+      gap: 8px;
+    }
+
+    .input-modal-label {
+      font: 500 var(--text-xs) var(--font-mono);
+      color: var(--text-secondary);
+    }
+
+    .input-modal-input {
+      width: 100%;
+      border: 1px solid var(--border-mid);
+      border-radius: var(--radius);
+      background: var(--surface-fill);
+      color: var(--text);
+      font: 13px var(--font-mono);
+      padding: 8px 10px;
+      outline: none;
+    }
+
+    .input-modal-input:focus-visible {
+      outline: 2px solid var(--accent);
+      outline-offset: 2px;
+    }
+
+    .input-modal-actions {
+      display: flex;
+      justify-content: flex-end;
+      gap: 8px;
+    }
+
     /* Markdown rendered styles */
     .md-viewer-content h1,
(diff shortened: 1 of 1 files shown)


silverstein:
Verified functionally without a manual click-test: added jsdom coverage (`crates/mcp` vitest) driving the real modal code. **8/8 pass** — modal opens + focuses, Confirm resolves the trimmed value, Enter submits, Cancel/Escape/backdrop resolve null, empty resolves null, focus restores — plus an end-to-end assertion that clicking a speaker Confirm button opens the modal and invokes `cmd_confirm_speaker` once with `{ meetingPath, speakerLabel, name }`. Full MCP suite green (71 tests). Since the fix swaps a WKWebView-unsupported native `prompt()` for standard DOM (which behaves identically in jsdom and WKWebView), this is high-fidelity; residual manual surface is pixel rendering only.

Issue #401, opened 2026-07-03
Title generator can echo its own instructions as the meeting title


silverstein:
Real occurrence (2026-07-02, maintainer machine): a meeting was saved as
`2026-07-02-i-ll-create-a-concise-meeting-title-mat.md`. The LLM replied to the title prompt conversationally ("I'll create a concise meeting title...") and the first line was taken verbatim as the title, then slugified into the filename.

Source: the title prompt in `crates/core/src/summarize.rs` (~line 251, "You create concise meeting titles..." / `refine_title`). There is no guard between "model output" and "becomes the filename".

Suggested fix (small, deterministic):
1. Reject candidate titles that look like instruction echo or meta-talk: starts with "I'll" / "I will" / "Here is" / "Sure", contains "concise meeting title" or other phrases from the prompt itself, ends with a colon, or is a question about the task.
2. On rejection, retry once with a stricter prompt, then fall back to the existing calendar/context-derived title path (the machinery `generate_title` already has).
3. Log a `title_generation` outcome (ok / rejected_echo / fallback) in the JSONL log so occurrences are greppable, in line with the honest-diagnostics pattern from #395/#396.
4. Unit tests with a few echo-shaped outputs.

Also worth noting: the meeting rename recovery already exists (`minutes rename`), so affected files are fixable by hand; this is about stopping the leak.

PR #421, opened 2026-07-09
fix(screen-context): retain screenshots and deliver them to agent summarizers

rymalia:
> Draft status while my user testing continues.

## Summary

Screenshots captured via `screen_context` were (1) deleted even with `keep_after_summary = true`, and (2) never delivered to the `agent`/`auto` summarization engines at all. Four commits: the retention fix, capability-gated agent delivery, coverage-bounded sampling, and an invariant guard for lean mode.

No configuration or schema changes. Gemini, OpenCode, and Pi remain text-only; direct-API screenshot routing is unchanged.

## 1. Retention fix (`jobs.rs`)

`preserve_audio_alongside_output()` ran `remove_dir_all` on the recording's screens dir unconditionally on job completion. The pipeline's own cleanup honors `keep_after_summary`; this second deletion site didn't, so kept screenshots vanished moments after the pipeline preserved them. Now guarded, with a regression test covering both branches.

## 2. Agent-CLI delivery (`summarize.rs`)

`screen_files` previously never reached the agent path — screenshots were only used by the direct-API engines (base64 vision, introduced in #25). Delivery is now per-agent, matched to verified headless capability:

| Agent | Delivery |
|---|---|
| claude | prompt section naming the screenshot dir/files (with meeting-time labels) + `--allowedTools Read --add-dir <dir>` |
| codex | native `--image <file>` attachments + positional time labels |
| gemini / opencode / pi | none — text-only, unchanged |

The claude flags were verified live: headless `-p` needs *both* — the cwd sandbox otherwise blocks `~/.minutes` reads and the run silently degrades to text-only. pi runs `--no-tools`; gemini/opencode headless file access to `~/.minutes` is unverified — instructing an agent to open files it can't reach silently degrades the summary, so those stay text-only until proven out.

## 3. Coverage-bounded sampling

Replaces "first 8 chronologically" (which covered only the opening minutes of a meeting) with even positional sampling across the captured set for the agent path, governed by one rule: **never deliver images from time ranges whose transcript was omitted from the model call.** The 100k-byte transcript cap now cuts at the last complete line and bounds selection by its final `[M:SS]` stamp (truncated transcripts without parseable timestamps deliver no images); truncation logs a warning instead of silently dropping the meeting tail. Frames carry meeting-time labels parsed from capture filenames.

**The API path is deliberately unchanged**: it attaches images to the first transcript chunk only, so whole-meeting sampling there would pair late-meeting images with early-meeting text. Chunk-aware delivery is a follow-up.

## 4. Lean-mode guard

Upstream lean mode (#383, fixing #382) runs claude with no MCP servers and no tools — incompatible by construction with screen delivery, which requires the Read tool. The combination is now rejected with an error rather than shipping a prompt claude can't follow. All existing callers are unaffected (speaker mapping is lean and text-only; summarization passes screens with `lean=false`).

## Prompt-injection hardening

The system prompt's source policy now covers screenshots, and both screen preambles (agent-CLI and the four direct-API engines, unified into one constant) instruct that text visible inside images is meeting content, never instructions. Related: #401 tracks the same class of issue in the title-refinement path — not fixed here; natural follow-up.

These preambles guide model behavior but are not an authorization boundary. The Claude invocation grants `Read` access to the screenshots directory, which also contains frames not selected for the prompt. Staging only selected frames in an isolated directory is a candidate follow-up hardening.

## Known limitations / follow-ups

- Chunk-aware API delivery (unlocks whole-meeting sampling + transcript correlation for API engines).
- Agent map/reduce chunking for >100k-byte transcripts (the cap currently drops the tail with a logged warning; surfacing that warning in user-facing status is also open).
- Kept screenshots remain keyed by the jobs-dir recording stem; `remove_capture_artifacts()` derives a different directory after the audio moves (pre-existing lifecycle gap).
- Capture stops after 60 successful screenshots; duration depends on the configured interval (30 minutes at the default 30s).
- Sampling is positional across available captures, not time-targeted.

## Testing

23 new unit tests: retention both branches; per-agent invocations including the lean interplay and rejection; sampling/bounding including a >100k-transcript regression test; truncation line-boundary behavior; meeting-time label formats; injection-policy coherence; instruction/invocation delivery-sync consistency; capture-filename parsing; notes/transcript budget independence and the notes-stamp coverage regression; fail-closed selection for unstamped truncated transcripts.

Verified with:

```
cargo test -p minutes-core                                  # 951 lib + 10 integration tests pass
cargo clippy -p minutes-core --tests -- -D warnings
cargo clippy --all --no-default-features -- -D warnings
cargo fmt --all -- --check
```

### Update 2026-07-10 — post-review coverage-rule fixes

Two gaps in the coverage invariant found in continued auditing, fixed in the latest commit:

- **User notes are budgeted independently of the transcript.** Notes share the `[M:SS]` stamp format and were concatenated ahead of the transcript before the 100k byte cap, so bulky notes could starve the transcript budget, and a late-stamped note could authorize screenshots from truncated (undelivered) transcript ranges. Notes now get their own 10k cap, the transcript keeps its full 100k budget, and the coverage bound derives from delivered transcript text only. Non-agent engines receive byte-identical input.
- **Truncated + unstamped transcripts now deliver zero images** instead of first-N — the delivered time range is unknowable, so selection fails closed and the coverage rule is absolute rather than best-effort.

+3 unit tests (notes-stamp coverage regression; bulky-notes independence; fail-closed selection); `site/lib/release.ts` regenerated for the test-count consistency gate.

### DIFF
--- a/crates/core/src/jobs.rs
+++ b/crates/core/src/jobs.rs
@@ -1056,7 +1056,7 @@ fn terminal_state_for_artifact(artifact: &pipeline::TranscriptArtifact) -> JobSt
 /// Move the captured audio alongside the output markdown so users can reprocess later.
 /// e.g. ~/meetings/2026-04-02-standup.md → ~/meetings/2026-04-02-standup.wav
 /// or, for native call captures, ~/meetings/2026-04-02-call.mov.
-fn preserve_audio_alongside_output(job: &ProcessingJob) {
+fn preserve_audio_alongside_output(job: &ProcessingJob, config: &Config) {
     let Some(ref output_path) = job.output_path else {
         return;
     };
@@ -1088,10 +1088,14 @@ fn preserve_audio_alongside_output(job: &ProcessingJob) {
         use std::os::unix::fs::PermissionsExt;
         fs::set_permissions(&audio_dest, fs::Permissions::from_mode(0o600)).ok();
     }
-    // Clean up any screen capture artifacts left in the jobs dir
-    let screens_dir = crate::screen::screens_dir_for(&audio_src);
-    if screens_dir.exists() {
-        fs::remove_dir_all(screens_dir).ok();
+    // Clean up any screen capture artifacts left in the jobs dir, unless the
+    // user opted to keep screenshots after summarization. Without this guard we
+    // would delete screenshots the pipeline deliberately preserved.
+    if !config.screen_context.keep_after_summary {
+        let screens_dir = crate::screen::screens_dir_for(&audio_src);
+        if screens_dir.exists() {
+            fs::remove_dir_all(screens_dir).ok();
+        }
     }
     // Update the job record so audio_path points to the new location
     let dest_str = audio_dest.display().to_string();
@@ -1398,7 +1402,7 @@ where
                 // Run post_record hook (async, non-blocking)
                 pipeline::run_post_record_hook(config, &result.path);
                 if completed_job.state == JobState::Complete {
-                    preserve_audio_alongside_output(&completed_job);
+                    preserve_audio_alongside_output(&completed_job, config);
                 }
                 // Reload job after preserve may have updated audio_path
                 let final_job = load_job(&completed_job.id).unwrap_or(completed_job);
@@ -1697,7 +1701,7 @@ mod tests {
             };
             write_job(&job).unwrap();
 
-            preserve_audio_alongside_output(&job);
+            preserve_audio_alongside_output(&job, &Config::default());
 
             let preserved_audio = output_path.with_extension("wav");
             let preserved_stems = crate::capture::stem_paths_for(&preserved_audio).unwrap();
@@ -1756,7 +1760,7 @@ mod tests {
             };
             write_job(&job).unwrap();
 
-            preserve_audio_alongside_output(&job);
+            preserve_audio_alongside_output(&job, &Config::default());
 
             let preserved_audio = output_path.with_extension("mov");
             let preserved_stems = crate::capture::stem_paths_for(&preserved_audio).unwrap();
@@ -1767,6 +1771,74 @@ mod tests {
             assert!(!audio_path.exists());
             assert!(!stems.voice.exists());
             assert!(!stems.system.exists());
+        });
+    }
+
+    #[test]
+    fn preserve_audio_alongside_output_respects_keep_after_summary() {
+        with_temp_home(|tmp| {
+            let jobs_root = jobs_dir();
+            fs::create_dir_all(&jobs_root).unwrap();
+
+            // Run the preserve step for one job per config value; return
+            // whether that job's screens dir survived.
+            let mut screens_dir_survives = |job_id: &str, keep: bool| {
+                let audio_path = jobs_root.join(format!("{job_id}.wav"));
+                fs::write(&audio_path, b"mixed").unwrap();
+
+                let screens_dir = crate::screen::screens_dir_for(&audio_path);
+                fs::create_dir_all(&screens_dir).unwrap();
+                fs::write(screens_dir.join("0001.png"), b"png").unwrap();
+
+                let output_path = tmp.path().join(format!("meetings/{job_id}.md"));
+                fs::create_dir_all(output_path.parent().unwrap()).unwrap();
+                fs::write(&output_path, "# out").unwrap();
+
+                let job = ProcessingJob {
+                    id: job_id.into(),
+                    mode: CaptureMode::Meeting,
+                    content_type: ContentType::Meeting,
+                    title: Some("keep screens".into()),
+                    audio_path: audio_path.display().to_string(),
+                    output_path: Some(output_path.display().to_string()),
+                    state: JobState::Complete,
+                    stage: None,
+                    created_at: Local::now(),
+                    started_at: None,
+                    finished_at: None,
+                    notice_dismissed_at: None,
+                    recording_started_at: None,
+                    recording_finished_at: None,
+                    context_session_id: None,
+                    user_notes: None,
+                    pre_context: None,
+                    consent: None,
+                    consent_notice: None,
+                    calendar_event: None,
+                    template_slug: None,
+                    recording_health: None,
+                    word_count: None,
+                    error: None,
+                    owner_pid: None,
+                    retry_count: 0,
+                };
+                write_job(&job).unwrap();
+
+                let mut config = Config::default();
+                config.screen_context.keep_after_summary = keep;
+                preserve_audio_alongside_output(&job, &config);
+
+                screens_dir.exists()
+            };
+
+            assert!(
+                screens_dir_survives("job-keep-screens", true),
+                "screens dir must survive when keep_after_summary is on"
+            );
+            assert!(
+                !screens_dir_survives("job-drop-screens", false),
+                "screens dir must be cleaned up when keep_after_summary is off"
+            );
         });
     }
 
--- a/crates/core/src/jobs.rs
+++ b/crates/core/src/jobs.rs
@@ -1782,7 +1782,7 @@ mod tests {
 
             // Run the preserve step for one job per config value; return
             // whether that job's screens dir survived.
-            let mut screens_dir_survives = |job_id: &str, keep: bool| {
+            let screens_dir_survives = |job_id: &str, keep: bool| {
                 let audio_path = jobs_root.join(format!("{job_id}.wav"));
                 fs::write(&audio_path, b"mixed").unwrap();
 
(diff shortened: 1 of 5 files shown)


rymalia:
**Test-drive report: end-to-end verification on real recordings (2026-07-09/10)**

Ran multiple real dogfood recordings on this branch (CLI- and app-initiated, macOS, `screen_context.enabled=true, interval_secs=15, keep_after_summary=true`, `summarization.engine="agent"` with `agent_command="claude"`). Everything this PR ships now works, verified end-to-end:

- **Retention fix**: screenshots survive job completion with `keep_after_summary=true` (previously deleted by the unconditional cleanup in `preserve_audio_alongside_output`).
- **Agent delivery**: the claude CLI received `--allowedTools Read --add-dir <screens dir>` and opened all 8 delivered frames — verified from the raw session transcript of the summarizer run: 9/9 Read calls returned native image blocks to the model (~3.3 MB of pixels, zero errors), and the summary wove in on-screen details that were never spoken aloud.
- **Coverage-bounded sampling**: 8 frames delivered out of 21 captured, evenly spread across the full meeting (0:15 → 5:23 of a ~5:20 recording), each labeled with its meeting-time offset.
- **Injection hardening**: text visible inside screenshots (including instruction-looking chat content on screen) was treated as content to describe, never as instructions to follow.

Two pre-existing issues (not introduced here) surfaced during testing, filed separately: #423 (`minutes stop` leaves the screen-capture thread running — hang + post-meeting captures) and #424 (Screen Recording permission failure is completely silent). The second one cost days of confused dogfooding — recordings looked perfectly healthy while producing zero screenshots.

**Observation + small enhancement proposal** on the known follow-up about kept screenshots detaching from the meeting stem: after the final title-based rename of the meeting files, the screenshots remain under their `job-<id>` directory name, so nothing ties a meeting back to its frames. Since the folder is now a first-class, user-visible artifact (that's the point of `keep_after_summary`), the summary output should say where it lives — mirroring how the recording's `.wav` path is handled:

- a line in the meeting markdown body, e.g. `Screenshots collected during this recording: ~/.minutes/screens/job-<id>/`
- and a frontmatter field (e.g. `screens_dir:`) so tooling and follow-up features (keyframe selection, reprocessing) can find the frames without guessing stems.

_(screenshot: Finder view of `~/.minutes/screens/` mid-recording — `current/` actively filling while previous sessions persist under their `job-<id>` names — attached below)_
screenshots-underway-in-screens-folder

silverstein:
This is really nice work — reviewed the diff and the shape is right:

- **Retention fix**: guarding the second `remove_dir_all` in `preserve_audio_alongside_output` on `keep_after_summary` is exactly correct — that duplicate deletion site was silently undoing the pipeline's own honored preservation. Threading `config` in for it is clean.
- **Agent delivery**: I especially like that you added a prompt-injection guard for *text visible in the screenshots*, not just the transcript — that's a real threat surface most people miss. The byte-capped transcript + coverage-bounded sampling keeps prompts sane.
- 24 tests across the branches is great.

No concerns from the read. Since it's marked draft while you finish user testing, I'll hold off merging — just ping here when you're confident and I'll re-review the final state and take it through. Thank you for the depth (and for the three lifecycle issues alongside it).

silverstein:
Saw the new commits + your dogfood report (real CLI- and app-initiated recordings with agent summarization) — that's exactly the kind of end-to-end validation this needed, and budgeting the user notes independently of the transcript is a good catch. No rush from my side; whenever you flip it out of draft I'll do a final re-review of the delta and take it through. Thanks for the thoroughness.

rymalia:
**A design reflection after enabling the screenshot feature and user testing: where's the privacy line?**

Having now used this end-to-end for a couple of days, I want to flag a tension I didn't fully appreciate when building it. Minutes' core promise is *local* processing of personally sensitive media — whisper and parakeet exist so that voice recordings never leave the device. Sending transcripts to a cloud summarizer is a deliberate, defensible exception: a transcript is already one abstraction layer away from the raw recording, prevents voiceprinting, and contains what people knowingly said in a recorded conversation.

Raw screenshots don't have either property. They're equally sensitive media this product touches — they could capture notification banners, message previews, sensitive or draft documents, other people's faces and shared screens, and whatever else crosses the desktop (as #423 showed, sometimes even after the meeting ends). So while I'm genuinely happy with how well this MVP of screenshots incorporated into the meeting summary pipeline works, "send the raw frames to a cloud model" shouldn't be the only processing mode — and probably shouldn't be the default posture long-term.

Two conclusions I'm working toward:

1. **Automated raw screen captures are valuable even with no AI at all.** Just having the folder to flip through next to the summary recovers things no transcript holds — which slide was up, the mood in the video gallery, my own desktop trail during a phone call (pulling up someone's LinkedIn mid-conversation, digging for a doc I referenced). This is why I was excited to get this feature working. And it's arguably better than a full-video recording of the meeting — easier to quickly scan through to jog your memory. Retention isn't just corpus-building for the future keyframe selector; it's a feature. This strengthens the case for linking the screenshots folder from the meeting markdown (per my earlier comment).

2. **The delivery knob should be a three-position switch**, something like `screen_context.processing = "save-only" | "local" | "cloud"`. `save-only` keeps everything on disk and sends nothing anywhere. `local` is the interesting one, analogous to whisper & parakeet: pixels→text on device — maybe some deterministic keyframe extraction / scene-change detection first, maybe an OCR pass (macOS Vision is on-device and screen content is mostly text; notably, the most impressive extractions in my test summaries were surprisingly all on-screen text), local VLM captions/descriptions as a later layer — so the cloud summarizer only ever sees text, the same abstraction layer as transcripts. `cloud` stays as what this PR ships: the opt-in maximum-quality tier. Same shape as transcription: whisper/parakeet made local speech-to-text a reality; I think the goal is the same for screen understanding.

None of this changes the code in this PR — cloud delivery remains correct as the explicit opt-in it already is. Flagging it so the follow-up direction is on record before this lands. That being said, this phase is tested and **ready for review** — post-review coverage fixes are documented in the description's *Update 2026-07-10* section, the branch is synced with `main` after #427/#422 landed, and CI is green. Marking ready now.

silverstein:
This is really strong work — the coverage invariant is handled carefully (bound derived from the truncated transcript only, never the notes block), the delivery-vs-instruction sync test is exactly the guard I'd want, and the fail-closed behavior is the right default. I ran it through an adversarial review pass (Codex, gpt-5.5) alongside my own read. Retention fix, capability gating, subprocess argv construction, and panic-safety all came back clean.

One real gap surfaced, and it's worth fixing before merge because it silently defeats the feature on its most important case:

**`last_transcript_stamp_secs` doesn't parse diarized stamps.**

When diarization runs, `apply_speakers` rewrites every line as `[SPEAKER_0 0:00] text` (see `diarize.rs`, e.g. the `[SPEAKER_1 0:05]` assertions around line 2896). But the parser assumes a bare `[M:SS]`:

```rust
let (stamp, _) = rest.split_once(']')?;   // stamp = "SPEAKER_0 0:00"
let (mins, secs) = stamp.split_once(':')?; // mins = "SPEAKER_0 0"  -> parse fails -> None
```

So for a **diarized transcript that exceeds the 100k cap**, every line returns `None`, `select_agent_screen_files` fails closed, and claude/codex get *zero* screenshots — degrading to text-only exactly on long meetings, which is where screenshots earn their keep. It fails safe (never violates the "no images past coverage" invariant), so it's not a correctness bug — but it quietly turns the feature off for the common long + diarized path.

Localized fix — take the last whitespace token of the bracket contents before splitting, so both `[M:SS]` and `[SPEAKER_N M:SS]` parse:

```rust
let (stamp, _) = rest.split_once(']')?;
let stamp = stamp.split_whitespace().last().unwrap_or(stamp); // tolerate `[SPEAKER_N M:SS]`
let (mins, secs) = stamp.split_once(':')?;
```

Worth a companion test with a `[SPEAKER_0 M:SS]` transcript in `select_agent_screens_respects_truncation_bound_*` so it doesn't regress. That's the only thing standing between this and merge from my side — everything else is good to go. Thanks for the thorough tests and the injection hardening.

rymalia:
Fixed in 2cb625c. `last_transcript_stamp_secs` now parses the final whitespace-delimited timestamp inside both `[M:SS]` and diarized `[SPEAKER_N M:SS]` / attributed-speaker brackets. Added parser coverage plus a long diarized-transcript regression proving screenshots remain bounded but non-empty.

silverstein:
Fix is spot-on — `split_whitespace().last()` handles both bare and diarized stamps, and the `[Ryan Malia 3:21]` test case is a nice touch (confirms multi-word attributed names resolve too). Rust CI green across all three platforms. Thanks for the careful work on this one — the coverage invariant, the injection hardening, and the delivery-sync test are exactly the kind of rigor that makes this easy to merge. Shipping it. 🚀

PR #427, opened 2026-07-10
fix(screen-context): stop the screen-capture thread on every record-loop exit; make Screen Recording permission failures visible

### LINKED ISSUE #423: `minutes stop` leaves the screen-capture thread running: recorder hangs in join while screenshots continue post-meeting
**Summary**: Stopping a recording via the `minutes stop` CLI (stop-sentinel path) never signals the screen-capture thread. The recorder's main thread then blocks in `ScreenCaptureHandle::drop → join()` while the thread keeps screenshotting the desktop every `interval_secs` — after the meeting has ended — until it hits `MAX_SCREENSHOTS=60` (~15 min at the default 15s interval). Privacy issue plus an apparently-hung recorder process.

**Mechanism**:
- `crates/core/src/capture.rs` (~line 1636): the record loop exits on the stop sentinel with `break`, **without setting `stop_flag`**.
- The capture thread in `crates/core/src/screen.rs` only watches `stop_flag` (or the 60-frame cap).
- `ScreenCaptureHandle::drop` (`screen.rs` ~line 260) is a bare `join()` with no stop signal of its own.

**Observed** (real run, 2026-07-09): recording stopped at 1:51; screenshots continued past frame 33 (8:45+ elapsed); the main thread was sampled sitting in `_pthread_join`; sending SIGINT (the Ctrl-C path, which *does* set `stop_flag`) unstuck it immediately and processing completed normally. The desktop app's own stop button is unaffected — it sets `stop_flag` in-process (`tauri/src-tauri/src/commands.rs`, `request_stop`).

**Suggested fix**: set `stop_flag` when the stop sentinel is detected (both sentinel check sites), and/or give `ScreenCaptureHandle` its own `AtomicBool` that `drop` sets before joining, so the handle is safe regardless of the caller's flag discipline.

**Interaction with PR #421**: pre-existing bug (not introduced there), but that PR's retention fix means the post-meeting frames now persist to disk, so the privacy surface is larger once it lands.

### LINKED ISSUE #424: Screen-context: Screen Recording permission failure is completely silent (stderr-only; System Settings shows stale grant as enabled)
**Summary**: When `check_screen_permission()` (`crates/core/src/screen.rs`) fails, the only diagnostic is an `eprintln!` to stderr — invisible when the desktop app records (GUI stderr is discarded). Nothing reaches `~/.minutes/logs/minutes.log`, there is no UI signal, and recordings proceed normally minus screenshots. The user has no way to discover why screenshots aren't happening.

**Why this bites hard on macOS**: macOS never re-prompts for Screen Recording (unlike microphone). After an app re-sign — routine during development — the stale TCC grant *still shows as enabled* in System Settings while being silently denied. Every layer that could tell the user something is wrong stays quiet.

**Repro**:
1. Re-sign (or rebuild ad-hoc) the app so its code-signing identity changes.
2. Record with `screen_context.enabled = true`.
3. Transcript and summary are produced normally; zero screenshots; zero log evidence. (Observed across four real recordings on 2026-07-09 before the cause was found.)

User-side fix once diagnosed: toggle the app off/on under System Settings → Privacy & Security → Screen & System Audio Recording (or remove with **−** and re-add with **+**).

**Suggested fixes** (any subset):
- Emit a structured log line at recording start: `screen_context` requested / permission-denied / started (the information already exists; today it only goes to stderr).
- Surface the denial in the app UI — a badge on the recording indicator, or a health-check row in Settings alongside the other capability checks.
- Docs note for contributors: changing the signing identity invalidates TCC grants; Screen Recording must be re-granted manually because macOS will never re-prompt for it.

rymalia:
Fixes #423. Fixes #424 — grouped as discussed on both issues, since they share `screen.rs`.

### #423 — capture thread outliving the recording

- `ScreenCaptureHandle` owns a second stop flag, set in `Drop` before the `join()`, so no caller exit path can leak the thread. This also covers the safety-guard auto-stop and reconnect-failure breaks, which had the same latent leak as the sentinel path.
- The stop-sentinel breaks in both record loops now set the shared `stop_flag`, so the live sidecar and streams wind down the same as on Ctrl+C.
- Regression test uses a long interval so the thread stays in its pre-first-capture wait — the test never captures a real frame.
- Verified against the original repro: a live recording stopped via `minutes stop` exited within seconds (previously: blocked in `pthread_join` while screenshotting the post-meeting desktop toward the 60-frame cap).

### #424 — silent permission failure

Screen Recording failures are now visible through four surfaces (was: one `eprintln!` that GUI launches discard):

- `~/.minutes/logs/minutes.log` — structured error (step `screen_context`)
- `~/.minutes/events.jsonl` — new `screen_context.unavailable` event, available to agents and future event consumers
- a desktop notification at recording start (reusing the safety-guard notification path) — the immediate user-visible signal for the app's own failed probe, which is exactly the stale-grant case
- `minutes health` — new "Screen recording" item that probes with a real capture attempt, catching the stale grant that System Settings displays as enabled. It validates the environment it runs in (from a terminal: CLI recordings — the app path is covered by the recording-time notification above); on Linux it checks for a screenshot tool, and on unsupported platforms it reports that honestly instead of promising screenshots.

All messages name the recovery explicitly (toggle off/on; macOS never re-prompts for Screen Recording). The duplicated startup block in the two record paths is factored into `start_screen_capture_if_enabled`, and a contributor-docs section documents the mic/screen TCC asymmetry.

**Testing**: 931 core lib tests (3 new) and the CLI suite pass; clippy incl. `--tests` at `-D warnings` (default and `--all --no-default-features`) and `rustfmt --check` clean; `site/lib/release.ts` regenerated for the new test count. Two Tauri tests fail identically on clean `upstream/main` in my environment — pre-existing/environmental, not from this change.

### DIFF
--- a/crates/core/src/capture.rs
+++ b/crates/core/src/capture.rs
@@ -1196,36 +1196,7 @@ fn record_to_wav_dual_source(
     );
     emit_recording_started(started_context);
 
-    let _screen_handle = if config.screen_context.enabled {
-        if !crate::screen::check_screen_permission() {
-            eprintln!("[minutes] Screen context disabled — grant Screen Recording permission in System Settings > Privacy & Security");
-            None
-        } else {
-            let screen_dir = crate::screen::screens_dir_for(output_path);
-            match crate::screen::start_capture(
-                &screen_dir,
-                std::time::Duration::from_secs(config.screen_context.interval_secs),
-                Arc::clone(&stop_flag),
-            ) {
-                Ok(handle) => {
-                    eprintln!(
-                        "[minutes] Screen context capture enabled (every {}s)",
-                        config.screen_context.interval_secs
-                    );
-                    Some(handle)
-                }
-                Err(e) => {
-                    tracing::warn!(
-                        "screen capture init failed: {} — continuing without screen context",
-                        e
-                    );
-                    None
-                }
-            }
-        }
-    } else {
-        None
-    };
+    let _screen_handle = start_screen_capture_if_enabled(config, output_path, &stop_flag);
 
     let preflight_intent = config
         .recording
@@ -1254,6 +1225,9 @@ fn record_to_wav_dual_source(
 
         if crate::pid::check_and_clear_sentinel() {
             tracing::info!("stop sentinel detected — stopping dual-source recording");
+            // Set the shared flag so everything watching it (screen capture,
+            // live sidecar, streams) winds down like on Ctrl+C (#423).
+            stop_flag.store(true, Ordering::Relaxed);
             break;
         }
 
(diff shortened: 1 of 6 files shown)


silverstein (approved):
Reviewed — this is exactly right, and clean:

**#423**: making `ScreenCaptureHandle` own its stop flag and set it in `Drop` before join is the self-contained fix I hoped for — `should_stop() = external || own` means the thread winds down on drop regardless of which caller break-path fired (sentinel, safety-guard, anything future). And setting `stop_flag` on the sentinel/guard breaks in capture.rs means everything else watching that flag stops promptly too. The regression test (1-hour interval, drop, assert prompt stop without the external flag) proves the load-bearing behavior directly.

**#424**: three signals (structured `log_error` → minutes.log, the new `screen_context.unavailable` event → events.jsonl, and the desktop notification) is the right honest-degradation shape — even if the notification is missed, the log + event carry the truth, and the event lets agents react live. Nice.

CI green (all platforms), 5 tests, no concerns. Merging — thanks for the precise diagnosis-to-fix on both.

PR #526, opened 2026-07-23
feat(resummarize): re-run the AI pass on edited meeting/memo transcripts

rymalia:
Implements #523 — re-run the AI pass on an edited meeting/memo transcript,
per the maintainer-approved plan (`docs/plans/rerun-ai-pass-resummarize-2026-07-21.md`,
maintainer notes M1–M5 folded in).

## What this adds

- **`minutes-core::resummarize::resummarize_meeting`** — re-runs the
  summarize stage + local derivations (action items, decisions, intents,
  entity links) against the *current* transcript text and splices the
  regenerated AI-owned sections back in under a strict safety contract:
  - **Never destroy on failure.** Engine `none`, provider errors, empty
    output, scaffolding-only output, and unstructured-regression output are
    hard no-write failures.
  - **Concurrent-edit guard.** Full-content snapshot before inference,
    re-compared before the backup copy and again immediately before the
    atomic rename.
  - **Splice, don't rewrite.** Only AI-owned body sections
    (`## Summary` … `## Commitments`) and derived frontmatter change;
    `## Notes`, `## Transcript`, `speaker_map`, capture/consent metadata are
    preserved byte-for-byte / field-for-field. Fence-aware, fail-closed
    section parser (duplicate headings reject the write).
  - **Status-preserving merge (M2).** Exact normalized identity only — never
    fuzzy. User-curated state (`status: done`, due dates, decision
    authority/supersession) carries forward; everything ambiguous, unmatched,
    or dropped is surfaced in the preview, never silently resolved.
  - **Timestamped backups.** Each apply writes
    `.<name>.pre-resummarize.<unix-secs>.bak` (mode-preserving); the newest
    3 per artifact are kept.
- **CLI: `minutes resummarize <meeting|memo|search-term>`** — preview by
  default (the preview **does invoke the model**, stated in the output per
  M3); `--apply`, `--engine`, `--template`, `--json` (errors also emit a JSON
  envelope). Serves `minutes import text` archives (#516) as their first AI
  pass.
- **Run record (M4):** `summarization:` health block shaped like the existing
  `speaker_mapping` block (model, template, duration_ms, last_run).
- **Engine choke point (M1):** engine resolution happens only inside
  `summarize_with_template` — resummarize passes config through and can never
  bypass per-meeting egress policy. Note for the in-flight privacy work:
  this adds a second caller of that choke point.
- Scope per M5: text-only, CLI + core. Desktop button and MCP tool are
  follow-ups (plan Phases 3–4).

[...]

Issue #523, opened 2026-07-23
Feature request: re-run the AI pass (summary + derived frontmatter) on an edited transcript


rymalia:
## Motivation

The desktop app's "Open in Editor" button makes it easy — and clearly intended — to edit a meeting file directly, and the transcript section is a legitimate edit target:

- correcting transcription errors that are significant enough to drift the summary
- correcting speaker attribution in transcript
- deleting off-topic sections 
- striking hot-mic moments or other sensitive comments from the record
- fixing a recording error that isn't worth editing the audio for

But after editing, nothing re-runs summarization. The `## Summary` section and the derived frontmatter (`action_items`, `decisions`, `intents`, `entities`, `people`) go stale against the corrected transcript, which largely defeats the purpose of the edit. Today the alternatives don't work: a full audio reprocess re-transcribes from scratch, so it doesn't safely refresh an edited file (it typically creates a duplicate output, and in rewrite/retry paths can replace the edits), and hand-rewriting the AI-owned sections is exactly the toil the pipeline exists to avoid.

It also completes the new `minutes import text` (#516) story: imported archives receive no LLM summarization pass and no generated summary-derived frontmatter (by design — the importer makes no LLM calls), so this command would give those files their first AI pass.

Conceptually: **"rerun the AI pass on the meeting transcript"** — audio untouched.

## Proposal

1. **Core:** a shared `minutes-core` function that re-runs the summarization stage on the file's existing transcript (plus `## Notes`, which the summarizer already weights) and splices only the AI-owned sections — never touching `## Transcript`, `## Notes`, `speaker_map`, or recording metadata.
2. **CLI:** `minutes resummarize <meeting>` following the `redo-speaker-mapping` CLI conventions — resolve by path or search term, preview by default (no write; the preview does invoke the model), `--apply`, `--engine`, `--template`, `--json`.
3. **Follow-up milestones (separate, once the core lands):** a "Re-Summarize" button in the desktop app (the jobs queue has a `Summarizing` progress state, but its jobs are audio-pipeline-specific today, so this needs a new text-only job kind), and an MCP `resummarize_meeting` tool.

## Why this fits the existing architecture

[...]

PR #573, opened 2026-07-27
feat(resummarize): refresh derived views after an applied AI pass

rymalia:
**Title:** feat(resummarize): refresh derived views after an applied AI pass

**Body:**

Follow-up to #526 — the graph/index invalidation item from the plan's Phase 4, which that PR deferred.

## The gap

A successful `resummarize --apply` rewrites summary-derived frontmatter — `entities`, `people`, `intents`, `action_items`, `decisions` — which invalidates every cached view computed from it. Nothing refreshed them, so `graph.db` went silently stale after every edit-and-resummarize cycle. `minutes people` and `minutes commitments` would answer from pre-edit data with no indication anything was out of date.

## Approach

Adds `minutes-core::derived` as a single owner of "this artifact changed, so its derived views are stale", rather than a 13th copy of `graph::rebuild_index`. This is the same doctrine you asked for on engine resolution in #523 (M1) applied to invalidation: one choke point, no parallel logic.

| View | Refresh | Automatic? |
|---|---|---|
| `graph.db` | full rebuild | yes |
| `search.db` | per-file `upsert_file` | yes |
| Vault copy (`strategy = "copy"`) | `sync_file` | yes |
| QMD collection | `qmd update -c` | yes, when configured |
| Knowledge base | `ingest_file` | **no** — `--ingest` |

**Knowledge ingestion is opt-in on purpose.** `update_from_meeting` deduplicates facts, but `append_log` appends a chronological entry unconditionally — so an automatic re-ingest would add a duplicate line to a user-owned append-only file on every resummarize. A disabled knowledge base now also reports a warning rather than the zero-fact "success" that `update_from_meeting` returns in that case, which was indistinguishable from a real run that found nothing.

**The search index is included despite syncing lazily on read.** `SyncMode::Auto` compares only mtime and size, and `SyncMode::Force`'s own doc comment says it exists to catch "mtime-collision edge cases" — so a regenerated summary of identical byte length, written within the filesystem's mtime granularity of the last sync, could stay stale. `upsert_file` reindexes unconditionally, closing that window for the one file we know changed without paying for a whole-corpus re-walk.

**Best-effort by construction.** `refresh_derived_views` has no error return: failures land in `RefreshReport::warnings` and the tracing log. A stale derived view is a nuisance; failing a write that already landed is not acceptable. The refresh runs from the public `resummarize_meeting` only after a real apply, so the crate-private `resummarize_meeting_with` test path stays free of side effects — and Phase 3's desktop button inherits the refresh for free.

## A hazard this surfaced

`graph::db_path()` and `SearchIndex::default_db_path()` are both global (`~/.minutes/*.db`) and independent of `config.output_dir`. A test that redirects only `output_dir` still hits the real database — my first draft's unit test cleared my own graph index and repopulated it from a `TempDir` fixture before I caught it.

So `refresh_derived_views_at` takes both paths explicitly, mirroring the existing `graph::rebuild_index_at` precedent, and this PR adds `SearchIndex::open_at` for the same reason. Two regression tests assert the injected paths are the targets.

This is a pre-existing repo-wide trap, not one this PR introduces — any test calling `rebuild_index` has always had it, and there is already evidence of it happening to `search.db` (I saw `output_dir changed; rebuilding index old=/var/folders/…/tmp…` on a fresh checkout). Happy to file that separately if you'd like it tracked.

## Findings I declined

Two items from review that I deliberately did not act on:

1. **`dirs::home_dir()` panics in `graph::db_path()`**, so a missing home directory would panic *after* the markdown write and make a completed apply look like a failure. Declined: it is pre-existing and shared by all ~13 existing `rebuild_index` callers, so fixing it here would be inconsistent and partial, and wrapping this one call in `catch_unwind` is worse than the disease. Better as its own change if you want it.

2. **The graph rebuild is whole-corpus on every single-file apply.** Accepted rather than optimized: it matches what the pipeline already does after every recording, and it is what makes the refresh inherently idempotent. Measured on an M3 Air at 92 meeting files: **93–142 ms**. If that becomes a problem at larger corpus sizes it argues for an incremental `rebuild_index`, which would benefit every caller, not just this one.

One known limitation worth recording: a *wrong* QMD collection name is undetectable here, because `qmd update -c <unknown>` exits 0. Only a spawn failure or non-zero exit is reported.

## Testing

- 9 new tests in `minutes-core::derived` (1168 → 1177 core unit tests), plus CLI flag coverage (80 → 81).
- `cargo clippy --all --no-default-features -- -D warnings` and `cargo fmt --all --check` clean; `check:llms` reports generated agent docs up to date.
- Verified end-to-end against a real meeting file (on a copy): apply printed `refreshed: graph (31 meetings), search`, the graph row count moved 30 → 31, and the row landed in `search.db` with the correct title. Confirmed the global databases are untouched by a full `cargo test` run.
- Rebased onto current `main` (`7037253b`) so the generated site constants regenerate against the released tree, per your note on #526.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1032,10 +1032,13 @@ minutes redo-speaker-mapping <meeting> --json                   # machine-readab
 minutes resummarize <meeting>                    # preview: show the regenerated summary + merge decisions
 minutes resummarize <meeting> --apply            # splice the regenerated content into the file
 minutes resummarize <meeting> --engine apple     # override the engine for this run
+minutes resummarize <meeting> --apply --ingest   # also re-ingest into the knowledge base
 minutes resummarize <meeting> --json             # machine-readable output
 ```
 
 `<meeting>` is a path or a search term (memos and `minutes import text` files work too — for imported archives this is their first AI pass). The preview **does invoke the model** (on cloud engines the transcript leaves the machine), it just doesn't write. Only the AI-owned sections (`## Summary`, `## Decisions`, `## Action Items`, `## Open Questions`, `## Commitments`) and derived frontmatter are replaced — `## Notes`, `## Transcript`, `speaker_map`, and capture/consent metadata are never touched. Action items and decisions keep user-curated state (`status: done`, due dates, decision authority) by exact-identity matching; anything ambiguous is surfaced in the preview instead of silently resolved. A failed run (engine `none`, provider error, empty output) never modifies the file, and the previous version is backed up to a hidden `.<name>.pre-resummarize.<unix-secs>.bak` sibling on every apply; the newest 3 backups per artifact are kept. One redaction caveat: editing the transcript never touches the retained WAV — for true audio redaction, delete or re-record the audio (`minutes cleanup` / retention settings).
+
+Because an apply rewrites summary-derived frontmatter, everything computed from it goes stale. A successful `--apply` therefore refreshes the derived views the same way the recording pipeline does: the relationship graph index is rebuilt, the vault copy re-synced (`strategy = "copy"` only), and the QMD collection reindexed if one is configured. Knowledge-base ingestion is **not** automatic — its chronological log is append-only, so re-ingesting the same meeting would add a duplicate entry every run; pass `--ingest` when you want it. All of this is best-effort: a refresh that fails warns and leaves the derived view stale, but never undoes a write that already succeeded.
 
 **Privacy**: Voice enrollment is self-only (no enrolling others). Level 3 confirmed profiles require explicit opt-in per person. Voice embeddings are stored locally in `~/.minutes/voices.db` with 0600 permissions. Nothing leaves your machine.
 
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -784,6 +784,10 @@ enum Commands {
         #[arg(long)]
         template: Option<String>,
 
+        /// With --apply, also re-ingest into the knowledge base (adds a new entry to its append-only log)
+        #[arg(long)]
+        ingest: bool,
+
         /// Emit machine-readable JSON instead of human output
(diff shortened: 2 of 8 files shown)


silverstein (approved):
Approved — this completes resummarize by closing the derived-view staleness the #526 plan deferred, and the engineering is excellent.

Verified on a rebase onto current main: 1172 core tests pass (including the 9 new derived tests), clippy -D warnings and fmt clean. The design matches your writeup exactly: refresh_derived_views is a single best-effort choke point (no error return, failures land in RefreshReport::warnings, never fails a write that already landed), refresh_derived_views_at injects paths to avoid the global ~/.minutes/*.db trap (mirroring rebuild_index_at), and it runs only after a real apply so the test path stays side-effect-free. The doctrine call is right — one owner of invalidation, not a 13th rebuild_index copy, exactly the #523 M1 pattern.

On your declined findings: both calls are correct. The dirs::home_dir() panic and the whole-corpus rebuild are pre-existing and shared by all rebuild_index callers; fixing them here would be partial and inconsistent. Please do file the global-db-path test trap as its own issue — it's a real repo-wide hazard worth tracking, and you've already got the evidence.

I merged current main into your branch and regenerated the site constants (the usual release.ts collision), no logic touched. Merging once CI is green. This lands in the next release alongside resummarize. Thank you — another exceptionally clean contribution.

silverstein:
Shipped in v0.24.0: https://github.com/silverstein/minutes/releases/tag/v0.24.0 — resummarize now refreshes graph/search/vault/QMD after an applied pass, so cross-meeting queries reflect the edit. Between this and #526, resummarize is yours end to end. The release Thanks section says so. Genuinely excellent work — thank you.

Issue #592, opened 2026-07-29
summarize: agent timeout kills only the direct child, so a surviving grandchild blocks the call indefinitely


rymalia:
## Summary

`summarize_with_agent_impl_timeout` enforces `summarization.agent_timeout_secs` by killing the spawned agent when the deadline passes. It kills only the **direct child**, not the child's process group. A grandchild that outlives it inherits the stdout pipe and holds it open, so the reader thread's `join()` blocks until that grandchild exits — long past the configured timeout.

The result is that `agent_timeout_secs` is **not a hard bound** on how long summarization can block.

Line references are against `7a61dd57` (current `main` at time of filing).

## Where

`crates/core/src/summarize.rs`, timeout branch of `summarize_with_agent_impl_timeout` (fn at `1682`, branch at `1820`):

```rust
if start.elapsed() > timeout {
    child.kill().ok();
    let _ = child.wait();
    let _ = stdout_handle.join();   // <-- blocks while a grandchild holds the pipe
    let _ = stderr_handle.join();
```

`stdout_handle` / `stderr_handle` are real reader threads spawned at `1760` / `1766`. Killing the direct child does not close the pipe, because the surviving grandchild still holds the write end.

**The same file already contains the correct pattern.** `run_speaker_mapping_via_agent` puts its agent in its own process group precisely so a timeout can kill the whole tree:

- `2948` — comment: *"Put the agent in its own process group so a timeout can kill the WHOLE tree"*
- `2954` — `command.process_group(0);`
- `3002` — `kill_process_group(child_pid);` **before** `child.kill()`
- `3022` — `kill_process_group` helper (with a non-Unix no-op stub at `3037`)

The summarize path does none of this.

### Secondary, milder site

`run_title_refinement_via_agent` (fn at `2600`, timeout branch at `2660`) also kills only the direct child. It does **not** hang — the thread it spawns is a stdin *writer* and the timeout branch never joins it, so it returns on schedule — but it still leaks orphaned grandchildren. Worth fixing in the same pass, though it is not the blocking bug.

## Reproduction

An "agent" that spawns a child surviving the shell being killed:

```bash
printf '#!/bin/sh\ncat >/dev/null 2>&1\nsleep 900\n' > /tmp/hang-engine.sh
chmod +x /tmp/hang-engine.sh
```

`~/.config/minutes/config.toml`:

```toml
[summarization]
engine = "agent"
agent_command = "/tmp/hang-engine.sh"
agent_timeout_secs = 30
[...]

PR #596, opened 2026-07-29
feat(desktop): add Resummarize button to meeting detail view

rymalia:
Adds a **Resummarize** button to the in-app meeting detail view, so a meeting's AI
pass can be re-run from the desktop app instead of dropping to the CLI.

Screenshot - New Resummarize button

This is Phase 3 of the resummarize plan (`docs/plans/rerun-ai-pass-resummarize-2026-07-21.md`).
Phase 1 (core + CLI) landed in #526, derived-view invalidation in #573, and the MCP
tool in #589. This PR touches none of those — it only calls the existing
`resummarize_meeting` core function. No change to `minutes-core`; the diff is
`tauri/src-tauri/src/commands.rs` and `tauri/src/index.html`.

Rebased onto `main` after #593 and #594, so the regenerated version/dependency
surfaces are current.

## Deviation from the plan — no new job kind

The plan's Phase 3 specifies a new `Resummarize` job kind in the processing queue.
**This PR deliberately takes the plan's own escape hatch instead:**

> "If this scope is too heavy for the first PR, defer Phase 3 entirely rather than
> wedging a text job into audio-shaped plumbing."

`ProcessingJob` requires `mode: CaptureMode`, `content_type: ContentType`, and
`audio_path: String`, and ~20 of its 24 fields describe a capture — a resummarize job
would have to fabricate all three. The worker loop calls `transcribe` unconditionally,
and `job_context`, `preserve_audio_alongside_output`, `remove_capture_artifacts`, and
`maybe_mark_context_session_*` all assume a capture exists.

Meanwhile `resummarize_meeting` is already self-sufficient: it owns its concurrent-edit
guard, the timestamped backup, the atomic rewrite, and the post-write
`refresh_derived_views` call. A queue entry would mostly duplicate lifecycle machinery
it doesn't need.

So this follows the app's existing precedent for long non-capture operations —
`tauri::async_runtime::spawn_blocking`, as used by `cmd_download_model`,
`cmd_warm_parakeet`, `cmd_confirm_speaker`, and the recall-chat commands.

**Accepted consequence:** a run doesn't survive an app restart and doesn't appear in the
jobs list. That's intentional — those surfaces model the audio pipeline. Happy to
revisit as a queue job if you'd rather have the durability.

## Concurrency

Desktop resummarize runs are serialized **globally**, not just per-path: every
successful apply rebuilds the shared graph DB and refreshes shared indexes, so two
concurrent runs on different files would contend there. An RAII guard holds the
in-flight slot and releases it on every exit path — success, error, and panic — with
mutex-poison recovery so a panic elsewhere can't permanently disable the button.

## Deliberately not included

- **No sensitivity gate.** The CLI's `resummarize` has none (its `restricted` gating is
  on the bulk read surfaces — search/actions/research), and the MCP tool's refusal
  exists because MCP hands preview content to an agent. This is a local, explicit,
  user-initiated action on a file the user already has open. Flagging it explicitly
  because a cloud engine plus a `restricted` artifact means transcript egress on one
  click — if you'd rather gate it, say so and I'll add it.
- **No preview mode** — the button applies; preview stays in the CLI and MCP.
- **No `--retitle`** (Phase 4, off by default — the filename slug derives from the title).
- **No knowledge-base ingest** — its log is append-only, so re-ingesting isn't
  idempotent; it stays opt-in.
- **No "Summarize" label for a meeting that has no summary yet.** The button always says
  "Resummarize" (for now). Deferred to its own PR; `cmd_get_meeting_detail` already
  returns `status` and per-heading `sections`, so it needs no backend work.

## What the button reports

On success the detail view names, one fact per line: what the run did to each AI-owned
section, the timestamped backup, any merge notes, and any derived-view refresh warnings.
It persists until the user leaves the meeting rather than auto-hiding — a run costs real
time and money, and merge notes name action items the user has to make a decision about.

**Getting that report to be *true* is most of the last commit.**
`sections_replaced` is the only outcome data core returns, and it is the list of
AI-owned sections that existed *before* the run. It cannot distinguish three different
outcomes, and describing a run by it alone got two of them wrong:

- **A first pass on a transcript-only meeting replaces nothing**, because there was
  nothing to replace. The banner said "Run finished in 32.7s but replaced no sections —
  the artifact is unchanged" while a summary and four derived sections had just been
  written. Simply false, and the misleading half is the half a user acts on.
- **A section the new pass did not produce is deleted, not replaced.**
  `splice_ai_sections` removes every pre-existing AI section and inserts only what the
  pass wrote, so a meeting that had `## Commitments` and a pass that found none ends up
  with no Commitments section — while core still lists it under `sections_replaced`.
  (Verified with a probe against core: `REPLACED=["Commitments","Summary"]`, resulting
  body contains no `## Commitments`.)

So the payload now also carries `sectionsWritten`, derived in the command layer from the
report's `new_ai_body`. The banner partitions the two lists into **replaced / created /
removed**, and says "written" rather than "rewritten" when there was no prior Summary. If
`sectionsWritten` is ever absent it falls back to the previous reporting rather than
inventing removals.

**That helper deliberately does no parsing of its own.** It reassembles the block
`splice_ai_sections` inserts and asks `markdown::find_unique_section` — core's own
finder — for each name. The first version hand-scanned lines for `## <name>` equality and
was wrong in both directions, because `h2_headings` skips headings inside fenced code
blocks and trims after `## `: a fenced ```` ## Commitments ```` in raw model prose counted
as written (so a deleted section was reported "replaced"), and `##   Commitments` was
missed (so a section on disk was reported "Removed"). Both come straight out of model
output, and both are covered by tests. Happy to move this into `minutes-core` beside
`splice_ai_sections` if you'd rather core owned the "what did this pass write" question —
it's only in the command layer to keep this PR off core entirely.

Merge notes render in the artifact's normal body text rather than the banner's accent
colour: they quote the user's own action items and decisions in full, and a dozen of
them in bold green buried the one line that says what happened.

The banner is also now selectable and wraps unbroken text. `body` sets
`user-select: none` page-wide and every content surface opts back in; this one had not,
which mattered little for a 3.2s toast but means a persistent report naming a backup
file cannot be copied out of.

## Bugs found and fixed while reviewing this PR

Click-testing the real dev app and repeated second-agent review turned up several
defects in my own drafts. Listing them because each is a behaviour worth knowing about:

1. **The second click of a double-click landed on `Delete`.** Arming the confirm
   inserts a hint banner into `.detail-card`, which is centred by
   `.detail-overlay { align-items: center }`. A card shorter than its max-height
   therefore *grows*, sliding the actions row **up ~24px** — directly under the cursor
   sat the red Delete button, which opened its archive dialog. Reproduced live, then
   fixed by pinning the card's height (`.detail-scroll` absorbs the banner). This is the
   vertical twin of a horizontal reflow bug fixed earlier in the same row.
2. **Rewrites were only constrained to `$HOME`, not the meetings directory.** The read
   endpoint uses `validate_meeting_path(path, &config.output_dir)`; this write path
   only checked home-containment, so a direct IPC call could rewrite any
   Minutes-shaped `.md` anywhere under `$HOME`. Now requires the configured meetings
   root. (Containment is checked inline rather than by reusing
   `notes::validate_meeting_path`, because that helper re-checks the extension
   case-sensitively and would have rejected a legitimate `.MD` artifact.)
3. **A completed run could hijack whatever view was open.** Both completion paths
   reopened the initiating meeting unconditionally. Now guarded on the detail overlay
   actually showing that path — deliberately *not* on `currentMeetingPath` alone,
   since Recall reuses that variable while the overlay is closed.
4. **Armed state leaked across views.** `closeDetailView()` and `showMdViewer()` hide
   the overlay without resetting the button, so the `Confirm?` label and its pinned
   width/height could persist. `openMeetingDetail()` now also resets synchronously on
   entry, before its awaits.
5. **A button disabled by someone else's run never re-enabled.** Only the view that
   started a run learned when it ended; any other view stayed disabled for the rest of
   its life, and its explanatory message had already auto-hidden — so it just looked
   broken. Added a 2s status re-check while a run is in flight, plus a request-sequence
   guard so an out-of-order status reply can't strand the button.
6. **Making the report persist reopened the layout hazard and leaked across meetings.**
   A banner that no longer auto-hides is a banner that is still there when you navigate.
   The in-card "Source"/"Open" buttons reach `openMeetingDetail` without closing the
   overlay, so meeting A's backup filename and merge notes could end up stapled to
   meeting B. Fixing that surfaced a chain: the card hold could be released one IPC
   round-trip *after* the report was drawn by an unawaited `refreshResummarizeButton()`;
   a superseded open still returned normally, so the handler could report onto a view it
   no longer owned; and *any other feature's* banner could grow the card first, making a
   later hold capture an already-grown height. The hold is now owned solely by the
   banner lifecycle — one take in `showDetailFeedback`, one release in
   `clearDetailFeedback` — which is the only shape that is correct for all of them.
7. **A failed refresh claimed success over visibly stale content.** If the post-run
   reload of the detail view fails, the write still happened but the screen still shows
   the old summary. It now says so instead of reporting an unqualified success.

## Disclosed, not fixed — with reasoning

- **TOCTOU on the validated path.** The path is canonicalized, then reopened by name
  inside the blocking task. Swapping a parent for a symlink in between could redirect
  the write. Real, but it needs an attacker who can already write inside your home
  directory, and it's the same validate-then-reopen pattern used throughout the app;
  fixing it properly means open-by-handle plumbing in `minutes-core`.
- **`applied: false` is unreachable but unhandled.** The UI unconditionally reports a
  write. Today `resummarize_meeting` either sets `applied = true` or returns `Err`, so
  the message can't lie — but it would if the core ever changed.
- **Cross-platform untested.** Verified on macOS 26 / Apple Silicon only.

### Previously disclosed here, now fixed upstream

An earlier revision of this description carried a long disclosure about agent
summarization having no reliable completion bound — a surviving grandchild holds the
stdout/stderr pipes, so the reader-thread join blocks on **both** the timeout path and
the clean-exit path (measured 12.014s against a 2s timeout, and 12.016s with a 300s
timeout never consulted). Its consequence *here* was that the global guard stayed held
and every meeting's button stayed disabled app-wide, **including when summarization
succeeded**.

That was filed as #592 and **fixed in #593**, which this branch now sits on top of. No
desktop-side change was needed, and a desktop-side timeout would have been the wrong
fix — the guard has to outlive anything that could still write.

## Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy --all --no-default-features -- -D warnings` — clean
- `cargo test -p minutes-app` — **274 passed / 0 failed** (`--test-threads=1`), 8 new
- Frontend assertions **31/31**; both inline script blocks parse
- Report copy rendered for **10 outcome shapes** by extracting the shipped functions out
  of `index.html` and eval'ing them, so the sample output cannot drift from the app's

Every frontend assertion has been deliberately broken and confirmed to go red; that
falsification pass caught three of my own guards being vacuous (a substring match where
`pinDetailCardHeight` also matched `unpinDetailCardHeight`, a refutation defeated by the
word "innerHTML" appearing in the very comment explaining why it isn't used, and a bare
`grep 'Created '` that matched unrelated text elsewhere in the file).

### A pre-existing global-`HOME` race in the app suite — not from this PR

Under parallel execution `list_documents_merges_assistant_and_meeting_sources_by_recency`
fails intermittently (reproduced 3/3). `with_temp_home` swaps the **process-global**
`HOME`, and tests that read `dirs::home_dir()` without taking the shared guard race
against it. Everything passes serially (274/0). Confirmed pre-existing by stashing this
PR's changes and re-running: the same single test fails identically without them. #594
fixed this class in `minutes-core`; the app crate still has its own copy of the pattern.

### Click-tested end-to-end in `~/Applications/Minutes Dev.app`

- **Global serialization is real and user-visible.** With a run in flight on meeting A,
  opening meeting B shows B's button disabled with "Another meeting is being
  resummarized — this button unlocks when it finishes."
- **Failure writes nothing.** After a failed run the artifact was byte-identical and
  **no `.bak` was created** — the guard aborts before any write.
- **The double-click/Delete collision**, reproduced and then verified fixed by clicking
  identical coordinates twice and getting a run instead of the archive dialog.
- **A view disabled by another meeting's run re-enables on its own** once that run ends.
- **The actions row does not move.** With a six-line report replacing the one-line
  pending banner, `Delete` stayed at the same y in all captures while the content below
  scrolled — i.e. `.detail-scroll` absorbed it.
- **Navigating away via a commitment's "Source" button clears the banner** — the exact
  cross-meeting leak path above.
- **On-disk merge contract holds:** a hand-written `## Notes` section survived, and
  pre-existing action items kept `status: done` and their due dates.
- **First-ever pass on a transcript-only meeting** (24m51s, 300+ transcript lines,
  `status: transcript-only`, no AI sections). Reported
  `Summary written in 30.7s. Created Decisions, Action Items, Open Questions, Commitments.`
  and the artifact went to `status: complete` with all five sections present. This is the
  case that previously read "the artifact is unchanged".
- **A `carried-with-conflict` merge note**, produced by seeding a disagreeing assignee on
  an action item whose text the pass reproduces:
  `- Action item "Give feedback on the contract today." — kept your version: assignee kept as 'Ryan' (regenerated pass said 'Ryan (contractor)')`
- **The actions row never moves.** `Delete` held the identical y across a **2-line**, a
  **9-line**, and a **13-line** report on a tall card — the condition where the card is
  already at its viewport cap and the banner has the most room to shove things around.

**Still not exercised end-to-end:** the `ConcurrentEdit` branch specifically. I verified
that a *failed* run writes nothing, but I did not win the race to modify the file
mid-run, so the compare-before-swap abort path and its "save your editor and re-run"
message remain covered only by the core's own tests.


rymalia:
Edit: added screenshot.

silverstein:
Merged. That completes all four phases of the resummarize plan you scoped in #523: core plus CLI in #526, derived-view invalidation in #573, the MCP tool in #589, and the desktop button here.

On the deviation: taking the plan's own escape hatch instead of adding a Resummarize job kind was the right call, and your reasoning for it is the kind I want on record. ProcessingJob requires mode, content_type, and audio_path, roughly 20 of its 24 fields describe a capture, and the worker loop calls transcribe unconditionally. Fabricating a fake capture to reuse that plumbing would have been worse than not reusing it, and spawn_blocking already has precedent in cmd_download_model and cmd_confirm_speaker.

Verification notes for the record, since this is the first UI-touching contribution: I built and launched the branch on a real Mac and it starts cleanly alongside a running production instance. The button sits correctly inside detail-actions between Discuss with Agent and Open in Editor, uses btn btn-secondary btn-sm which is the most common existing combination in the file, and introduces no new colors, fonts, or radius values, so DESIGN.md is satisfied. The async handling stood out: monotonic request IDs with staleness guards so an unawaited refresh cannot release state underneath a newer one, plus the slot claim naming the currently-running artifact. The comment about the resting pointer landing on the red Delete button is the sort of detail that only comes from actually using the thing.

I could not complete a pixel-level capture because screencapture over SSH is blocked by Screen Recording TCC and this is a menu bar app. Your screenshot covered that gap. We are building a purpose-scoped helper so future UI contributions get real automated verification rather than depending on the contributor's own screenshot.

Four features, three bug diagnoses, and two flaky-test fixes in a week. Thank you.

PR #598, opened 2026-07-29
test(desktop): take the HOME guard in list_documents test (#591)

### LINKED ISSUE #591: Flaky test: list_documents_merges_assistant_and_meeting_sources_by_recency fails in full-suite runs
`commands::tests::list_documents_merges_assistant_and_meeting_sources_by_recency` (tauri/src-tauri) fails when the full `minutes-app` suite runs, and passes when run in isolation.

Confirmed on macOS (Apple Silicon), with controls to rule out a dependency change:

| Branch | How run | Result |
|---|---|---|
| origin/main | full suite | FAIL |
| origin/main | isolation | pass |
| desktop dep batch (#590) | full suite | FAIL |
| desktop dep batch (#590) | isolation | pass |

So it is pre-existing on main and independent of the dependency bumps that surfaced it.

The pass-alone / fail-together pattern points at shared mutable state or an ordering assumption rather than a logic bug: the test merges assistant and meeting sources and asserts recency ordering, so a likely cause is another test writing into the same directory or timestamps colliding within filesystem granularity when tests run concurrently.

This is the third flake of this shape, after #529 (capture relay reconnect) and #566 (bump-version readiness marker). Both of those were reddening unrelated contributor PRs before they were fixed, which is the real cost here.

Wanted: make the test independent of sibling tests (own temp directory, explicit timestamps rather than wall-clock ordering) rather than serializing the suite.

silverstein:
Fixes #591. The framing in that issue was wrong and this PR corrects it.

## Not a flake

`list_documents_merges_assistant_and_meeting_sources_by_recency` fails **100% of the time** on macOS main, not intermittently. Measured on an Apple Silicon Mac:

| Build | Full-suite runs | Result |
|---|---|---|
| origin/main | 5 | 5 failed |
| main before #596 | 3 | 3 failed |
| this PR | 5 | **0 failed** (274 passed) |

It passes on Linux, which is why it looked intermittent from the VM.

## Cause

The test resolves `dirs::home_dir()` and builds its fixture inside the real home directory, but never takes the `test_guard()` mutex that the HOME-mutating helpers hold. `with_temp_home()` repoints HOME to a temp dir and recursively deletes it afterwards, so when the two overlap, that cleanup removes the directory this test is still writing into. The symptom is a `NotFound` partway through setup, at the `CLAUDE.md` write, rather than the recency assertion the issue guessed at.

Fix: take the same guard. One line plus a comment explaining why.

## The larger finding

**CI has never run this test suite.** Every workflow runs `cargo check -p minutes-app --tests`, which compiles them; nothing runs `cargo test -p minutes-app`. All 274 Tauri desktop tests compile and never execute, on any platform. That is why a deterministic macOS failure sat unnoticed, and it means the desktop app has had no test enforcement while CLAUDE.md counts these tests as coverage.

Adding `cargo test -p minutes-app` to CI is deliberately **not** in this PR. Turning on a suite that has never run is likely to surface more than this one failure, and it deserves its own change rather than being smuggled in behind a one-line fix. Filed as a follow-up.

## Note on my first attempt

The initial commit on this branch replaced the test's 5ms sleeps with explicit mtimes, on the theory that timestamp granularity was collapsing the recency ordering under load. That theory was wrong: the failing assertion is a `NotFound` during setup, not an ordering check. That change is discarded rather than kept as an unrelated improvement.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -14531,19 +14531,16 @@ mod tests {
         assert_eq!(text_file_kind(Path::new("/tmp/test.txt")), Some("text"));
     }
 
-    /// Pin a file's mtime to `secs_ago` before now, so recency assertions do
-    /// not depend on sleep timing or filesystem timestamp granularity (#591).
-    fn set_test_mtime(path: &std::path::Path, secs_ago: u64) {
-        let when = SystemTime::now() - Duration::from_secs(secs_ago);
-        let file = std::fs::File::options()
-            .write(true)
-            .open(path)
-            .expect("open for mtime");
-        file.set_modified(when).expect("set mtime");
-    }
-
     #[test]
     fn list_documents_merges_assistant_and_meeting_sources_by_recency() {
+        // This test resolves HOME and builds its fixture inside it, so it must
+        // hold the same guard the HOME-mutating helpers take. Without it,
+        // with_temp_home() can repoint HOME and then recursively delete that
+        // directory while this test is still writing into it, which showed up
+        // as a NotFound partway through setup (#591). Deterministic on macOS,
+        // invisible on Linux, and never caught because CI does not run this
+        // suite at all.
+        let _guard = test_guard();
         let home = dirs::home_dir().expect("home dir");
         let temp = tempfile::Builder::new()
             .prefix("minutes-documents-test-")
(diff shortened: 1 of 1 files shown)


silverstein:
Follow-up for turning the suite on is filed as #599.

PR #600, opened 2026-07-29
ci: actually run the desktop test suite (#599)

### LINKED ISSUE #599: CI never runs the desktop test suite (cargo test -p minutes-app)
Every workflow runs `cargo check -p minutes-app --tests`, which compiles the desktop tests. Nothing runs `cargo test -p minutes-app`. All 274 Tauri tests compile and never execute, on any platform.

Confirmed by grepping every `cargo test` invocation across `.github/workflows/`:

```
cargo test -p minutes-core --no-default-features --lib
cargo test -p minutes-core --no-default-features --lib copilot::
cargo test -p minutes-cli  --no-default-features -- copilot
cargo test -p minutes-core --no-default-features --features diarize --lib -- diarize
cargo test -p minutes-core --no-default-features --test integration      (macOS, continue-on-error)
cargo test -p minutes-core --no-default-features --test live_sidekick_eval
```

No `minutes-app` entry.

## Why it matters

This is how #591 hid. That test failed **100% of the time on macOS** for an unknown period, and nobody saw it because the suite never runs. CLAUDE.md counts the Tauri tests as part of the project's coverage, so the documented safety net does not exist for the desktop app, which is also the surface CI can least otherwise validate.

## Wanted

Add a `cargo test -p minutes-app --no-default-features` step. Two cautions:

1. **Expect more failures than #591.** A suite that has never been enforced has probably accumulated others. Worth running locally on macOS and Linux first and triaging the full list before wiring it into a required check, otherwise this lands as a red gate.
2. **Prefer `--test-threads=1` or fix the shared state.** Several of these tests mutate process-global state (HOME, USERPROFILE) behind a local mutex, and #591 was caused by a test that read HOME without taking that guard. The same class of bug in minutes-core was #588. An audit of which desktop tests touch global state would be the durable version of this.

silverstein:
Closes #599.

## The gap

Workflows ran `cargo check -p minutes-app --tests`, which **compiles** the desktop tests, and nothing ever ran them. All 274 Tauri tests compiled and never executed, on any platform, while CLAUDE.md counts them as part of the project's coverage.

That is exactly how #591 hid: a test failing 100% of the time on macOS sat unnoticed because the suite never ran.

## Change

One step, running single-threaded.

Single-threaded is deliberate rather than incidental. Several of these tests mutate process-global state (HOME, USERPROFILE) behind a local mutex, and #591 was caused by a test that read HOME without taking that guard. Serializing is safer than trusting every current and future test to remember, and the suite runs in well under a minute either way.

## Measured before proposing

Green on both platforms I can test directly, with #598 in:

| Platform | Result |
|---|---|
| Linux (`--test-threads=1`, as CI runs it) | 262 passed, 0 failed |
| macOS, 5 consecutive full-suite runs | 274 passed, 0 failed each |

I had originally warned in #599 to expect a backlog of failures from switching on an unenforced suite. The data says otherwise, so this is a small change rather than a triage project.

## The one unknown

**Windows.** The test matrix covers macOS, Windows, and Linux, and I can only run the first and third. These tests have never executed on Windows, so this PR's own CI run is the first time they will. If Windows surfaces failures, the honest options are to fix them here or scope the step to the two verified platforms and file the Windows gap separately. I would rather find out on a PR than keep the suite dark.


silverstein:
Scoped to macOS. Enabling the suite surfaced 5 pre-existing Windows failures (Unix-only path and HOME assumptions), now tracked in #602. Guarding macOS today is strictly better than guarding nothing, which is what this suite had before, and Windows extension is gated on that issue rather than left as a red check here.

PR #606, opened 2026-07-30
test(resummarize): close edge coverage gaps

rymalia:
## What

Three test-only commits against `crates/core/src/resummarize.rs`. No production code
changes.

1. **Concurrent-edit byte preservation** — strengthens the existing concurrent-edit
   test to assert the user's bytes survive verbatim when an apply aborts, rather than
   only asserting the error variant. Renamed to
   `concurrent_edit_aborts_apply_without_overwriting_user_bytes`.
2. **Large-artifact round trip** — new
   `apply_round_trips_generated_artifact_larger_than_one_megabyte`, covering an
   artifact over 1MB through a full apply and asserting the transcript section
   survives intact.
3. **Tightened assertions** — byte-length equality between original and replacement,
   and section-text preservation for notes.

## Why

`ConcurrentEdit` is the guard standing between a resummarize pass and a user's
in-flight edits, and #596 noted it was not exercised end-to-end. These close that
gap and pin the "abort must not overwrite" property, so a future change to the
guarded-rewrite path cannot silently weaken it.

## Verification

`cargo test -p minutes-core --no-default-features` → **1182 passed, 0 failed,
1 ignored** (the ignore is pre-existing and unrelated).

Includes a `chore(site)` commit bumping the embedded test count in
`site/lib/release.ts` from 1675 to 1676, since the new test moves it. Generated with
`node scripts/sync_site_release_version.mjs` — not hand-edited.

The agent-docs generator's outputs (`llms.txt`, `llms-full.txt`, and the error and
MCP tool catalogs) are deliberately not included. Re-running `generate_llms_txt.mjs`
on this branch produces a date-stamp diff and nothing else — verified line by line
across all six files it writes — and `generate_llms_txt.mjs --check` normalizes that
stamp away by design (`scripts/generate_llms_txt.mjs:753-764`). Both generated-docs
CI jobs pass on this branch as committed.

### DIFF
--- a/crates/core/src/resummarize.rs
+++ b/crates/core/src/resummarize.rs
@@ -1470,10 +1470,20 @@ mod tests {
     }
 
     #[test]
-    fn concurrent_edit_aborts_apply() {
+    fn concurrent_edit_aborts_apply_without_overwriting_user_bytes() {
         let dir = TempDir::new().unwrap();
         let path = write_meeting(&dir);
         let path_clone = path.clone();
+        let user_edit = fs::read(&path)
+            .unwrap()
+            .into_iter()
+            .chain(
+                b"\n<!-- saved mid-inference; preserve every byte -->\n"
+                    .iter()
+                    .copied(),
+            )
+            .collect::<Vec<_>>();
+        let expected_user_edit = user_edit.clone();
         let err = resummarize_meeting_with(
             &path,
             &test_config(),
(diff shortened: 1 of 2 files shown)


silverstein (approved):
Approved. Verified locally: 35 resummarize tests pass, including both new ones (`concurrent_edit_aborts_apply_without_overwriting_user_bytes` and `apply_round_trips_generated_artifact_larger_than_one_megabyte`).

Strengthening the concurrent-edit test from asserting the error variant to asserting the user's bytes survive verbatim is the right instinct. `ConcurrentEdit` is the guard between a resummarize pass and someone's in-flight edits, and a test that only checks which error came back would still pass if the file were clobbered on the way to returning it. That is the difference between testing the message and testing the property.

PR #607, opened 2026-07-30
feat(qa): narrow UI screenshot helper for automated verification (#577)

silverstein:
First phase of the agent-driven QA harness from #577: a capture primitive narrow enough to hold a Screen Recording grant safely.

## Why

Desktop UI changes are the one class CI cannot verify, so each one routes through the maintainer clicking around. When that gets skipped, regressions ship. #605 fixed a title truncation caused by #596 adding a fifth action button, and no automated check could have caught it: I merged #596 on static review plus the contributor's screenshot precisely because I could not screenshot the running app.

The tempting shortcut is granting Screen Recording to `sshd`. That is a bad trade: identical capability, enormous blast radius, handing continuous screen access to anything that ever runs over SSH on a machine with client financial data and password manager windows on screen.

## The design

TCC grants attach to an executable, so the fix is a binary narrow enough that the grant is defensible.

**The boundary is the target, not the trigger.** `ui_shot` only captures windows owned by an allowlisted Minutes bundle, and the allowlist is a compile-time literal rather than an argument or prefix match, so it cannot be widened by the caller. Even fully compromised, this helper cannot photograph a password manager or a banking session.

It matches on owning process id rather than window title, so it cannot be tricked into capturing an unrelated window that shares a name, and it sorts candidates largest-first so index 0 is the main window rather than a tooltip.

## Verified on macOS

| Case | Result |
|---|---|
| `com.apple.Safari` | refused, exit 3, **no file written** |
| `com.1password.1password` | refused, exit 3 |
| no arguments | usage, exit 1 |
| allowlisted bundle not running | exit 2 with explanation |
| allowlisted bundle, no grant yet | exit 4 with the exact Settings path |

A zero-pixel capture is treated as failure rather than written out, because that is what macOS returns when the grant is missing but the call itself succeeds. Writing that file would look like success.

## Remaining setup (needs a GUI session)

Signing with a stable identity and granting Screen Recording both require the machine itself: the login keychain and TCC prompts are not reachable over SSH. Documented in `docs/development/ui-screenshot-helper.md`, including why the signing step matters (ad-hoc signing means TCC keys on content hash, so every rebuild invalidates the grant, the same failure mode as a rebuilt dev app losing Accessibility).

## Scope

This gives the capture primitive only. A screenshot proves a window rendered; the accessibility-tree fingerprints from #577 are what prove it rendered the *same as last time*. That is the next phase.

### DIFF
--- a/docs/development/ui-screenshot-helper.md
+++ b/docs/development/ui-screenshot-helper.md
@@ -54,39 +54,41 @@ persistence, and it writes exactly the output path it is given.
 One-time, and it needs a GUI session because TCC prompts and the keychain are
 not reachable over SSH.
 
-1. Build and install the helper. `cargo build -p minutes-app` compiles it to
-   `tauri/src-tauri/bin/ui_shot` as part of the normal build; copy it somewhere
-   stable:
+Run this in a Terminal **on the machine**, not over SSH:
 
-   ```bash
-   mkdir -p ~/.minutes/bin
-   cp tauri/src-tauri/bin/ui_shot ~/.minutes/bin/ui_shot
-   chmod 755 ~/.minutes/bin/ui_shot
-   ```
+```bash
+./scripts/install-ui-shot.sh
+```
 
-2. **Sign it with a stable identity.** This step is what makes the grant
-   survive rebuilds, and it must run in a Terminal on the machine (the login
-   keychain is not available over SSH):
+It builds, installs to `~/.minutes/bin/ui_shot`, signs with the first stable
+identity it finds (override with `MINUTES_DEV_SIGNING_IDENTITY`), and then tells
+you whether the Screen Recording grant is already held.
 
-   ```bash
-   codesign --force --options runtime \
-     --sign "Developer ID Application: Your Name (TEAMID)" \
-     ~/.minutes/bin/ui_shot
-   ```
+Two constraints make this a script rather than a copy-paste:
 
-   Without it the binary is ad-hoc signed, which means TCC identifies it by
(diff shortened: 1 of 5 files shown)


silverstein:
Closing unmerged. Mat has ChatGPT's computer use (CuaDriver) already installed and granted on the Mac, and it is sufficient for the ad-hoc "does this UI look right" checks this was built for. Merging this would add a Swift source, a build.rs step compiling it on every macOS desktop build, an install script, a LaunchAgent, and docs, for tooling with no user. Dead code with a build cost is worse than no code.

The branch `feat/ui-screenshot-helper` stays if we ever want unattended, repeatable UI verification, which is the one thing an interactive agent cannot provide: a fresh judgment each run is not a regression test. #577 keeps the design.

Three platform constraints are worth recording, since each cost a cycle to discover and would cost them again:

1. **CGWindowListCreateImage is dead for window capture on macOS 14+.** It returns nil even with Screen Recording granted, which reads as a permission failure. ScreenCaptureKit (SCScreenshotManager) is the supported path.

2. **macOS attributes screen capture to the responsible process.** A capture invoked over SSH is attributed to sshd and denied even when the helper binary itself holds the grant. `launchctl asuser` fixes the session context but needs root. The workable answer is a LaunchAgent, which runs in the GUI session, ideally on-demand via WatchPaths rather than resident.

3. **codesign cannot reach the login keychain over SSH** (`errSecInternalComponent`, `User interaction is not allowed`), including from the user's own SSH session. Signing must happen in a Terminal on the machine. That is also a useful safety property: no remote agent can sign a binary and grant itself capture access.

PR #612, opened 2026-07-31
Archive: private local legal evidence pilot

silverstein:
## Outcome

Adds a separate `Minutes Archive Dev` macOS app for an attorney’s long-lived document archive. The workflow starts with a multi-location metadata-only census, then requires a distinct content-access action before it builds a temporary local search vault.

## Security and evidence boundary

- folder-picker-authorized roots only; broad home and filesystem roots are refused
- census emits aggregate format and coverage signals without names, paths, hashes, or content
- searchable PDF, DOCX, TXT, TEXT, and Markdown only; scans and unsupported formats are reported honestly
- exact provision and document-level retrieval with visible constraints and exact source excerpts
- final root, membership, no-link, file identity, byte, and SHA-256 revision fences before every result
- stale, replaced, moved, or mutated evidence is withdrawn
- PDF/DOCX conversion and Apple semantic embedding run in separate resource-limited, network-denied workers
- semantic results are separately labeled suggestions, never deterministic legal matches
- source content, SQLite/FTS rows, and semantic vectors remain in memory and are not persisted
- closing the only window terminates the process and discards the in-memory legal session
- no QMD runtime, downloaded model, cloud AI, generated legal answer, shell, opener, or webview network permission

## Verification

`./scripts/verify-archive-dev-app.sh` passed against the installed app:

- 47 focused Rust tests, 0 failures
- synthetic legal benchmark
- strict focused Clippy
- both real worker sandbox/self-tests
- installed-executable TXT/DOCX/PDF retrieval and mutation withdrawal
- deterministic UI interaction smoke
- installed native lifecycle smoke: visible window, close event received, process purged
- valid installed ad-hoc bundle seal
- zero open network sockets observed on fresh launch
- macOS Archive dependency tree uses patched `quick-xml 0.41.0`; vulnerable `0.37.5` is absent

Release-candidate evidence and exact gates: `docs/release/archive-pilot-rc-2026-07-30.md`.

## Why this remains draft

This is a strong local development candidate, not an attorney handoff. Before distribution it still requires:

- review and merge of the fixed `signed-archive-acceptance` workflow, followed by an exact protected candidate tag and explicit environment approval; that workflow then performs Developer ID signing, notarization, stapling, Gatekeeper assessment, and provenance packaging without executing candidate code after credentials unlock
- native human click-testing of the installed Tauri app and folder picker
- full installed-app testing with networking disabled
- independent security review
[...]

PR #615, opened 2026-07-31
CI: gated signing and notarization for Archive pilot

silverstein:
## Outcome

Adds the fixed, reviewer-gated signing service needed to produce a Developer ID signed and Apple-notarized `Minutes Archive.app` from an exact Archive candidate without merging that candidate first.

## Trust boundary

- manual dispatch only, from `main`, by repository owner `silverstein`
- requires a protected `acceptance-<full-sha>` tag resolving exactly to the requested candidate
- candidate checkout, tests, production bundle build, document-worker smoke, and native close-to-purge smoke happen in a no-secret job
- the unsigned app is hashed and transferred as an inert artifact
- the secret-bearing job is behind the existing `signed-dev-acceptance` reviewer environment
- it never checks out candidate source or executes candidate code after credentials unlock
- requires a `Developer ID Application` identity and exact Team ID `63TMLKT8HN`
- signs inside-out, submits to Apple notarization, staples, validates, runs Gatekeeper assessment, and emits a seven-day provenance-bound zip
- no public release, tag creation, deployment, or updater publication

## Policy enforcement

A dedicated checker golden-hashes the complete pre-secret boundary and secret-bearing job. Dynamic negative tests prove CI rejects extra signing steps, candidate execution after unlock, early secret access, extra triggers, removed owner binding, or removed notarization. All action references are required to be full-SHA pinned.

## Verification

- `actionlint`
- `node scripts/check_signed_archive_acceptance_policy.mjs`
- `node scripts/check_signed_archive_acceptance_policy.test.mjs`
- `git diff --check`

This PR intentionally contains only the fixed signing infrastructure. After review and merge, it can notarize the exact still-unmerged Archive candidate from PR #612.

### DIFF
--- a/.github/workflows/lint-workflows.yml
+++ b/.github/workflows/lint-workflows.yml
@@ -32,6 +32,8 @@ jobs:
         run: |
           node scripts/check_signed_acceptance_policy.mjs
           node scripts/check_signed_acceptance_policy.test.mjs
+          node scripts/check_signed_archive_acceptance_policy.mjs
+          node scripts/check_signed_archive_acceptance_policy.test.mjs
 
       - name: Check graph worker packaging parity
         run: |
--- a/.github/workflows/signed-archive-acceptance.yml
+++ b/.github/workflows/signed-archive-acceptance.yml
@@ -1,0 +1,316 @@
+name: Signed Archive Pilot Acceptance
+
+on:
+  workflow_dispatch:
+    inputs:
+      candidate_sha:
+        description: Full SHA protected by the matching acceptance-<sha> tag
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: signed-archive-acceptance
+  cancel-in-progress: false
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_BUILD_JOBS: "2"
+  TAURI_CLI_VERSION: "2.10.1"
+
+jobs:
+  authorize-candidate:
+    name: Authorize exact protected Archive candidate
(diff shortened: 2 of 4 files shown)


silverstein:
Signing-gate follow-up at `3f50c379193258aa9b820a6ac644aea8f8cfbf9e`:

- the secretless unsigned-build job now runs the candidate-owned client-free human-QA fixture test before building the production Archive app;
- the reviewed pre-signing golden hash was updated, and the policy checker now requires the fixture-validation step;
- mutation tests prove removal of that step is rejected alongside extra triggers, pre-signing secret exposure, candidate execution after credential unlock, and weakened notarization;
- local policy and mutation tests, actionlint, and PR CI are green.

The secret-bearing signing job and its golden hash are unchanged. This PR remains unmerged; merging it is still the explicit owner-controlled boundary before any acceptance tag or protected workflow dispatch.

silverstein:
Closing in favour of a different approach.

This PR introduced a second, parallel signing pipeline for the Archive pilot, with its own authorization ritual and its own policy guard. Adversarial review of that design surfaced enough structural issues — in the guard's coverage model and in the bundle-integrity checks — that hardening it further would mean maintaining two independent implementations of the same security-critical path.

`signed-dev-acceptance.yml` already exists, already ships notarized artifacts, and now has hardened tag resolution (#616). The Archive pilot will reuse those primitives via a private, Archive-only artifact path rather than a separate pipeline.

The app itself (#612) is unaffected — none of this touches the local-only retrieval boundary or its verification.

Findings from the review are tracked separately and will be addressed against the existing workflow.

Issue #617, opened 2026-07-31
Flaky: corpus-lease process budget exceeded on windows-latest in CI


silverstein:
## Symptom

`Agent Corpus + MCP Lockfile (windows-latest)` fails intermittently with:

```
Error: meeting corpus retained snapshots exceeded their process budget
  ❯ reserveCorpusMemory src/corpus-lease.ts:253:11
  ❯ withStableCorpusLease src/corpus-lease.ts:1747:23
Error: Test timed out in 15000ms / 30000ms
```

## Why it looks like resource contention rather than a code defect

Across two runs of the *same commit*, a **different suite** failed each time — `SDK corpus policy tests` on one, `MCP corpus policy tests` on the next. A deterministic defect would fail the same test. The budget is process-global, so whichever test runs while the runner is under memory pressure is the one that trips.

Assertions also surface as the wrong-error shape, e.g.:

```
expected [Function] to throw error including 'stable meeting corpus authorization f…'
but got 'meeting corpus retained snapshots exceeded their process budget'
```

— i.e. the budget error preempts the error the test was actually asserting on.

## Frequency

Roughly 1 in 12 recent CI runs, across multiple unrelated branches, and it has also failed on `main` (`70ebfbcf`). Windows only; `ubuntu-latest` and `macos-latest` are unaffected.

## Impact

`CI Gate` is a required status check, so a flake here blocks merges on PRs whose diffs cannot touch this code path.

## Possible directions

- Raise or make configurable the retained-snapshot process budget on Windows runners
- Reset/isolate the process-global budget between suites so one leak does not preempt unrelated assertions
- Raise the 15s/30s test timeouts for the corpus suites on Windows

🤖 Generated with [Claude Code](https://claude.com/claude-code)

silverstein:
Picked this up in the Archive/legal lane. One contributing cause identified and fixed in #623; the issue should stay open for the rest.

## Cause found: an unmirrored test in the SDK copy

`corpus-lease.ts`, `corpus-lease.test.ts`, `corpus-lease-worker.ts` and `secure-read.ts` exist once per package and are byte-identical mirrors. Checked against history — every commit that touched either copy kept them the same, until `640ad42e` landed a determinism fix in the MCP copy alone and `244cc558` followed it. Nothing compares the copies, so the drift was silent.

[...]

PR #621, opened 2026-08-02
feat(i18n): add switchable Simplified Chinese localization

liliangouw:
Follow-up to the Discussion: https://github.com/silverstein/minutes/discussions/420

Adds a switchable Simplified Chinese (zh-CN) localization for the desktop app.

Design principles:

- English stays canonical — it's the default and the translation key. Any missing translation falls back to English cleanly.
- Additive runtime — no CI gate on translations; a partial catalog never blocks a build or release.
- Clean separation — the runtime (scripts/i18n.js) and the catalog (locales/zh-CN.js) are separate, making it easy to add other languages later (e.g. pt-BR from #124).
- Opt-in — defaults to `auto` (system locale) or `en`; users switch via Settings → Language.

What's included:

- Frontend DOM translator (walks text/placeholder/title/aria-label, plus MutationObserver for dynamic content)
- Scoped CJK font fallback (only when locale is zh-CN; Latin design untouched in English)
- Rust-side `tr()` for tray menu, notifications, window titles
- Settings language dropdown + config `[ui] language`

Verification:

- Core i18n + config unit tests pass
- Frontend runtime exercised in a real browser (static text, attributes, interpolation patterns, MutationObserver-driven dynamic text, user content left untouched, lossless switch-back to English, scoped CJK font applied only in zh-CN)

Notes:

Translations are a first pass (partly machine-assisted) and isolated in the catalog — terminology can be tweaked in `locales/zh-CN.js` without touching code.


silverstein:
Drive-by while this is still a draft, since all three red checks are mechanical rather than anything wrong with your work.

**Format check (macOS only)** is the real failure. `cargo fmt --all` should clear it.

**Site Release Link Consistency** is our trap, not yours. Several generated files embed counts (tests, CLI commands, MCP tools), so adding tests leaves them stale:

```
node scripts/generate_llms_txt.mjs && node scripts/sync_site_release_version.mjs
```

then commit the regenerated files. This has caught two other contributors, so we added a pre-push hook for it (`scripts/setup-hooks.sh` enables it locally).

**CI Gate** just aggregates those two and will clear itself.

Nothing in the test suites is failing. The `panicked at` lines in the macOS log are deliberate crash-recovery fixtures whose tests pass, which is confusing to read, and I misread them myself before checking the failing step name.

On the substance: the design principles you laid out in #420 are what make this reviewable at 1,300 lines. English staying canonical as the key, the additive runtime so a partial catalog never gates a build, and scoping the CJK font fallback so the Latin design is untouched in English are each the choice that avoids a long-term maintenance cost. Ping me when you take it out of draft and I will do a proper review.

silverstein:
Font question settled: the scoped CJK fallback is **approved**.

DESIGN.md requires explicit approval for any font outside the defined set, so rather than leave that in this thread I recorded it in DESIGN.md itself (#622). Otherwise the next contributor or agent reading that file correctly flags your PR as introducing an undefined font, and you would have to relitigate it.

The approval is scoped the way you built it, which is why it was easy to say yes to: the CJK stack is appended **after** Geist and Instrument Serif rather than ahead of them, and applies **only** when a CJK locale is active, so English rendering is unchanged. That scoping is the condition, and it extends to future locales on the same terms, so pt-BR (#124) would not need a fresh decision.

Nothing needed from you on this. Still just `cargo fmt --all` plus the doc regeneration when you are ready.

silverstein:
Both mechanical items are done, pushed to your branch. Nothing of yours was touched.

- `cargo fmt --all`, which is whitespace and trailing commas only. Verified rather than assumed: the comma-normalized token stream is byte-identical before and after.
- Regenerated the count-embedding files (`site/lib/release.ts`, the llms and docs catalogs), which go stale whenever a PR adds tests.

You will want to `git pull` before your next push.

Correcting something I said earlier, since I would rather you have the accurate version. I claimed nothing in the test suites was failing. That was true of CI, where the only failing step was the format check, but running the suite locally I saw about a dozen failures and briefly suspected your branch. It is not: I compared the failing sets on your branch and on main, and they differ **in both directions**, two tests failing only on yours and two only on main's, which is the signature of flaky tests rather than a regression. This repo has a known cluster of order and load sensitive tests, several of which I fixed last week. Nothing is attributable to your PR.

So the only thing between this and a review is taking it out of draft.

silverstein (approved):
Approved. Verified rather than taken on description, since this is 1,300 lines touching the desktop app, core, and config.

**The fallback contract holds.** `tr()` returns the English input unchanged unless the locale is zh-CN *and* the catalog has the key, so a missing or partial translation yields English rather than a blank or a key. That is what makes a partial catalog safe to ship and is the single most important property here.

**It is genuinely opt-in.** `language` defaults to `auto`, which resolves from the OS locale and falls back to English for anything unsupported, and `#[serde(default)]` means existing configs are unaffected.

**The CJK font is scoped exactly as approved.** Geist stays first and the system CJK stack is appended after it, injected only when `locale === 'zh-CN'`. English rendering is untouched, which was the condition. That exception is now recorded in DESIGN.md (#622) so nobody has to relitigate it, and it extends to future locales on the same terms, so pt-BR (#124) is unblocked.

I merged current main into your branch and regenerated the count-embedding files. Main had moved 28 commits, and the only conflict was `site/lib/release.ts`, which is generated. No i18n logic touched. Verified after the merge: core and the desktop app compile, i18n tests 4/4, fmt clean.

Design note worth recording for whoever adds the next language: keeping the runtime (`scripts/i18n.js`) separate from the catalog, and deliberately not gating CI on translation completeness, is what stops localization becoming a tax on every future UI change. That was the right call and it is why this was reviewable at this size.

silverstein:
Resolved the last red check, which was a genuine security guard rather than noise, and worth explaining since it will recur for anyone touching that file.

`tauri/src-tauri/src/main.rs` is a golden-hashed Apple Speech authority file. Any change to it fails the packaging parity guard until a maintainer reviews the diff and updates the recorded hash. Your PR legitimately modifies it (tray labels wrapped in `tr()`, plus the relabel path for switching language without a restart), so the guard fired exactly as designed.

I reviewed it line by line and resealed. Every change is localization; nothing touches Apple Speech, XPC, entitlements, codesigning, cdhash verification, or process spawning, so the trust boundary is unchanged. Part of the diff is also my own `cargo fmt` reflow of that file, which is whitespace and trailing commas only.

Nothing for you to do. Approval stands, all other checks were already green, and this merges once CI confirms.

silverstein:
Merged. This is the project's first localization, and it set the pattern the next ones will follow.

What made it mergeable at 1,300 lines was the design you proposed in #420 rather than the code volume: English staying canonical as the translation key, an additive runtime so a partial catalog never gates a build or release, and the runtime kept separate from the catalog. Those three choices are why adding pt-BR (#124) is now a catalog and a config value rather than an architecture discussion.

The scoped CJK fallback is recorded in DESIGN.md (#622) as an approved exception, on the terms you built it: appended after the primary faces, applied only when a CJK locale is active, English rendering untouched. Future locales inherit that, so nobody has to relitigate it.

For the record on the last red check, since it was not obvious: `tauri/src-tauri/src/main.rs` is a golden-hashed Apple Speech authority file, so touching it fails a packaging guard until a maintainer reviews the diff and reseals. Your change was localization only and left the trust boundary alone, so I reviewed and resealed it. The Windows failure after that was the known corpus-lease flake in #617, unrelated to your work.

Thank you. This one opens a door the project did not have.

PR #610, opened 2026-07-30
feat(macos): secure Apple Speech byte transport

silverstein:
## Summary

- add a dedicated one-shot Apple Speech XPC worker that accepts bounded 16 kHz mono PCM bytes, never an audio pathname
- authenticate the embedded service and parent through sealed code-signing requirements and an exact embedded worker CDHash before sending private utterance bytes
- construct `AVAudioPCMBuffer` directly in the Swift bridge, with no named plaintext WAV or generic process-launch surface
- add direct Apple parser/state-machine negative tests, stronger packaging mutations, and a fixed-input signed-app acceptance route
- add a post-signing macOS 26 job that exercises the real byte path under a same-UID open-holder watcher and verifies fallback remains Whisper
- keep Apple Speech product selection fail-closed to Whisper pending a separately authorized exact-SHA signed runtime run

## Exact scope

- base: `6ac1c4f18361343ff088b4f5fd45cf60dce092a6`
- transport implementation: `eccdd7791c0a6053b5eb13c05ffa201dcb2ec355`
- acceptance/test remediation: `415ae1214b39adee525fee895e0c81015d25d738`
- build and workflow-boundary remediation: `1830f48a47cb032d709e2833205d8be048123031`
- final macOS clippy-scope remediation: `0501e06a6e3bb559d91f6c291f1234964f5a0f36`
- exact head: `0501e06a6e3bb559d91f6c291f1234964f5a0f36`
- bead: `minutes-hueo`

## Exact-head local evidence

- `cargo check -p minutes-core --no-default-features`: passed
- `cargo test -p minutes-core --no-default-features --lib -- --test-threads=1`: 1,651 passed, 0 failed, 1 ignored on Linux; seven additional parser/acceptance/XPC tests are macOS-only
- `cargo check -p minutes-core --features streaming,whisper`: passed
- `cargo test -p minutes-core --features streaming,whisper --lib pathname_only_live_backends_resolve_to_whisper`: passed
- `cargo test -p minutes-core --features streaming,whisper --lib accumulate_insert_defers_delivery_to_flush_not_per_utterance`: passed
- `cargo clippy -p minutes-core --no-default-features --features parakeet -- -D warnings`: passed
- `cargo test -p minutes-core --features parakeet --lib write_wav_roundtrip`: passed
- Apple Speech and graph packaging guards and mutation self-tests passed
- signed-acceptance policy tests and changed-workflow actionlint passed
- generated agent docs and portable skill audits passed during push
- production site build passed

## Review remediation

[...]

PR #624, opened 2026-08-03
fix(ci): hash the post-signing region and prove the goldens tile the workflow

silverstein:
Closes the unhashed region in the signed acceptance policy guard.

## The gap

The guard hashed two regions — everything before the signing job, and the signing job itself. When #610 added `run-signed-runtime` *after* the signing job, the new job landed outside both goldens. **2,027 bytes of workflow were checked for exactly one thing: whether they mention `secrets`.**

That job matters. It checks out the candidate at its acceptance tag and runs `scripts/test_apple_speech_signed_transport.py` to produce the provenance receipt acceptance relies on. It holds no secrets — which is precisely why the one pattern that looked at it never fired.

## Verified before fixing

Against the **live** guard and **live** workflow from `main`, all of these passed with exit 0:

| mutation | live guard |
|---|---|
| append a step at EOF | **accepted** |
| append an entire new job at EOF | **accepted** |
| swap the acceptance script for another path | **accepted** |

The third only trips a regex that happens to mention the script name; nothing hashed the job.

## The fix

The post-signing region is hashed like the other two.

Hashing it alone would fix this instance and not the next one — the gap opened *silently*, when a job was appended. So the three regions must also **tile the file exactly**: if their lengths don't sum to the file's, the guard fails and reports how many bytes are outside every golden.

```
hashed regions cover 21271 of 22798 bytes; some of this workflow is outside every golden
```

That assertion holds even when the region logic is edited to under-cover **and** its golden is updated to match the shrunken region — verified by executing exactly that attack.

## Tests

Three mutations added, all of which the previous guard accepted:
- a step appended after the runtime job
- a whole job appended after it
- the receipt's upload weakened to `if-no-files-found: ignore`

Each is anchored on a string unique to the runtime job. `if-no-files-found` appears three times in this workflow, and replacing the first match mutates the unsigned-build job instead — which a *different* golden already covers, so the test would pass while proving nothing about the region under test. I hit that while writing it.

Reverting either half of the fix fails the suite.

## Scope

The workflow itself is **unchanged**. This only closes the hole in what the guard reads.

### DIFF
--- a/scripts/check_signed_acceptance_policy.mjs
+++ b/scripts/check_signed_acceptance_policy.mjs
@@ -206,13 +206,17 @@ if (!signingJobFixture) {
   // unreviewed again -- which is precisely how the region above was
   // introduced. A coverage failure means a region boundary moved, so the
   // goldens no longer mean what they claim, whatever they hash to.
+  //
+  // Stated as a concatenation rather than as arithmetic on lengths: the
+  // identity is then self-evident and cannot hold by coincidence. Comparing
+  // sums also invited an off-by-encoding reading, since `.length` counts
+  // UTF-16 code units and the message said "bytes".
   const covered =
-    preSigningBoundary.length + signingJobMarker.length + signingJobRegion.length +
-    afterSigningJob.length;
-  if (covered !== source.length) {
+    preSigningBoundary + signingJobMarker + signingJobRegion + afterSigningJob;
+  if (covered !== source) {
     errors.push(
-      `hashed regions cover ${covered} of ${source.length} bytes; ` +
-        "some of this workflow is outside every golden",
+      `hashed regions reconstruct ${covered.length} of ${source.length} characters ` +
+        "and do not reproduce this workflow; some of it is outside every golden",
     );
   }
   const beforeSigningJob = source.split(/^  sign-reviewed-artifact:\n/m, 1)[0];
--- a/scripts/check_signed_acceptance_policy.mjs
+++ b/scripts/check_signed_acceptance_policy.mjs
@@ -8,6 +8,8 @@ import { readFileSync } from "node:fs";
 // authority: these goldens make comment/dead-step duplication fail closed.
 const EXPECTED_SIGNING_JOB_SHA256 =
   "903c7c962b5eac067fbe423699943969b0405cec732306299366f172f5037894";
+const EXPECTED_POST_SIGNING_BOUNDARY_SHA256 =
+  "e0a1feef44fa29000affee57a51a55156c232d677307181a68e9cc2341594807";
 const EXPECTED_PRE_SIGNING_BOUNDARY_SHA256 =
   "1ddf57258d26c54dab346639d179a4cce64e6d3e333bb92982b38706ac7fe333";
 const EXPECTED_TRIGGER_BLOCK = `on:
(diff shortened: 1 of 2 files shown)


silverstein:
Adversarial review came back. Summary and what changed in `ed8867af`.

## The guard holds

Verified by execution, not reading — the three goldens plus the tiling assertion make this **provably equivalent to a whole-file hash**:

- ~250k single-code-unit mutations at all 22,798 positions (delete / substitute / insert, incl. astral chars) — **0 accepted**
- exhaustive line-level mutations over all 511 lines — **0 accepted**
- 20,000-trial randomized multi-edit fuzz — **0 accepted**
- 22 hand-built structural attacks — boundary insertion, exploiting the newline `signingJob` strips, a job interposed between the signing and runtime jobs (incl. `Uppercase`/`_underscore` names that dodge the next-job regex), same-length rename, CRLF, BOM, trailing whitespace, YAML anchors, a second `sign-reviewed-artifact:` — **all rejected**

## What was wrong: the coverage check was untested

Deleting it outright, or weakening `!==` to `>`, left **all eleven** suite mutations passing. My commit message claimed "reverting either half of the fix fails the suite" — only the post-signing hash half did. That's corrected.

The workflow mutations can't reach it by construction: for an edit to the workflow alone, a matching pre-signing golden already implies the regions tile. Its purpose is defending against a future edit to the *guard* that narrows a region — so the only test that can kill it has to mutate the guard.

`ed8867af` adds that test. It runs the attack the check exists to stop: append a job that curls the acceptance receipt to an external host, narrow `afterSigningJob` so the new job falls outside it, then regenerate **all three** goldens from the shrunken regions — exactly what someone following the "review it and update its golden hash" instruction would do. Every hash agrees; only coverage objects.

Both mutants now die:
- coverage check deleted → `a narrowed region with regenerated goldens was accepted; the coverage check is not load-bearing`
- `!==` weakened to `>` → same

## Also changed

The check is now a concatenation identity rather than arithmetic on lengths — self-evident instead of true by coincidence, and it drops a misleading message (`.length` is UTF-16 code units, not bytes; the file is ASCII today, which made "bytes" accidentally correct).

Second correction: my test comment claimed anchoring on the first `if-no-files-found` match would let the test pass for the wrong reason. It wouldn't — the harness's expected-substring check catches it, verified by running it. The anchoring is still right; my stated reason was wrong.

## Noted, not addressed here

- `--signing-job-fixture` skips the goldens and coverage, but nothing in CI routes a real workflow through it (`lint-workflows.yml:93` uses it only on two negative fixtures) — **not a bypass**.
- Pre-existing: a *missing* negative fixture makes the guard exit 1, which `lint-workflows.yml:82-98` reads as "correctly rejected", so deleting a fixture silently disables its test.
- Pre-existing: nothing hashes `lint-workflows.yml` itself, and nothing prevents a *new* workflow declaring `environment: signed-dev-acceptance` — the real gate there is GitHub's environment reviewers.
- Scope: hashing the runtime job pins *which* script runs and with what arguments. `test_apple_speech_signed_transport.py` comes from the candidate ref and is not hash-pinned, so what the receipt *says* is closed at the workflow level only. That's the acceptance design, but my earlier wording read stronger than what's enforced.

Issue #633, opened 2026-08-03
Parakeet transcription aborts the CLI: race between main and the sidecar-stderr thread


theklr:
### Summary
Every transcription attempt that routes through the Parakeet engine crashes the `minutes` desktop app with `EXC_CRASH (SIGABRT)`, triggered on the `parakeet-sidecar-stderr` thread. This is 100% reproducible across multiple recordings and multiple relaunches, and happens before a processing job record is even created (so there's no retry path — the job just disappears).

### Environment
- Minutes desktop app (macOS), `app_version` reports as `???` in the crash log
- macOS 26.6 (25G70)
- MacBook Pro 18,2, Apple M1 Max, 32GB RAM
- Parakeet model: `tdt-600m`, sidecar launched with `--gpu --vad`
- `libomp` confirmed installed and present at `/opt/homebrew/opt/libomp/lib/libomp.dylib` (Cellar 22.1.8)

### Steps to reproduce
1. Record a meeting (screen + mic) with Parakeet selected as the transcription engine.
2. Stop the recording. Processing kicks off and fails.
3. Re-submit the same audio for processing (via `process_audio` MCP tool or CLI).
4. App crashes with SIGABRT before a job record appears in `list_processing_jobs`.

Reproduced this identically 3 times across separate app launches (PIDs 4976, 6431, and one earlier instance), same crashing thread each time.

### History leading up to this
1. Original symptom: two recordings both failed with a *handled* error (job recorded as `failed`, not a crash):
   ```
   parakeet transcription failed:   Reason: tried: '/opt/homebrew/opt/libomp/lib/libomp.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/libomp/lib/libomp.dylib' (no such file), '/opt/homebrew/opt/libomp/lib/libomp.dylib' (no such file)
   ```
2. Installed `libomp` via Homebrew — confirmed present at the exact path the error was looking for.
3. Noticed two orphaned `parakeet-sidecar` processes running concurrently, with socket filenames referencing PIDs that didn't match the actual running process PIDs (e.g. socket `parakeet-sidecar-3817-tdt-600m.sock` owned by process PID 3843), suggesting a prior crash/restart had already left stale sidecar state around. Killed both stale processes and removed their socket files.
4. Re-attempted processing — now the app hard-crashes (SIGABRT) instead of surfacing the libomp error. Reproduced 3x.

### Crash log (representative — thread 1, faulting thread)
```
[...]

Issue #369, opened 2026-06-26
Sherpa engine is unreachable for desktop-app and release-binary users (CLI source-build only)


silverstein:
## Summary

0.19.0 shipped the sherpa engine (in-process parakeet-tdt-0.6b-v3) as a CLI build-from-source opt-in. As shipped, it is NOT reachable by:
- anyone using a released binary (DMG, Homebrew, cargo, GitHub CLI assets), or
- the Tauri desktop app at all (no feature path, even from source).

So the "Recommended SOTA engine" in the README is currently only usable by people who build the CLI from source with `--features engine-sherpa`.

## Evidence (verified)

1. No release workflow includes `engine-sherpa`:
   - release-cli.yml: `parakeet,metal` (macOS), `parakeet` (linux/windows)
   - release-macos.yml: `parakeet,metal` (CLI + tauri)
   - release-windows-desktop.yml: `parakeet`
   - Homebrew formula builds from source with default features (no sherpa)
   Result: `engine = "sherpa"` on a released binary silently falls back to Whisper.

2. The tauri crate (`tauri/src-tauri/Cargo.toml`) has no `engine-sherpa` passthrough. Its features are `parakeet/coreml/cuda/hipblas/metal/vulkan`, and it depends on minutes-core with `features = ["streaming", "diarize"]`. There is currently no way to build the desktop app with sherpa.

3. Docs (README section + docs/SHERPA-ENGINE.md) describe only the CLI source-build path. README calls sherpa the "Recommended SOTA engine" without flagging that app + release-binary users cannot reach it today.

## Proposed fix

1. Add `engine-sherpa = ["minutes-core/engine-sherpa"]` to the tauri crate features. (Verified locally: the app builds and runs with this added.)
2. Add `engine-sherpa` to the release builds (release-cli, release-macos, release-windows-desktop) so installed binaries contain the engine. Confirm the sherpa-rs native lib is correctly linked/bundled and codesigned in the .app and .exe on each OS. The ~670MB model stays a separate `minutes setup --sherpa` download, and the existing graceful Whisper fallback covers the no-model case, so shipping the engine on-by-default in binaries is low risk.
3. Validate that Linux and Windows bundles actually run sherpa end to end before flipping it on (the opt-in-first / protect-Win-Linux caution).
4. README/docs: state current availability accurately, and simplify the build-from-source framing once binaries include the engine.

## Context

[...]

PR #644, opened 2026-08-04
feat(release): ship the sherpa engine in macOS builds

silverstein:
## Why

Benchmarked on LibriSpeech test-clean, 12 clips, 777s of audio, 2050 reference words, on an M1 iMac, all three engines on identical `v0.24.0` code with an isolated `XDG_CONFIG_HOME` per engine:

| Engine | WER | Errors / 2050 | Wall clock | RTF |
|---|---|---|---|---|
| sherpa parakeet-v3-int8 | **2.29%** | 47 | 221s | 0.28 |
| parakeet tdt-ctc-110m (helper) | 2.39% | 49 | 187s | 0.24 |
| whisper small (current default) | 3.46% | 71 | ~2224s | 2.86 |

Harness calibration: whisper-small measured 3.46% against the 3.4% published in `site/app/writing/whisper-cpp-vs-parakeet-cpp`, so this reproduces the existing methodology rather than inventing a new one.

**int8 quantization was the risk and it did not materialize.** Sherpa ties the external parakeet helper (47 vs 49 errors in 2050 words is noise, not a win) and cuts errors by roughly a third against the shipping default while running an order of magnitude faster in wall clock.

The decisive property is not the tie with parakeet, it is that sherpa runs **in-process**. It needs no pathname-based private-audio transport, so it is unaffected by the gate that makes `engine = "parakeet"` resolve to Whisper, and it cannot spawn the sidecar that crashes in #633.

## Scope

macOS only, deliberately. `sherpa-onnx` links **statically** on darwin (self-contained binaries, #369) but **dynamically** on Linux and Windows, where the binary needs `libsherpa-onnx-c-api` alongside it. I hit exactly that failure locally. Enabling it on those targets without bundling the dylibs and fixing rpath would ship binaries that fail to start.

## What this does not do

Compiling the engine in does not bundle the model. Users still opt in explicitly with `minutes setup --sherpa` (~670 MB) and `engine = "sherpa"`. Default stays whisper.cpp.

No measurable size cost: the shipped v0.24.0 macOS CLI is already 33 MB and a sherpa build measures 32 MB. (Separately, the "6.7MB binary" line in CLAUDE.md is stale.)

## Limits of the evidence

LibriSpeech is clean read speech. It says nothing about meeting audio, overlapping speakers, or domain vocabulary, which is where I would expect the gap to widen further given whisper-small rendered "Parakeet" as "perikete" in an earlier sample. A proper-noun run on real meeting audio is the natural follow-up and is not in this PR.

Follow-up worth tracking: bundling the sherpa dylibs so Linux and Windows can ship the engine too.


silverstein:
## Correction: the speed claim in this PR is wrong

The WER numbers stand. The speed numbers do not, and the recommendation changes with them.

**What went wrong.** The LibriSpeech run was built `--features parakeet,engine-sherpa`, with no `metal`. Whisper.cpp therefore ran CPU-only, while every shipping macOS artifact is built with Metal. I compared sherpa against a deliberately hobbled whisper and reported the result as a 10x win.

**Measured again on an M4 Max with a Metal-enabled build** (`parakeet,metal,engine-sherpa`), on real meeting audio:

| Engine | 4 min audio | 6 min audio | RTF |
|---|---|---|---|
| whisper-small (Metal) | 4.2s | 5.2s | ~0.015 |
| sherpa v3-int8 | 27.0s | 30.7s | ~0.09 |

Whisper-small is roughly **6x faster** than sherpa on the configuration that actually ships. Sherpa's onnxruntime path is not using the GPU; whisper.cpp is.

**Domain vocabulary did not favour sherpa either.** On a real meeting containing a product name whisper had already mangled, both engines failed and neither was self-consistent:

- sherpa: `patientity`, `patient tristi`, `patient trity`
- whisper-small: `patientricity`, `patient tricities`, `patient tricity`

Whisper-small got `RxVIP` right; sherpa produced `RX V`. I expected sherpa to widen its lead here and it did the opposite.

## Revised position

What survives is the accuracy result on clean read speech: 2.29% vs 3.46% WER, a third fewer errors, measured identically for both engines and unaffected by Metal.

So this PR should land as it is written, because it only makes the engine **available** and opt-in behind `minutes setup --sherpa` plus `engine = "sherpa"`. It should **not** be read as an argument for changing the default. On Metal hardware the default is faster and no worse on domain nouns.

Worth investigating separately: whether sherpa can use a CoreML or GPU execution provider. Today's comparison is Metal-accelerated whisper against CPU-only sherpa, which is fair to what ships but not a ceiling for sherpa.

PR #651, opened 2026-08-04
feat(recall): support OpenAI-compatible local providers in chat (#650)

### LINKED ISSUE #650: Not able to chat in recall
Hi, I am using mac os.  After the transcribe and when i click recall and when i say 'hi' in the chat. It is giving me this error:
Error: No AI agent found for Recall chat. Install Claude Code (npm install -g @anthropic-ai/claude-code), Codex, or Gemini CLI to enable chat — or configure Ollama ([summarization] engine = "ollama" in config.toml) for a fully local option. Minutes doesn't require an API key for this feature.

I use LM Studio for summary. But not able to chat with it.
This is the config:

[summarization]
engine = "openai-compatible"
agent_command = "ollama"
agent_timeout_secs = 300
speaker_mapping_timeout_secs = 30
chunk_max_tokens = 4000
ollama_url = "http://localhost:1234/v1"
ollama_model = "qwen/qwen3.5-9b"
openai_compatible_base_url = "http://localhost:1234/v1"
openai_compatible_model = "qwen/qwen3.5-9b"
openai_compatible_api_key_env = ""
mistral_model = "mistral-large-latest"
language = "auto"

Instead of ollama url i am using lmstudio url as i use lmstudio for everything

silverstein:
Closes #650.

## Problem

Recall chat accepted only `engine = "ollama"` or an installed agent CLI. A user running LM Studio through `engine = "openai-compatible"` could summarize meetings fine but hit *"No AI agent found for Recall chat"* the moment they typed a message. Their config was reasonable; summarization and chat simply supported different engine sets.

There is also a trap worth naming: pointing `ollama_url` at LM Studio does **not** work around it. The Ollama path calls the native `/api/chat` NDJSON endpoint, while these servers speak OpenAI-compatible `/v1/chat/completions`. It would fail differently rather than work.

## Change

A third provider path with the same local-only contract as Ollama.

**`recall_openai_compatible_chat_url`** requires a loopback destination and refuses credentials, query strings and fragments. Unlike the Ollama guard it deliberately **preserves the path**, because these servers are rooted at a prefix, so `http://localhost:1234/v1` becomes `http://localhost:1234/v1/chat/completions`. It does not double-append when the base already names the endpoint.

**`reauthorize_recall_openai_compatible_egress`** reloads config and re-checks engine, URL and model immediately before the request, mirroring the Ollama guard, so a provider swapped mid-turn cannot receive context authorized against the previous one.

**`openai_compatible_stream_delta`** parses server-sent events, ignoring keepalives, the `[DONE]` sentinel, role-only frames, finish frames and empty deltas.

No API key is sent. The destination is loopback by construction, and adding credential handling would widen the surface for no benefit.

The fallback error message now names both local options rather than pointing only at Ollama.

## Security notes for review

This touches the Recall egress boundary, so the parts worth scrutiny:

- Local-only is enforced the same way as Ollama: `http` scheme, loopback host (`localhost`, any loopback IP, `[::1]`), no userinfo, no query, no fragment.
- The only intentional relaxation is allowing a path prefix, which the Ollama guard refuses. That is required for this protocol and cannot reach a different host.
- Config is reloaded inside the worker rather than trusted from turn start, matching the existing pattern.
- History is cleared and the turn aborted if reauthorization fails.

## Testing

New tests cover prefix preservation, idempotent endpoint append, IPv6 loopback, and refusal of https, non-loopback hosts, embedded credentials, query strings and fragments; plus the quiet SSE frame cases.

Verified the guard is load-bearing by disabling the loopback check and confirming the test fails.

`cargo test -p minutes-app --no-default-features -- --test-threads=1`: **287 passed, 0 failed**. clippy clean.

## Not verified

I have no LM Studio instance to run end-to-end. The URL construction, egress guard and SSE parsing are unit-tested, but a real streaming session against LM Studio has not been exercised. Worth a manual check before release, and #650's reporter is well placed to confirm.


silverstein:
## Verified against a real server, and it found a bug the unit tests could not

The maintainer has LM Studio on a Mac, so I checked. LM Studio's app was running but its local server was not listening (it is off until you start it in the Developer tab), so I verified against ollama's OpenAI-compatible endpoint instead, which implements the same protocol.

The result was worth the trip. `qwen3.5:4b` streamed **3861 frames**, reached `finish_reason: "stop"`, and produced **zero non-empty `content` deltas**. The whole budget went into a `reasoning` field:

```
data: {"choices":[{"index":0,"delta":{"role":"assistant","content":"","reasoning":" friend"},"finish_reason":null}]}
data: {"choices":[{"index":0,"delta":{"role":"assistant","content":""},"finish_reason":"stop"}]}
data: [DONE]
```

Delta keys observed across the stream: `reasoning`, `content`, `role`.

The parser did the right thing and ignored all of it. The problem is what the user sees: an empty panel with no explanation, indistinguishable from a hang.

### Change

Emits a specific error when a stream completes with no answer text, pointing at the likely cause (a reasoning model spending its budget thinking) and at what to change. Adds the real frame shapes as regression cases, including the blank separator lines real servers interleave.

### Still not verified

An actual LM Studio stream. The protocol is the same and the frame handling is now exercised against real traffic, but LM Studio's exact framing has not been observed. Worth one manual check before release.

### Related gap, deliberately not touched

The native Ollama path has the same blank-answer behavior. Left alone to keep this reviewable; worth a follow-up.

287/287 desktop tests, clippy clean.

silverstein:
## Verified against real LM Studio

The maintainer started LM Studio's server (`http://127.0.0.1:1234`, `google/gemma-4-12b-qat`), so the remaining gap is closed.

A real streaming turn: **508 data frames, 141 KB, terminated with `[DONE]`**. Delta keys observed: `content`, `reasoning_content`, `role`.

Running the shipped parser logic over that capture reconstructs the answer exactly:

```
frames with real content: 4
reconstructed: 'Hello to you.'
```

504 of 508 frames were `reasoning_content` and were correctly ignored; the 4 content frames produced the answer.

### Worth recording

The two servers use **different reasoning keys**: ollama emits `reasoning`, LM Studio emits `reasoning_content`. The parser is unaffected because it reads `content` specifically rather than inferring which fields are answer text. That is the behaviour to preserve if anyone extends this later.

### UX note, not a blocker

With a reasoning model the panel stays silent for the whole thinking phase and then prints a short answer. Correct, but it looks idle. Surfacing reasoning as a "thinking" indicator would be a reasonable follow-up; deliberately not in this PR.

The empty-answer error added in the last commit does not fire here, since content is non-empty. It remains the guard for the ollama case where a model finishes with no content at all.

Both real-server cases are now exercised: LM Studio (answer produced) and ollama (no answer produced).

silverstein:
## Added: a thinking status while a local model reasons

The real-server capture showed the panel would render nothing for almost the entire turn. LM Studio's gemma sent **504 of 508 frames** as `reasoning_content` and only 4 as answer text. That span is silent, which reads as a hang rather than a model working.

The design already accounted for this. From the CSS at `index.html:665`:

> Pre-first-token grounding state: an empty streaming bubble narrates what the agent is doing via data-status — "thinking", then live MCP tool labels — with pulsing dots, so a slow agent cold-start reads as working, not frozen (#404).

Only the Claude CLI path ever set that attribute. The local-model paths left it unset, so users got bare dots.

**Change:** the backend emits a one-shot `{"type": "thinking"}` chunk on the first reasoning frame of a turn, and the panel sets `data-status = 'thinking'`. Once per turn, because it drives a label rather than rendered text, and the label removes itself as soon as real text arrives (the bubble is no longer `:empty`).

Recognizes both dialects, since the servers disagree: ollama emits `reasoning`, LM Studio emits `reasoning_content`. Reasoning is never rendered as an answer — it is the model's scratchpad, not its reply.

## Verification status

- Rust: 288/288 desktop tests, clippy clean. New unit tests cover both dialects and the cases that must *not* count as reasoning (empty string, role-only, content frames, `[DONE]`).
- **Not verified: the rendering.** Per the repo's UI rule, `index.html` changes need a click-test in `Minutes Dev.app`, and unit tests cannot cover it. Codesigning also fails over SSH (`errSecInternalComponent`), so I cannot build a trustworthy dev app remotely.

To click-test locally, with LM Studio serving and `engine = "openai-compatible"`:

```
cd ~/Sites/minutes
git fetch origin && git checkout feat/recall-chat-openai-compatible
./scripts/install-dev-app.sh
```

Then open `~/Applications/Minutes Dev.app`, ask Recall anything, and confirm the bubble reads `thinking • • •` during the pause and that the label disappears when the answer starts.

silverstein:
Picked this up: resolved the conflicts, ran an adversarial security review, and fixed what it found.

## Conflicts

Six, all in generated agent-docs files where both sides had only bumped the "Last generated" date stamp. Took main's side and re-ran `scripts/generate_llms_txt.mjs`; `--check` passes. This PR's additions are Tauri commands, which those files don't enumerate, so nothing from the branch was lost. Branch was 99 commits behind; now merged and `MERGEABLE`.

## Security review found a real one, and it wasn't yours

The review blocked on the egress guard accepting the literal string `localhost` without resolving it. The guard runs once, but `ureq` resolves the name again at request time through the system resolver, so a poisoned or merely misconfigured NSS entry would send meeting context to a remote host while the configured URL still reads as local. For a feature whose entire contract is "chat never leaves the machine," the enforcement point was trusting a name it never checked.

**The weakness pre-existed on `main`** in the Ollama guard; this PR replicated the pattern for the new path. So blocking the feature would have protected nobody while leaving the existing hole open. Both guards now share `recall_destination_is_loopback`, which checks an IP literal directly and otherwise resolves and requires *every* returned address to be loopback, failing closed on empty or errored resolution. Redirects and proxies were already disabled, so DNS was the remaining way out.

Added tests pinning what regressed silently before: IP literals accepted without resolution, routable literals refused (`0.0.0.0`, `10.0.0.5`, IPv4-mapped IPv6), unresolvable names failing closed, `localhost` acceptance now flowing through resolution rather than spelling, and userinfo confusion (`http://127.0.0.1@evil.com`) refused. 7 pass locally, including the existing Ollama guard test.

## What the review cleared

Worth recording, since these were the obvious attacks: userinfo confusion, `0.0.0.0`, exotic IPv4 encodings, IPv4-mapped IPv6, trailing-dot names, and non-equivalent IDNA names are all blocked by the parsed-host validation. Path construction uses `set_path`, so a path can't replace the authority. Redirects and proxies are disabled. Config changes after reauthorization can't alter the already-captured request URL. No API key is attached, payloads aren't logged, and model output is escaped before `innerHTML`.

## Two findings I did not fix, filed instead

Both are pre-existing patterns rather than regressions, and both apply to the Ollama path too, so they want their own change:

1. **Unbounded response body.** `body.as_reader()` is unlimited, `BufRead::lines()` allocates a whole line before returning, and deltas append to an uncapped `String`. A malfunctioning or hostile local server can OOM the desktop app; the 120s timeout bounds time, not bytes. Wants total-body, per-frame, and accumulated-answer caps.
2. **Partial output persisted as a complete answer.** On a read error the loop logs and breaks, and any text received is then stored as successful conversation history with no failure signal to the UI.

PR #654, opened 2026-08-05
fix(build): run tauri builds from the app crate after the Archive app joined the workspace

silverstein:
## The bug

`feat(archive): ship private census app` added a second `tauri.conf.json` at `archive/src-tauri/`. The tauri CLI resolves the first config it finds from the working directory, so **any `cargo tauri build` run from the repo root now targets `minutes-archive-app` instead of `minutes-app`.**

Reproduced on a Mac against the real toolchain:

```
# from the repo root
$ cargo tauri build --bundles app --features parakeet,metal --no-sign
error: the package 'minutes-archive-app' does not contain these features: metal, parakeet

# from tauri/src-tauri
$ cargo tauri build --bundles app --features parakeet,metal --no-sign
   Compiling ...        # builds the real app
```

## Blast radius

Two CI jobs already set `working-directory: tauri/src-tauri` and were unaffected — the Windows installer in `ci.yml` and in `release-windows-desktop.yml`. That is why this did not show up as a red build.

Three call sites did not:

| site | consequence |
|---|---|
| `release-macos.yml:205` | passes `--features parakeet,metal,engine-sherpa`, so **the next macOS release would have failed at tag time** |
| `scripts/install-dev-app.sh` | fails today, blocking the dev-app workflow the repo prescribes for UI verification |
| `scripts/build.sh` | same shape |

The failure mode with `--features` is loud. **The no-features case is worse:** a root-level `cargo tauri build` with no feature flags builds the archive app and *succeeds*, silently producing the wrong bundle.

## Change

Each of the three now runs from `tauri/src-tauri`, matching what the two healthy CI jobs already do. `install-dev-app.sh` keeps its `--config` working by passing an absolute path.

## Golden hashes

`release-macos.yml`, `build.sh` and `install-dev-app.sh` are all covered by the graph-XPC packaging guard, so it fired. Reviewed before resealing: each diff is confined to the working directory of the tauri build, and none of them touch signing, entitlements, helper packaging or ordering.

## How this was found

The dev-app build failed while preparing a UI click-test for #651. It looked like the known SSH codesigning limitation at first; it was not, and reading the actual error rather than assuming the familiar cause is what surfaced it.

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -116,6 +116,51 @@ jobs:
         with:
           files: ${{ matrix.asset }}
 
+      # Additive second artifact for Linux only: the same CLI plus the sherpa
+      # engine and its shared libraries, as a tarball. The bare binary above is
+      # unchanged, so existing installs and the Homebrew formula are unaffected.
+      #
+      # sherpa-onnx links dynamically on Linux, so the binary needs
+      # libsherpa-onnx-c-api.so and libonnxruntime.so beside it. `$ORIGIN`
+      # runpath makes the loader look next to the executable rather than in the
+      # build tree. Static linkage is deliberately not used here: sherpa-rs-sys
+      # demands a global `-C relocation-model=dynamic-no-pic`, which produces a
+      # non-PIE binary and loses ASLR for every Linux user (#645).
+      - name: Build sherpa-enabled Linux archive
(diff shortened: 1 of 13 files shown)


silverstein:
Superseded by a clean version: I branched this off the recall-chat branch by mistake, so it duplicated that work, and 6c1017cf has since fixed two of the three call sites on main. Reopened with only the remaining build.sh fix.

PR #659, opened 2026-08-06
fix(windows): statically link the MSVC runtime so binaries start on a clean PC

### LINKED ISSUE #657: Windows: every published artifact needs the VC++ runtime, none ship it (silent DLL_NOT_FOUND)
## Summary

`minutes-windows-x64.exe` from the v0.24.0 release **does not start** on a Windows machine that has never had developer tools installed. It exits with `0xC0000135` (`STATUS_DLL_NOT_FOUND`) and prints **nothing at all**, so the user sees the command silently do nothing.

## Evidence

Tested on a real Windows 11 Enterprise LTSC 2024 (build 26100) VM, x64, with no Visual Studio, no build tools, and no VC++ redistributable:

```
PS> Test-Path C:\Windows\System32\vcruntime140.dll
False
PS> Test-Path C:\Windows\System32\msvcp140.dll
False

PS> & C:\qa\minutes.exe --version
exit=-1073741515      # 0xC0000135 STATUS_DLL_NOT_FOUND
output=               # empty
```

The binary was hash-verified against the published asset (`b576503487610424d3d7e8c2…`) before and after transfer, so this is the shipped artifact rather than a corrupted download.

Its import table names the missing dependencies directly:

```
VCRUNTIME140.dll
VCRUNTIME140_1.dll
MSVCP140.dll
MSVCP140_1.dll
```

Those ship in the **Microsoft Visual C++ Redistributable**, not in Windows. `DirectML.dll` is also imported but is present in stock Windows 11, so it is not implicated.

## Why CI never caught it

GitHub's `windows-latest` runners have the VC++ redistributable preinstalled, so every build and smoke test passes. The failure only appears on an end-user machine that has never had it. `Install CLI (windows-latest)` is green today and will stay green.

## Impact

Any Windows user who downloads the CLI onto a clean machine gets a silent no-op with no diagnostic. There is no message to search for, which makes it close to unreportable from the user's side. Worth considering how many quiet Windows abandonments this accounts for.

## Options

1. **Statically link the CRT** — `-C target-feature=+crt-static` for `x86_64-pc-windows-msvc`. Self-contained, matches the "single binary" story. Requires the bundled C++ (whisper.cpp/ggml) to be built `/MT` so the runtimes agree; mismatched CRTs are their own class of bug, so this needs verifying rather than assuming.
2. **Ship the redistributable DLLs beside the exe.** App-local deployment is permitted by Microsoft. Changes the CLI download from one file to an archive, which is the same tradeoff #645 raises for sherpa.
3. **Document the prerequisite** and have the installer ensure it. Cheapest, but does nothing for the standalone CLI download, which is the path that fails silently.

Option 1 is the only one that keeps a single self-contained binary. Worth checking whether the NSIS desktop installer already installs the redistributable, since the desktop app may be unaffected while the CLI is not.

## Relationship to #645

That issue proposes bundling sherpa's DLLs for Windows. This one should be settled first: adding more dynamically-linked dependencies on top of a runtime that is already missing would compound the failure rather than introduce it.

Found using the Windows QA VM rather than CI, because the defect is specifically about what is *absent* on a clean machine.

silverstein:
Fixes #657.

## The problem

Every published Windows artifact imports `VCRUNTIME140.dll`, `VCRUNTIME140_1.dll`, `MSVCP140.dll` and `MSVCP140_1.dll`. Those come from the Visual C++ Redistributable, which is **not** part of Windows.

Verified against the released v0.24.0 artifacts on a clean Windows 11 (26100) VM with no VC++ runtime present:

```
minutes-windows-x64.exe --version         -> exit -1073741515 (0xC0000135), no output
minutes-desktop-windows-x64.exe --version -> exit -1073741515 (0xC0000135), no output
```

No window, no error, no message. A user downloads Minutes, runs it, and nothing happens.

CI is green and will stay green: GitHub's `windows-latest` runners ship the redistributable. The failure only exists on end-user machines, and produces no text to search for, so its absence from the issue tracker means nothing.

## Why static linking rather than shipping the DLLs

It keeps the single self-contained binary. Bundling would have to be solved twice, once for the bare CLI download and once for the NSIS installer, and the installer runs `installMode: currentUser` so it cannot install a machine-wide runtime anyway.

## The risk in this change

The C++ side has to use the same runtime or the link fails with a runtime-library mismatch. `cc` reads `crt-static` from `CARGO_CFG_TARGET_FEATURE` and switches to `/MT` (`cc/src/lib.rs:2098`), and `cmake` inherits its flags, so this should propagate to the bundled whisper.cpp. **That propagation is the thing to watch.** The Windows CI jobs (`Test`, `Install CLI`, `Desktop App Installer`) all compile the C++ and will fail loudly if the runtimes disagree.

macOS and Linux are untouched: the setting is scoped to `x86_64-pc-windows-msvc`.

## Verification plan

CI proving it builds is necessary but not sufficient, because the whole point is behaviour on a machine CI does not resemble. After it goes green I will trigger `release-cli.yml` via `workflow_dispatch`, revert the QA VM to its clean snapshot, and run the artifact there. Merging before that check would be repeating the mistake that let this ship.

Binary size impact will be reported from the built artifact rather than estimated.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,22 @@
 # The cc crate defaults to 10.13 for some targets, causing build failures.
 # Setting this ensures the correct -mmacosx-version-min flag. (See #14)
 MACOSX_DEPLOYMENT_TARGET = { value = "11.0", force = true }
+
+# Statically link the MSVC C and C++ runtimes on Windows.
+#
+# Without this, every published Windows artifact imports VCRUNTIME140.dll,
+# VCRUNTIME140_1.dll, MSVCP140.dll and MSVCP140_1.dll, which ship in the Visual
+# C++ Redistributable rather than in Windows. On a machine that has never had
+# it, the process fails to start with STATUS_DLL_NOT_FOUND and prints nothing
+# at all: no window, no error, no exit message. Verified on a clean Windows 11
+# 26100 VM against the released v0.24.0 CLI and desktop binaries (#657).
+#
+# CI never caught it because GitHub's windows runners ship the redistributable.
+#
+# The C++ side has to agree: a /MT Rust runtime against a /MD whisper.cpp is a
+# link-time mismatch. The `cc` crate reads crt-static from
+# CARGO_CFG_TARGET_FEATURE and switches to /MT, and `cmake` inherits its flags,
+# so this should propagate to the bundled whisper.cpp build. That propagation
+# is the risky part of this change and is what the Windows CI jobs verify.
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,9 +33,15 @@ jobs:
               - 'crates/cli/**'
               - 'crates/reader/**'
               - 'crates/whisper-guard/**'
+              - 'archive/**'
               - 'Cargo.toml'
               - 'Cargo.lock'
               - 'rust-toolchain.toml'
+              # Sets rustflags and build env per target, so it changes how every
+              # Rust target compiles. Omitting it meant a change that statically
+              # linked the Windows CRT skipped every Rust job and reported green
+              # (#657 follow-up).
+              - '.cargo/**'
               - '.github/workflows/ci.yml'
 
   test:


silverstein:
## This approach does not work. Static CRT is blocked by ONNX Runtime.

Built this branch on a real Windows machine with the full default feature set. It compiles and then fails at link:

```
libort_sys-93d30fc81601fc7e.rlib(unary-elementwise.obj) : error LNK2019:
  unresolved external symbol __imp_cbrtf referenced in function
  "void __cdecl `anonymous namespace'::unary_ukernel_quantized<...>"
...
minutes.exe : fatal error LNK1120: 38 unresolved externals
```

The `__imp_` prefix is the tell: those symbols are expected from a **DLL import**, i.e. the dynamic CRT. `ort-sys` ships **prebuilt** ONNX Runtime binaries linked against the dynamic runtime, so they cannot link against a statically linked Rust side. 38 unresolved symbols, all of this shape.

`ort` arrives through `diarize`, which is a default feature:

```
crates/cli/Cargo.toml: default = ["whisper", "diarize"]
crates/cli/Cargo.toml: diarize = ["minutes-core/diarize", "dep:pyannote-rs", "dep:ort"]
```

So the shipped configuration cannot use `+crt-static` without replacing or rebuilding ONNX Runtime.

## Why CI said this was fine

Two independent reasons, both worth fixing regardless of what happens to this PR.

1. The `rust` path filter did not include `.cargo/**`, so the first push skipped every Rust job and reported 14 green checks. Fixed in this branch.
2. `Install CLI (windows-latest)` runs `cargo install --path crates/cli --no-default-features`. That excludes `diarize`, which excludes `ort`, which excludes the exact dependency that breaks. It is the one Windows configuration that could not observe this failure.

whisper.cpp was never the problem: `cc` propagates `crt-static` to `/MT` correctly and that half worked.

## Where this leaves #657

The bug is real and still unfixed: every published Windows artifact needs the VC++ Redistributable and none ship it, so a clean machine gets `STATUS_DLL_NOT_FOUND` and no message. Confirmed by controlled experiment.

Static linking was the option that kept a single self-contained binary, and it is now off the table unless ONNX Runtime is rebuilt with a static CRT, which is a much larger undertaking than a build flag.

That leaves app-local deployment: ship `vcruntime140.dll`, `vcruntime140_1.dll`, `msvcp140.dll` and `msvcp140_1.dll` beside the executable, which Microsoft permits. It has to be solved in two places, the bare CLI download and the NSIS installer, and it changes the CLI artifact from one file to an archive.

Closing this rather than iterating on a dead approach. Reopening #657 as the tracking issue with the corrected direction.

Worth recording: this failure is invisible to CI by construction, and was only found by building the real feature set on a real Windows machine.

silverstein:
Superseded by the finding above: ONNX Runtime ships prebuilt against the dynamic CRT, so +crt-static cannot link the default feature set. #657 stays open with app-local DLL deployment as the remaining path.

PR #663, opened 2026-08-07
fix(windows): ship the MSVC runtime with the desktop app too (#657)

silverstein:
Completes #657. The CLI half shipped in #660; the desktop app has the identical dependency and was not covered.

## The gap

`minutes-desktop-windows-x64.exe` imports `MSVCP140.dll` and `MSVCP140_1.dll`. Verified on a clean Windows 11 VM against the released artifact:

```
& C:\qa\minutes-app.exe --version
desktop_exit=-1073741515      # 0xC0000135 STATUS_DLL_NOT_FOUND
desktop_output=
```

The installer does not save it. `tauri.windows.conf.json` handles WebView2 via `downloadBootstrapper` and has no equivalent for the C++ runtime, and `installMode: currentUser` means it runs unelevated so it could not install a machine-wide runtime even if it tried. The likely user experience is an install that appears to succeed followed by an app that will not launch, with no message.

## Fix

Tauri places bare-filename `resources` beside the executable in the install directory, which is where the Windows loader looks, so the four DLLs land next to the app. No new download, no extra step, no admin rights, and nothing changes for the user.

CI stages the files from the runner's VS redistributable rather than committing Microsoft binaries to the repo. The resource entries are globs so a checkout without them still builds locally, and the staging step fails the job if fewer than four are found rather than quietly producing an installer that reproduces the bug.

## Verification status

The mechanism is proven: the same four DLLs beside the same released binary took it from `exit=-1073741515` with no output to `exit=0` printing its version, on a machine with no runtime installed.

What is **not** yet verified is that Tauri's resource placement puts them beside the exe in the NSIS layout rather than in a subdirectory. That needs a built installer run on the clean snapshot, and GitHub Actions has been in a major outage. I will confirm it before the next release rather than assume; the CLI half is already merged and verified, so this does not block that fix reaching users.

### DIFF
--- a/.github/workflows/archive-pilot-private.yml
+++ b/.github/workflows/archive-pilot-private.yml
@@ -119,11 +119,13 @@ jobs:
             -p minutes-archive-semantic \
             -p minutes-archive-convert \
             -p minutes-archive-core \
+            -p minutes-archive-ocr \
             -p minutes-archive-app
           cargo clippy \
             -p minutes-archive-semantic \
             -p minutes-archive-convert \
             -p minutes-archive-core \
+            -p minutes-archive-ocr \
             -p minutes-archive-app \
             --all-targets -- -D warnings
 
@@ -150,6 +152,8 @@ jobs:
           set -euo pipefail
           app="target/release/bundle/macos/Minutes Archive.app"
           cargo run -p minutes-archive-core --example document_vault_smoke -- \
+            "$app/Contents/MacOS/$ARCHIVE_EXECUTABLE"
+          cargo run -p minutes-archive-core --example archive_pilot_soak -- \
             "$app/Contents/MacOS/$ARCHIVE_EXECUTABLE"
           scripts/archive-native-lifecycle-smoke.sh "$app"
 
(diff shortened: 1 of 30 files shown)


silverstein:
## Verified: the DLLs land beside the executable

Built the NSIS installer from this branch on the Windows QA VM and silently installed it:

```
head: 135b44c fix(windows): gate runtime staging on the target, not the host
checkout verified: resources entry and staging function both present
tauri build exit=0

staged into the source dir by build.rs:
  msvcp140.dll, msvcp140_1.dll, vcruntime140.dll, vcruntime140_1.dll

silent install exit=0
installed layout:
  minutes-app.exe
  msvcp140.dll
  msvcp140_1.dll
  uninstall.exe
  vcruntime140.dll
  vcruntime140_1.dll
```

Install root, beside the executable, not a `resources\` subdirectory. That is where the Windows loader looks, so this closes the one item this PR listed as unverified.

## A note on how nearly this went wrong

The first two attempts produced misleading results and are worth recording, because both would have read as success.

Run one reported `tauri build exit=0` with no DLLs staged and no glob error, which appeared to contradict the finding that an unmatched glob is fatal. It had silently built `ef9dea4`, a commit from an unrelated branch: the `--depth 1` fetch failed and `FETCH_HEAD` still pointed at the previous fetch. The tree it compiled had no `resources` entry at all.

Run two aborted on purpose. I had added a guard that refuses to build unless the checkout actually contains both the `resources` entry and `stage_msvc_runtime`. It caught a dirty working tree left by run one, where `git checkout` had failed and left the old commit in place.

Only run three, with an explicit `reset --hard` and the guard passing, produced a result worth trusting. A build succeeding is not evidence that it built what you think it did.

PR #665, opened 2026-08-07
Archive pilot: independent review of everything since candidate 4f251483

silverstein:
Independent adversarial security review before this goes to an attorney's
thirty years of privileged client documents. 31 commits, ~5,000 lines.

**What the application is.** A local-only macOS app that indexes documents from
folders the owner approves through the native picker, and answers questions with
exact quotations plus a source anchor. It must never reach the network, never
send a document anywhere, and never let a result read as more complete than it
is. The index is in memory only and is discarded when the window closes.

## Where to look hardest

**Worker execution and identity.** All three workers (converter, OCR, semantic)
execute in place from the bundle and are pinned with an `O_NOFOLLOW` descriptor
plus dev/ino, re-checked before each launch. A previous build copied them out of
the bundle and macOS SIGKILLed the copy, so every notarized build could index
nothing. `digest_file` in archive-convert and archive-semantic now uses
positional reads: `try_clone` is `dup` and shares one file offset, which made
concurrent verification report the executable as swapped.

**The OCR sandbox is weaker than the others, deliberately.** Apple Vision cannot
run under the read-scoping the other two use, so `minutes-archive-ocr` allows
`file-read*` while keeping `(deny default)`, no writes, `(deny network*)`, the
mach-lookup denylist, and RLIMIT_AS/CPU. A worker receiving attacker-controlled
image bytes can therefore read the filesystem. Image decoders are a classic
exploit target. Please weigh this directly rather than accepting the framing:
docs/design/archive-ocr-scope.md has the measurements.

**Claims that must not be overstated.** A transcription from OCR is never
returned as an exact quotation (enforced by an `ExtractedText` witness type). A
PDF or RTF may never answer a same-clause question because it cannot establish
where a clause ends. A partial index, or partial semantic coverage, must say so.

**Degradation paths, all new.** Losing the semantic worker mid-build now costs
suggestions rather than the index; a budget yields a partial index rather than
an error. Check that nothing here loses documents silently, and that every
"degrade instead of fail" path still reports what it dropped.

**Boundary.** The webview receives opaque location ids and never a path, in
either direction — including the new "skip folders" flow, which uses the same
native panel. Exported reports contain no paths, filenames, or text.

**Folder exclusions.** `ExcludedFolder` binds a skipped folder to one approved
root, and the session holds the location's stable id, resolving the index only
at build time. An exclusion that drifted onto the wrong root would silently stop
reading documents an attorney believes are indexed.

**The candidate ceiling.** Raised 2,000 -> 25,000 clauses. Beyond it the search
still refuses rather than truncating, because candidates are fetched as an `OR`
while an answer requires all terms, so a truncated set could hide a real match.
Confirm that reasoning holds.

## Known and deliberate

- No auto-updater. The config previously carried an update key and a GitHub
  endpoint while the footer promises networking is disabled; both removed.
- Handwriting is out of scope for OCR.
- Text-layer-less PDFs are counted, not rasterised and read.

## How it is verified

`scripts/verify-archive-dev-app.sh` runs against the installed artifact,
including `archive_pilot_soak`, which lowers its own RLIMIT_NOFILE to the 256
launchd gives a GUI app and indexes 4,000 documents. Both exist because earlier
failures were reachable only at real scale and under the real limits.

No client material is in this repository or any test run; all fixtures are
synthetic.


silverstein:
Pushed three commits addressing the ultrareview findings on this branch:

- **cf9d1bf3** — `search_documents` checked provenance only after aggregating matched concepts, the exact-phrase claim, and the exclusion flag, so a scan whose only matches were OCR passages could surface as a `DocumentEvidenceCard` claiming matches across zero provisions, and transcription cards on the document-scope path ignored `excluded_concepts`. Provenance now decides first (mirroring `search_same_provision`), the exclusion filter applies to transcription cards, the existing provenance test asserts no document card at all (it previously passed vacuously), and a new test pins the exclusion behaviour with a matching control.
- **cc2af866** — the OCR worker copy embedded in the app binary (the one `BoundedTranscriber::bind(current_exe)` actually runs in production) still read stdin unbounded; it now carries the same `.take(MAX_IMAGE_BYTES + 1)` bound 901217db gave the standalone worker.
- **b18f007b** — removed shadowed duplicate `document_id` keys and a stray field from the UI smoke-test mocks.

Verified: 91 archive-core tests pass (84 lib + integration), `cargo check` on the Tauri app binary, clippy `-D warnings`, `cargo fmt --check`, and `node --check` on the smoke script.

https://claude.ai/code/session_01MjmKq74A9UFXTYAUbYEjbx

PR #680, opened 2026-08-08
fix(search): derive the corpus authorization deadline from its own ceilings (#679)

### LINKED ISSUE #679: Flaky: search corpus reverification exceeds its 15s deadline on loaded CI runners
`search::tests::default_budget_supports_realistic_1500_meeting_multi_pass_list` fails intermittently on hosted runners. It has hit both platforms in the last day, on unrelated PRs, and it is currently the main source of red CI that is not attributable to a change.

```
thread 'search::tests::default_budget_supports_realistic_1500_meeting_multi_pass_list'
panicked at crates/core/src/search.rs:3986:86:
called `Result::unwrap()` on an `Err` value:
  Io(Custom { kind: Other, error: "meeting corpus could not be reverified safely" })
```

## Why it flakes

It is a wall-clock deadline, not a logic failure. `authorize_active_corpus_operation` (`crates/core/src/search.rs:40-77`) runs three full corpus passes under one shared absolute deadline:

1. `stable_active_corpus_revision_with_budget` (before)
2. the operation itself
3. `stable_active_corpus_revision_with_budget` (after) ← where it fails

`ACTIVE_CORPUS_AUTHORIZATION_DEADLINE` is 15s (`crates/core/src/markdown.rs:87`). The fixture writes 1,500 files totalling ~39 MB, so those three passes move well over 100 MB of I/O. On a fast machine that fits comfortably; on a shared runner under load it does not, and the failure lands on the third pass because that is the last one to run.

Note the error text says "reverified", which is the *after* pass specifically, matching that reading.

## What the test is actually for

The name says it: the **default budget** should accommodate a realistic corpus. The budget in question is the count and byte ceilings (`ACTIVE_CORPUS_MAX_FILE_COUNT = 4096`, `ACTIVE_CORPUS_MAX_AUTHORIZED_BYTES = 80 MB`) against a 1,500-file / 39 MB corpus. The 15s wall clock is incidental to that intent, and it is the part that decides the outcome on CI.

## Options

1. Keep the budget assertion, stop letting wall-clock decide it: give the test an explicit generous deadline so it tests the count/byte envelope rather than runner speed.
2. Raise `ACTIVE_CORPUS_AUTHORIZATION_DEADLINE`. This is a real availability/safety property, so it should not move just to quiet a test.
3. Shrink the fixture. Weakens the "realistic corpus" claim the test exists to make.

Option 1 looks right, but it touches corpus authorization, so it deserves a considered change rather than a quick patch.

## Notes

- Not reproducible in my sandbox: the process is killed before completing (SIGTERM at ~1.6s), so timing has to be measured on a real machine or in CI.
- Seen at `fcad33a4` (windows-latest) and on the #678 branch (macos-latest); passed on re-run both times, which is consistent with a load-dependent deadline.

silverstein:
Closes #679.

## This was not a test problem

`search::tests::default_budget_supports_realistic_1500_meeting_multi_pass_list` failed intermittently on both platforms, on PRs that touched nothing nearby. The test was reporting a real inconsistency.

`ACTIVE_CORPUS_AUTHORIZATION_DEADLINE` was a hardcoded 15s. The ceilings it guards authorize far more work than fits in that. One authorization reads the corpus six times per attempt (pre-snapshot, four materialization passes, post-snapshot) and may run two attempts under one shared deadline, so at the documented 80 MB ceiling it authorizes 960 MB:

```
960 MB / 15s = 64 MB/s sustained
```

That is what a corpus at the published limit needed merely to avoid a spurious failure. CI surfaced it first, but the same arithmetic fails a user whose meetings live on iCloud Drive, Dropbox, or an external disk, which this product explicitly supports. The 39 MB fixture needed 15.6 MB/s; locally it finishes in 5.77s of a 15s budget, so 2.6x headroom, which a loaded runner erases.

## The fix

The deadline is derived from the ceilings and an explicit floor throughput the envelope promises to tolerate (16 MB/s), giving 60s. Raise the byte ceiling or add a materialization pass and the deadline follows, instead of silently demanding a faster disk.

A test pins the relationship rather than the number. Confirmed it fails against the shipped value, with the arithmetic spelled out:

```
the deadline demands 67108864 B/s sustained to read the documented ceiling,
above the 16777216 B/s floor this envelope promises to tolerate;
raise the deadline or lower the ceiling (#679)
```

A const assertion rejects a zero-second deadline at compile time, since the derivation is an integer division. Verified by setting the floor to 2 GB/s:

```
error[E0080]: evaluation panicked: corpus authorization deadline truncated to
zero seconds: the assumed floor throughput is at least the worst-case
authorized byte total
```

## Why extending this is safe

This deadline is a cooperative backstop checked between filesystem operations, not the primary control. Its own doc comment says so: the byte, file, and directory ceilings bound the work, and agent-facing SDK/MCP reads wrap this boundary in a supervised, killable worker. Extending it does not widen what any caller may read. The cost is that a genuinely hung mount now errors in 60s rather than 15s; the benefit is that legitimate corpora stop failing.

Sizing is deliberately against what the envelope **authorizes**, not what today's code happens to use. Sizing against observed behavior would re-break the moment a change legitimately used its full materialization budget.

## The error discarded its own cause

Both revision passes mapped every `ActiveCorpusRevisionError` to one opaque string, so a deadline overrun and a genuinely oversized corpus were indistinguishable. Diagnosing this needed a controlled experiment to recover a fact the error already had. The message now names which limit was hit, and never content, a path, or a count:

```
meeting corpus could not be reverified safely: the authorization deadline elapsed
```

The two `canonicalize()` sites that share the old wording are a different error type (the corpus root cannot be opened, with no richer cause to preserve) and are left alone. Nothing in the repo asserts the old strings.

## Verification

- `minutes-core` lib suite: **1661 passed, 0 failed**
- the previously flaky test passes; `search::` 81, `markdown::` 81, `graph::` 78, all green
- `cargo fmt` clean; no new clippy warnings
- the 12 initial failures were `audio_decode_worker`/`diarize`/`health`/`watch` tests needing a `minutes-cli` binary beside the harness, an artifact of a fresh worktree. All 12 pass once it is built, which is the 1661 figure above.

### DIFF
--- a/crates/core/src/markdown.rs
+++ b/crates/core/src/markdown.rs
@@ -97,43 +97,71 @@ pub(crate) const ACTIVE_CORPUS_MAX_MATERIALIZATION_READ_PASSES: usize = 4;
 /// without tripping the deadline.
 pub(crate) const ACTIVE_CORPUS_MIN_ASSUMED_READ_BYTES_PER_SEC: u64 = 16 * 1024 * 1024;
 
+/// Times a snapshot physically reads each file for every byte it charges.
+///
+/// `read_bounded_markdown_twice` reads the file, seeks back to zero, and reads
+/// it again to prove the bytes did not change underneath it, but the budget
+/// charges `content.len()` once. Wall-clock cost therefore tracks twice the
+/// authorized byte count, so a deadline derived from authorized bytes alone
+/// understates the time the same work needs by exactly this factor.
+pub(crate) const ACTIVE_CORPUS_PHYSICAL_READS_PER_SNAPSHOT: u64 = 2;
+
 /// Bytes one authorization may legitimately read at the documented ceiling.
 ///
 /// Each attempt makes a pre-snapshot pass, a materialization phase worth
 /// `ACTIVE_CORPUS_MAX_MATERIALIZATION_READ_PASSES` passes, and a post-snapshot
 /// pass, and the whole thing may run `ACTIVE_CORPUS_MAX_AUTHORIZATION_ATTEMPTS`
 /// times under one shared deadline.
+///
+/// This is the *charged* total, which is what the byte ceiling governs. It is
+/// deliberately not the number of bytes read from the disk: see
+/// `ACTIVE_CORPUS_WORST_CASE_PHYSICAL_BYTES` for that.
 pub(crate) const ACTIVE_CORPUS_WORST_CASE_AUTHORIZED_BYTES: u64 = ACTIVE_CORPUS_MAX_AUTHORIZED_BYTES
     * (1 + ACTIVE_CORPUS_MAX_MATERIALIZATION_READ_PASSES as u64 + 1)
     * ACTIVE_CORPUS_MAX_AUTHORIZATION_ATTEMPTS as u64;
 
+/// Bytes actually pulled off the disk for that same worst case.
+///
+/// The deadline is wall-clock, so it has to be sized against this rather than
+/// the charged total.
+pub(crate) const ACTIVE_CORPUS_WORST_CASE_PHYSICAL_BYTES: u64 =
+    ACTIVE_CORPUS_WORST_CASE_AUTHORIZED_BYTES * ACTIVE_CORPUS_PHYSICAL_READS_PER_SNAPSHOT;
+
 /// Absolute wall-clock ceiling for one corpus authorization.
 ///
-/// Derived, not chosen. A hardcoded 15s silently required 64 MB/s sustained to
-/// read the documented 80 MB ceiling through all six passes of both attempts,
-/// so a corpus that fit every published limit still failed on ordinary storage
-/// (issue #679). It surfaced as an intermittent CI failure, but the same
-/// arithmetic fails a real user whose meetings sit on a synced folder.
+/// Derived, not chosen. A hardcoded 15s silently required 128 MB/s of real
+/// read throughput to get the documented 80 MB ceiling through all six passes
+/// of both attempts, each of which reads every file twice, so a corpus that
+/// fit every published limit still failed on ordinary storage (issue #679). It
+/// surfaced as an intermittent CI failure, but the same arithmetic fails a real
+/// user whose meetings sit on a synced folder.
 ///
-/// Deriving it keeps the two consistent: raise the byte ceiling or add a
-/// materialization pass and the deadline follows, instead of quietly
-/// tightening the throughput this code demands.
+/// Deriving it keeps the two consistent: raise the byte ceiling, add a
+/// materialization pass, or add a verification reread, and the deadline
+/// follows, instead of quietly tightening the throughput this code demands.
 ///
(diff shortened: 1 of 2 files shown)


silverstein:
Adversarial review (codex) found a **2x undercount in my own derivation**. Pushed three commits fixing it.

## The High finding, and it is correct

Every snapshot reads each file twice. `read_bounded_markdown_twice` reads the file, seeks back to zero, and reads it again to prove the bytes did not change underneath it, but the budget charges `content.len()` once. So wall-clock cost tracks twice the charged byte total.

That makes the version I first pushed wrong in exactly the way this PR exists to fix:

| | first push | corrected |
|---|---|---|
| worst case off the disk | 960 MiB (charged) | 1920 MiB (physical) |
| deadline | 60s | 120s |
| real throughput demanded | 32 MB/s | 16 MB/s |

So my documented promise that 16 MB/s storage "must be able to finish the whole documented envelope" was false. Same class of quiet inconsistency as the original bug, one layer down.

The derivation now models charged and physical bytes separately and sizes the deadline against the physical total, rounding up so an inexact ratio cannot promise a floor it does not honor.

My test could not catch it because it repeated the same charged-byte model. It now checks the physical derivation against the per-snapshot reread count and compares rates by cross-multiplication instead of division. Confirmed it fails the 60s version:

```
the deadline of 60s carries only 1006632960 B at the 16777216 B/s floor this
envelope promises to tolerate, short of the 2013265920 B the documented
ceiling can require
```

## Also addressed

**Graph lease window.** This constant bounds `graph`'s corpus capture too, and graph holds a projection lease across it, so a slow disk now refuses a competing projection for longer. That is the intended trade (the alternative is failing a projection the ceilings permit), but it is now written down at the constant rather than left to be discovered.

**The "leaks nothing" claim was wrong.** The cause string reaches CLI and desktop callers, so someone who cannot read restricted meetings can distinguish an unopenable corpus from an oversized one, a walk failure, or an exhausted deadline. Bounded and worth it, but the comment now says what is disclosed instead of claiming nothing is.

**`Traversal` wording.** Every `WalkDir` error maps to that variant, including permission denials and files that vanish mid-walk, so "the corpus contains an unsafe path" was too strong. It now says the corpus could not be walked.

## Findings I did not act on

**The test still uses the real filesystem.** True, and the timing window moved from 15s to 120s rather than being eliminated. Injecting a clock would make it deterministic but would stop it testing the thing that broke, which was the interaction between real read volume and a wall-clock bound. The arithmetic test is the deterministic half; this one stays an integration check with roughly 20x headroom instead of 2.6x.

**The commit's 15.6 MB/s figure for the 1500-meeting fixture.** Correct that the default list path takes two materialization snapshots rather than four, and that with the double read the fixture needed about 20.8 MB/s. That does not change the fix, since the deadline is sized against what the envelope authorizes rather than what one path happens to use.

`minutes-core` lib suite: 1661 passed, 0 failed.

PR #681, opened 2026-08-08
fix(sdk): present one corpus-lease refusal for one condition, and stop tests racing worker startup

silverstein:
Fixes the `corpus-lease.test.ts` flakes that have been costing a CI re-run each time. Two separate causes: one is a product defect, one is a test defect.

## 1. Product defect: two errors for one condition (the Windows failure)

The Windows CI failure was not a timeout being too tight. It was `withStableCorpusLease` reporting a different error depending only on timing:

```
Expected: "stable meeting corpus authorization failed"
Received: "meeting corpus authorization deadline elapsed"
```

`fail()` wraps refusals raised once the worker is running, but a `CorpusLeaseChangedError` thrown before that escapes raw. The dispatch computes the deadline and checks it on the very next statement:

```ts
const deadline = authorizationDeadline(hooks.timeoutMs);
const timeoutMs = remainingAuthorizationMs(deadline);   // throws raw
```

On a contended runner the process can be descheduled between those two lines, so a short budget elapses in the gap. This reaches CLI, MCP, and desktop callers, not just tests: the same failure has two different user-visible messages.

The public entry now maps every `CorpusLeaseChangedError` to the documented refusal. `CorpusLeaseBudgetError` and the plain `Error` refusals keep their own text, since those name a caller-actionable limit rather than an authorization outcome, and four tests assert them specifically.

Reproduced deterministically with the smallest accepted budget. Without the wrapper the new test fails with the exact CI text:

```
expected ... 'stable meeting corpus authorization f…'
but got 'meeting corpus authorization deadline…'
```

## 2. Test defect: asserting a positive count against a wall clock (the macOS failure)

`never publishes a projection when final authorization stalls` asserts `projections === 1` while giving the whole lease 500ms. That budget has to cover spawning a worker, arming a recursive watcher, and computing a baseline manifest before the projection runs at all. When it does not, the lease fails first and the count is 0, which is exactly what CI reported.

Measured on this machine:

| | time to reach the projection |
|---|---|
| unloaded | p50 81ms, max 87ms |
| 1.5x CPU oversubscription | p50 92ms, p90 113ms, max 132ms |

So 500ms held here with ~4x margin and did not on a loaded macOS runner.

It now ends the operation from inside the projection using `operationDeadlineForTest`, the seam already in this file for precisely this purpose, so the deadline fires at a point the test controls rather than whenever the runner gets there. It is also faster: 119ms instead of a 500ms wall-clock wait.

I hardened one latent case of the same shape, `uses one cumulative deadline across hooks and retry attempts`, which asserts `baselineCalls === 1` against a 500ms budget. Its hook still outlasts the budget, which is the property under test.

## Verification

- **0 failures in 8 full-file runs under 1.5x CPU oversubscription** (the loaded condition that reproduces CI)
- each fix confirmed to fail without its change
- SDK suite: 146 passed, 1 skipped
- MCP suite: 280 passed, 1 skipped

Before this, the same file was 0/12 in isolation locally, which is why it only ever showed up on CI.

### DIFF
--- a/crates/mcp/src/corpus-lease.test.ts
+++ b/crates/mcp/src/corpus-lease.test.ts
@@ -682,7 +682,16 @@ describe("stable corpus lease", () => {
           }
         )
       ).rejects.toThrow("stable meeting corpus authorization failed");
+      // The rejection is the non-publication proof: a published projection
+      // would have resolved to MUST_NOT_PUBLISH instead.
       expect(projections).toBe(1);
+      // And the stalled worker has to be reaped, not merely abandoned. Without
+      // this the deterministic deadline could satisfy the assertions above
+      // while leaving a wedged child holding admission, which is the failure
+      // this scenario exists to catch.
+      await expect(
+        withStableCorpusLease(root, () => "reused")
+      ).resolves.toBe("reused");
     });
   });
 
--- a/crates/mcp/src/corpus-lease.test.ts
+++ b/crates/mcp/src/corpus-lease.test.ts
@@ -585,10 +585,16 @@ describe("stable corpus lease", () => {
             operationCalls += 1;
           },
           {
-            timeoutMs: 500,
+            // `baselineCalls` is asserted to be 1, so the budget has to
+            // survive worker startup before afterBaseline is even reached, or
+            // the lease fails first and the count is zero. Same shape as the
+            // stall test below, which did exactly that on a loaded runner.
+            // The hook still outlasts the budget, which is the property under
+            // test; the budget just no longer races startup.
+            timeoutMs: 2_000,
             afterBaseline: () => {
               baselineCalls += 1;
-              return new Promise((resolve) => setTimeout(resolve, 750));
+              return new Promise((resolve) => setTimeout(resolve, 3_000));
             },
           }
         )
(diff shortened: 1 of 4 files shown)


silverstein:
Adversarial review (codex). The product fix survives; two of my own claims did not, and I have corrected them.

## The wrapper survives

Codex enumerated **37 distinct `CorpusLeaseChangedError` messages** (23 static plus 14 worker-side reasons) and confirmed nothing depends on telling them apart: the class is private and unexported, no `instanceof` check exists outside the module, `reader.ts` and the MCP helpers just await the lease, and the one internal retry loop already catches every body error without inspecting type or message. The `beforeSentinelCreate` in-process path is wrapped correctly too.

## Corrections to my own claims

**My PR description was wrong about the blast radius.** I wrote that the inconsistent error "reaches CLI, MCP, and desktop callers". It does not: the CLI and Tauri app depend on Rust `minutes-core`, not this TypeScript module. The real reach is MCP and SDK consumers. The code comment did not repeat the error, but the claim here was unsupported and I should not have made it.

**The regression test named a route it does not reliably take.** I called it "the budget is gone before the worker starts". Codex measured the synchronous spawn on this host at 1.06 to 15.7ms, so with a 1ms budget the guard that actually fires is most likely the one *after* spawn. Worse, one of the three possible routes reaches the refusal through `fail()`, which would pass even without the wrapper.

The test now pins the contract instead: an exhausted budget can be noticed by several guards, and every guard must produce the documented refusal. Which one fires is timing-dependent and deliberately not asserted. Claiming an unverified specific was exactly the kind of overclaim this fix exists to remove.

It does still pin the fix here: removing the wrapper makes it fail with the exact CI text.

## The stall test

Codex is right that firing the deadline from inside the projection can pre-empt the worker reaching its stall, so the test no longer proves the timing of final-authorization stalling specifically. I had already added an assertion that a follow-up lease on the same root succeeds, which proves the stalled worker is reaped rather than left holding admission, and that is the property the scenario cares about. The stall also still guarantees authorization can never complete, so the refusal is not incidental.

I judged the remaining fidelity loss worth it: the alternative is a test that only passes when the runner is fast enough, which is what we are removing.

## Not addressed here

Codex found that `uses one cumulative deadline across hooks and retry attempts` does not actually exercise a retry, and that roughly eight other short-budget tests in this file can pass for a startup timeout rather than their named cause. Both are pre-existing and neither is currently failing CI. My change preserves the old causality in the first and does not touch the others. That is a worthwhile follow-up but a larger one than fixing the two flakes that were costing re-runs.

`corpus-lease` suite: 49 passed. Mirror check: 4 files identical across mcp and sdk.

PR #682, opened 2026-08-08
test(sdk): make every corpus-lease refusal test prove its named cause

silverstein:
Follow-up to #681, closing the rest of what adversarial review found in that file.

## The problem

Most refusal tests here asserted only that the lease refused. Since #681 every refusal produces the same message, so a budget consumed by worker startup satisfied them without the named cause ever happening. Two were outright false passes; the rest were startup-sensitive.

## The audit

Starve every lease in the file to a 1ms budget so no test can reach its cause, then see what still passes.

| | passing while starved |
|---|---|
| before | 39 of 49 |
| after | no refusal-shaped test |

That is the property being bought: these tests can no longer be satisfied by a timeout alone. 13 now fail under starvation, up from 10.

## What each test gained

**`aborts an asynchronous operation at the cumulative authorization deadline`** now asserts the operation started and its abort listener fired. Previously a startup timeout satisfied both the rejection and the `< 1000ms` bound without cancelling anything.

**`kills a worker stalled before baseline publication`** now asserts most of the budget was spent. A lease that never got a worker running refuses almost immediately, so elapsed time is the evidence that a worker was alive and stalling. The old 300ms budget made the refusal equally consistent with a worker that never started.

**The three hostile-protocol tests** (out-of-order, oversized line, coalesced transitions) now have fixtures that record that they ran, so the marker proves the hostile protocol was actually produced. A fast refusal alone does not prove it, because a starved budget also refuses fast; I tried that first and the starvation audit caught it. The marker is written *before* emitting, because the parent detects the violation on the line and kills the child, which can beat a post-write marker.

**`fails closed when the sentinel fence is not observed`** counts the watcher-ready hook, which is where the suppression is applied. Timing cannot separate the causes here: a suppressed fence is retried, so the lease runs on to the deadline rather than ending at the fence timeout. I tried the timing approach and measured it going to the full 8s budget, not the 5s fence timeout, so the hook count is the honest evidence.

**`uses one cumulative deadline across hooks and retry attempts`** is renamed to `charges hook time against the one cumulative deadline`. It never exercised a retry: the parent kills the worker while it is still awaiting the hook, so the phase-result that would let the worker advance into a retry is never sent. The test is worth keeping; the retry in its name was not supported by its assertions.

**`accepts a bounded retry`, `pulse failure`, `replacement safety`** are de-raced only. Each already asserts positive phase counts or filesystem effects, so a starved budget fails them loudly rather than passing.

## Verification

- 49 passed
- **0 failures in 6 full-file runs under 1.5x CPU oversubscription**
- starvation audit above, run before and after
- SDK 146 passed; MCP 281 passed; mirror check clean

File runtime goes from about 25s to 30s. That is the cost of budgets that no longer race worker startup, and it buys tests that fail for the reason they name.

### DIFF
--- a/crates/mcp/src/corpus-lease.test.ts
+++ b/crates/mcp/src/corpus-lease.test.ts
@@ -5,6 +5,7 @@ import {
   linkSync,
   mkdtempSync,
   readFileSync,
+  existsSync,
   readdirSync,
   realpathSync,
   renameSync,
@@ -573,7 +574,12 @@ describe("stable corpus lease", () => {
     }
   }, 30_000);
 
-  it("uses one cumulative deadline across hooks and retry attempts", async () => {
+  // Not "across retry attempts": the parent's cumulative timer kills the
+  // worker while it is still awaiting this hook, so the phase-result that
+  // would let the worker advance into a retry is never sent. The test proves
+  // the hook shares the operation's budget, which is worth having, but naming
+  // a retry it never reaches would be a claim the assertions do not support.
+  it("charges hook time against the one cumulative deadline", async () => {
     await withCorpus(async (root) => {
       writeFileSync(join(root, "meeting.md"), "deadline canary");
       let operationCalls = 0;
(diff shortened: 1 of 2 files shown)


silverstein:
Adversarial review (codex) blocked this, correctly. Three High findings, all substantiated, all addressed.

## My evidence proved less than my comments claimed

**Sentinels don't prove the stall ran.** They are acquired before the watcher is registered, and the injected `before-baseline` stall is later still. So their presence bounds the refusal to "the worker got at least as far as sentinel acquisition" rather than pinpointing the stall. That still rules out the failure mode being fixed, where a startup-consumed budget refused with no worker progress at all, but it is not what I wrote.

**The protocol markers prove the fixture reached the pre-write point, not that bytes were emitted and classified.** The child can die between the marker and the write, and every refusal now reads the same. Recording after the write would prove more, but that is exactly the race I hit and backed out of: the parent kills the child on the offending line.

**The fence hook count doesn't prove suppression took effect.** The control only queues a command, which the parent sends after the hook returns and the worker applies later still. It proves the worker reached watcher-ready.

All three comments now state what they actually establish.

## The blanket claim was false

Codex found `fails closed after watcher failure on both attempts`, which I never touched, asserting the shared refusal and nothing else. So "every refusal test proves its named cause" was wrong as written. That test now counts its hook too, and the claim is scoped to what the assertions support.

## One thing I could actually strengthen

The renamed cumulative-deadline test did not distinguish cumulative from per-phase timing: a fresh 2s timeout would reject a 3s hook exactly as a shared deadline does. It now uses two hooks of 1.3s each, both individually inside the 2s budget and together over it. A per-phase timeout admits both and lets the lease proceed; only a shared deadline refuses. Both hook counts and the operation count are asserted, so the test fails if any phase is skipped.

## Margin

Protocol budgets 10s to 6s. Codex was right that at 10s a hang regression plus worker cleanup plus the recovery lease left under 1.5x margin against vitest's 15s testTimeout, which would replace a useful assertion failure with a bare timeout.

## Where the starvation audit stands

Codex is right that it is not a universal oracle: it does not touch leases using the default timeout, it cannot starve the in-process `beforeSentinelCreate` path, and some refusal tests should legitimately keep passing under it because their cause is synchronous validation. I am not claiming it as a proof of correctness. It is a screen for one specific defect class, startup-sensitive false passes, and it found real ones both times I ran it, including in my own first attempt at the protocol tests.

## What remains

Fully pinpointing which refusal fired would need the module to distinguish its refusals to callers, which #681 deliberately stopped doing. Bounding them is what is available without reopening that decision.

49 passed; 0 failures in 5 loaded full-file runs; mirror clean.

Issue #685, opened 2026-08-09
Isolate the sherpa stack behind its own namespace so engine-sherpa and diarize can share a macOS binary


silverstein:
Successor to #683, which is being closed by the compile guard that refuses the broken combination. This issue tracks the real fix: making the combination work.

## Why in-process linking cannot work today

sherpa-onnx's static macOS bundle and pyannote's dependency tree each vendor two of the same native libraries:

| library | sherpa's copy | pyannote's copy |
|---|---|---|
| ONNX Runtime | 1.17.1 (bundled in the static tarball) | 1.22.0 (ort-sys 2.0.0-rc.10) |
| kaldi-native-fbank | `libkaldi-native-fbank-core.a` from sherpa-onnx v1.12.9 | same archive name, vendored by knf-rs-sys 0.3.2 |

A Mach-O image keeps one definition per symbol, so whichever copy wins the link serves both consumers. Measured on macOS arm64 (iMac, commit be2cad55), every possible resolution breaks one side:

| ORT winner | kaldi winner | pyannote | sherpa |
|---|---|---|---|
| 1.17.1 | sherpa's | ✗ `API version [22] is not available` | ✓ |
| 1.22.0 | sherpa's | ✗ SIGABRT entering feature extraction | ✓ |
| 1.22.0 | knf-rs's | ✓ enrollment completes | ✗ SIGABRT |

Row 1 is the #683 report. Rows 2 and 3 come from substituting archives via `SHERPA_LIB_PATH`: ORT can be unified upward because its C API is backward compatible (sherpa transcription runs fine against 1.22.0), but kaldi-native-fbank has no ABI contract. Identical mangled names over divergent struct layouts are undefined behavior; the SIGABRTs we observed are the lucky failure mode.

Linux and Windows do not have this problem because sherpa links dynamically there and its shared library resolves against its own bundled dependencies, not the executable's.

## Candidate designs

**A. dlopen plugin (preferred).** A new cdylib crate (`minutes-sherpa-plugin`) statically embeds the whole sherpa stack, including its own ORT and kaldi, and exports a minimal C API (create engine, transcribe samples, destroy). `sherpa_engine.rs` grows a `libloading` backend that dlopens it. macOS two-level namespaces bind the dylib's internals to its own copies, so the main binary keeps its pyannote stack untouched.

- No IPC and no audio crossing a process boundary, so the secure-byte-transport requirement that parked the parakeet sidecar (docs/architecture/parakeet.md) does not apply.
[...]

PR #686, opened 2026-08-09
fix(build): refuse the sherpa+diarize combination on macOS (#683)

### LINKED ISSUE #683: macOS: engine-sherpa breaks pyannote-rs ORT initialization
## Environment

* macOS Apple Silicon (`arm64`)
* Minutes `0.24.0` @ `be2cad55`
* Build: default features + `engine-sherpa,metal`
* `vad_engine = "whisper-silero"`
* `ORT_DYLIB_PATH` unset
* Homebrew `onnxruntime` not installed

## Problem

Sherpa transcription works, but `minutes enroll` fails when `pyannote-rs` initializes:

```text
running diarization engine=pyannote-rs

The requested API version [22] is not available,
only API versions [1, 17] are supported in this build.
Current ORT Version is: 1.17.1

Failed to initialize ORT API
```

## A/B reproduction

With Sherpa:

```bash
cargo build --release -p minutes-cli \
  --features "engine-sherpa,metal"

./target/release/minutes enroll
```

→ `pyannote-rs` fails with ORT API 17/22 mismatch.

Same commit, without Sherpa:

```bash
CARGO_TARGET_DIR=target-pyannote \
cargo build --release -p minutes-cli \
  --features metal

target-pyannote/release/minutes enroll
```

→ succeeds:

```text
Session successfully initialized.
speaker clustering complete
diarization complete
✓ Voice profile saved!
```

Sherpa transcription itself also works correctly in the first build.

This looks related to #408/#369, but this is **not the ort-silero VAD path**: `vad_engine` is `whisper-silero`; the failure occurs when `pyannote-rs` initializes.

Expected: `engine-sherpa` should coexist with the default `diarize` feature.

silverstein:
Fixes #683. The isolation work that makes the combination actually work is #685; this PR stops the broken combination from shipping or silently corrupting anyone's build in the meantime.

## The defect, fully reproduced

traoredev's A/B repro held up exactly on an arm64 iMac at `be2cad55`: with `engine-sherpa` compiled in, `minutes enroll` fails at pyannote's ORT init with the API 17/22 mismatch; the identical commit without sherpa enrolls fine.

The root cause is two pairs of vendored native libraries fighting over one symbol namespace:

| library | sherpa's copy | pyannote's copy |
|---|---|---|
| ONNX Runtime | 1.17.1 | 1.22.0 (ort-sys rc.10, requests API 22) |
| kaldi-native-fbank | `libkaldi-native-fbank-core.a` (sherpa-onnx v1.12.9) | same archive name (knf-rs-sys 0.3.2) |

I tested every resolution by substituting archives via `SHERPA_LIB_PATH`:

| ORT winner | kaldi winner | pyannote | sherpa |
|---|---|---|---|
| 1.17.1 | sherpa's | ✗ API 17/22 error | ✓ |
| 1.22.0 | sherpa's | ✗ SIGABRT in feature extraction | ✓ |
| 1.22.0 | knf-rs's | ✓ | ✗ SIGABRT |

ORT unifies upward fine (its C API is backward compatible; sherpa transcription runs on 1.22.0). kaldi-native-fbank has no ABI contract, so there is **no link order that works**, and the near-miss failure mode is undefined behavior rather than a clean error.

## What this PR does

- **Compile guard** in `crates/core/src/lib.rs`: `engine-sherpa` + `diarize` on macOS refuses with the working invocation in the error text. macOS-only by construction; Linux/Windows link sherpa dynamically and resolve against sherpa's own bundled libraries.
- **Release builds** drop `engine-sherpa` again: `release-cli.yml` (macOS CLI) and `release-macos.yml` (CLI and the Tauri app, which hard-enables `diarize`). Without this the next release would have shipped broken voice enrollment in both artifacts. v0.24.0 shipped without sherpa and is unaffected.
- **CI asserts the guard fires** on macOS, so a refactor cannot silently drop the cfg and re-open the issue.
- README and docs/architecture/sherpa-engine.md document the constraint and the sherpa-without-diarize build.

## Verified on real hardware (arm64 iMac)

- guard fires with `--features diarize,engine-sherpa`: compile error with the message
- documented escape hatch builds: `--no-default-features --features whisper,engine-sherpa,metal`
- Linux: `--features diarize,engine-sherpa` still checks clean (guard correctly scoped)
- both broken directions and the working directions in the matrix above were each observed, not inferred

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,16 +173,30 @@ jobs:
         # kaldi-native-fbank; the compile guard in crates/core/src/lib.rs is
         # what keeps it out of release builds. Assert it actually fires, so a
         # refactor cannot silently drop the cfg and re-open #683.
+        # The oracle greps for the guard's own message. Merely asserting a
+        # nonzero exit would pass on any unrelated failure (network error in a
+        # build script, toolchain breakage) while proving nothing about the
+        # guard.
         run: |
-          if cargo check -p minutes-core --no-default-features --features diarize,engine-sherpa 2>/dev/null; then
-            echo "::error::the #683 compile guard did not fire for diarize+engine-sherpa on macOS"
-            exit 1
-          fi
-          if cargo check -p minutes-core --no-default-features --features vad-ort,engine-sherpa 2>/dev/null; then
-            echo "::error::the #683 compile guard did not fire for vad-ort+engine-sherpa on macOS"
-            exit 1
-          fi
-          echo "guard fired as expected for both combinations"
+          check_guard() {
+            # bash -e: a failing command substitution in a plain assignment
+            # would abort the step before the oracle runs, so capture the
+            # status in the same statement.
+            status=0
+            out=$(cargo check -p minutes-core --no-default-features --features "$1" 2>&1) || status=$?
+            if [ $status -eq 0 ]; then
+              echo "::error::the #683 compile guard did not fire for $1 on macOS"
+              return 1
+            fi
+            if ! printf '%s' "$out" | grep -q "cannot share a macOS binary"; then
+              echo "::error::$1 failed for a reason other than the #683 guard:"
+              printf '%s\n' "$out" | tail -20
+              return 1
+            fi
+            echo "guard fired for $1"
+          }
+          check_guard diarize,engine-sherpa
+          check_guard vad-ort,engine-sherpa
 
       - name: Build full workspace (macOS only)
         if: runner.os == 'macOS'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,7 +178,11 @@ jobs:
             echo "::error::the #683 compile guard did not fire for diarize+engine-sherpa on macOS"
             exit 1
           fi
-          echo "guard fired as expected"
+          if cargo check -p minutes-core --no-default-features --features vad-ort,engine-sherpa 2>/dev/null; then
+            echo "::error::the #683 compile guard did not fire for vad-ort+engine-sherpa on macOS"
+            exit 1
+          fi
+          echo "guard fired as expected for both combinations"
 
       - name: Build full workspace (macOS only)
         if: runner.os == 'macOS'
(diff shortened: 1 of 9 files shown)


silverstein:
Adversarial review (codex) ran before merge this time. Six findings, all addressed; two were independently caught and fixed while the review was running.

**High, the guard undercounted.** `vad-ort` also pulls `dep:ort`, so `engine-sherpa + vad-ort` links the same two static ONNX Runtimes with no pyannote involved, and `scripts/build.sh` documented exactly that combination as the example app build. The guard now covers `any(diarize, vad-ort)`, both script examples are updated, and both combinations are verified firing on the arm64 iMac.

**Medium, two more forbidden commands in the sherpa doc.** The architecture guide's later sections still said `--features engine-sherpa`, which inherits the default `diarize` and now refuses. Both rewritten, and the prose no longer implies the default set is Whisper alone.

**Medium, my CI oracle was vacuous.** Asserting a nonzero exit accepts any unrelated failure as proof the guard fired. It now greps for the guard's own message and reports the real output when the failure is something else. Also fixed the `bash -e` trap where a failing command substitution in an assignment would abort the step before the oracle ran. This is the same vacuous-check class review has caught in my work before; this time it was caught before merge.

**Low, the Linux claim outran the evidence.** Rather than soften it, I executed it: on x86_64 Linux with both stacks in one binary, enrollment saves a voice profile through pyannote's static ort 1.22.0 while `libsherpa-onnx-c-api.so` (bundling ORT 1.17.1) is loaded, and sherpa transcribes in the same process. The comment now states Linux as verified and Windows as structurally expected but not executed.

**Also from this round:** the graph-worker packaging gate correctly tripped on the `release-macos.yml`/`build.sh`/`install-dev-app.sh` edits; reviewed (feature lists only, nothing near signing or XPC) and goldens updated.

Claims codex confirmed: feature unification routes Tauri- and CLI-originated `diarize` into the guard; the release workflow edits leave `run:` keys attached and no operational macOS `engine-sherpa` references behind; the escape hatch loses exactly diarization and nothing else from the defaults.

Issue #689, opened 2026-08-09
Corpus worker: an early-deadline kill of a never-fed worker permanently poisons the process


silverstein:
Found while chasing the corpus-lease CI cascade in #684's run. The fail-closed design is doing its job, but it fails closed harder than its own threat model requires, and the consequence lands on real MCP servers, not just tests.

## Behavior today

When a lease refusal kills the corpus worker and the child's termination is not confirmed within `CORPUS_WORKER_TERMINATION_GRACE_MS`, the parent deliberately:

- keeps the process-global retained-memory reservation (`releaseReservation = false`)
- poisons the worker path (`corpusWorkerPoisoned = true`), so every later lease in the process throws `Access denied: meeting corpus worker requires a process restart`

That is correct when the child may hold corpus bytes: an unconfirmed zombie could still be streaming snapshot content, and forgetting its reservation would undercount retained memory.

## The over-reach

A worker that was killed **before the parent ever wrote `begin` to its stdin** provably holds no corpus bytes: the protocol is the only way the child learns the corpus root, and it does nothing before `begin`. Killing such a child and missing the grace window (slow machine, spawn still in flight, antivirus scanning node.exe) poisons the process anyway.

On a user's machine that means: one busy moment during a lease that timed out before doing anything, and every corpus read in that MCP server fails until restart. The reservation is also retained forever, shrinking the budget for the rest of the process lifetime even if the poison flag were survivable.

## Proposed change

Track whether `begin` was ever written to the child. In the unconfirmed-termination path, when it was not:

- still confirm/kill as today
- release the reservation and skip poisoning, since the child cannot hold corpus bytes it was never told how to find

Everything after `begin` keeps today's fail-closed behavior unchanged.

## Care needed

- The write-to-stdin ordering must be the authority ("we never sent begin"), not any state echoed by the child.
- The pipe teardown should be sequenced so a straggler child write cannot reach a live data handler after the decision.
- This module is mirrored (`crates/mcp/src/corpus-lease.ts` must stay byte-identical) and sits on the privacy epic's fail-closed reasoning, so the change deserves its own adversarial review, plus verification on the Windows VM where the slow-spawn window actually reproduces.

## Related

- #684's run shows the cascade signature (later tests failing with "retained snapshots exceeded their process budget").
- The test-side containment (reset seam scoped to the aggressive 1ms test) is separate and does not change product behavior.

PR #690, opened 2026-08-09
fix(sdk): contain the refusal test's worker-kill fallout, prove it deterministically

silverstein:
Fixes the corpus-lease cascade first seen in #684's CI run, where later tests failed with `retained snapshots exceeded` after the 1ms-budget refusal test killed its worker mid-spawn. Companion product issue: #689.

## The mechanism

The 1ms test from #682 kills a worker that may still be spawning. When the kill misses the 2s termination-grace window, the module deliberately retains the process-global memory reservation and poisons the worker path. Correct fail-closed product behavior; process-wide cascade in a shared test file.

## Three designs, two rejected on evidence

**1. In-process reroute** (the obvious deflake): verifiably wrong. The in-process dispatcher wraps its own errors before the public entry sees them, so the rerouted test kept passing with the #681 fix removed. Caught by re-running the negative direction. The worker route is load-bearing.

**2. Test-only reset/simulate seams in the production module**: built, then adversarially rejected (codex, review comment below). The reset seam shipped in published runtime modules and, deep-imported, could disarm the fail-closed restart barrier in a real process; the unconditional counter zero could undercount a concurrent live lease; and the simulation installed neither the exact reservation nor the retained admission slot, so its "deterministic proof" proved only a boolean flip. The review also caught two machine-absolute `node_modules` symlinks accidentally committed (gitignore's `node_modules/` matches directories, not symlinks).

**3. File isolation** (this PR): vitest runs each test file in its own process (default forks pool), so the aggressive test moves verbatim to `corpus-lease-refusal.test.ts` and whatever state it leaves behind dies with its own process. No production code changes at all. The branch was rebuilt from main, so the rejected seams and the symlinks never existed in this history.

## What's in the diff

- the 1ms test moves, byte-for-byte, to its own file with a header explaining why it must stay alone there
- the new file joins the corpus-lease mirror set (sdk/mcp byte-identical, now 5 files)
- `.gitignore` gains a bare `node_modules` entry so a symlink can never be committed again

## Verification

- SDK: 146 passed across 4 files; MCP: 281 passed; mirror check clean
- negative direction: with the #681 wrapper removed, the relocated test still fails with the exact raw CI message
- 0 failures in 5 two-file runs under CPU oversubscription
- Windows VM verification was attempted and rejected as evidence: the corpus suite fails 28/49 there at default budgets (126s file runtime) because VirtualBox on an Intel Mac host is far outside windows-latest's envelope. Reported rather than laundered.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,10 @@ __pycache__/
 *.pyc
 docs/marketing/
 docs/demo/node_modules/
+# Also as a bare name: a symlink named node_modules is not a directory, so the
+# trailing-slash pattern above does not ignore it, and `git add -A` in a
+# worktree using linked deps once committed two machine-absolute symlinks.
+node_modules
 docs/demo/.remotion/
 
 # Build artifacts
--- a/crates/mcp/src/corpus-lease-refusal.test.ts
+++ b/crates/mcp/src/corpus-lease-refusal.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from "vitest";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { withStableCorpusLease } from "./corpus-lease.js";
+import { retireBoundReadersForProcessShutdown } from "./secure-read.js";
+
+// This test lives in its own file on purpose, and must stay alone here.
+//
+// A one-millisecond budget kills the corpus worker while it may still be
+// spawning. When the kill misses the termination-grace window, the module
+// deliberately retains the process-global memory reservation and poisons the
+// worker path; that fail-closed design is correct product behavior, but it is
+// process-wide, so in a shared test file it cascades into every later corpus
+// test as "retained snapshots exceeded" failures. Observed on windows-latest
+// in CI.
+//
+// Vitest runs each test file in its own isolated process (the default forks
+// pool), so whatever state this test leaves behind dies with this process and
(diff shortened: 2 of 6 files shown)


silverstein:
Adversarial review (codex) blocked the previous design, and every finding was right. Rather than patch it, the branch was rebuilt from main with a structurally different fix; the rejected design never existed in this history.

What the review caught, and what happened to each:

**P1, the reset seam was a production hazard.** `resetCorpusWorkerStateForTest` shipped in `dist/corpus-lease.js` of the published packages, and a deep import could disarm the fail-closed restart barrier on a live process while the unconfirmed worker might still be alive, with one admission slot permanently charged. Gone: the new design adds no production exports, and no production code changes at all.

**P1, the unconditional counter zero corrupted concurrent accounting.** The lease-A/reset/lease-C interleaving undercounts a live lease despite the clamp. Gone with the seam.

**P2, the simulation was not the real state.** It clobbered the counter to MAX instead of retaining the lease's exact charge, omitted the admission slot, and the test never even observed the reservation because the poison check fires first, so the "deterministic proof" proved a boolean flip. Dropped entirely; the honest containment is process isolation, which needs no simulation.

**P2, two machine-absolute node_modules symlinks were committed.** Root cause: gitignore's `node_modules/` matches directories only, and a symlink is a file. History rebuilt without them, and `.gitignore` now carries the bare-name entry so it cannot recur.

The one thing kept from the blocked design is the negative-direction evidence: the relocated test still fails with the exact raw CI message when the #681 wrapper is removed, so the regression coverage survives the move.

Issue #169, opened 2026-04-22
Native call capture: multi-party meetings only cluster local vs remote; system stem needs ML diarization


silverstein:
## Summary

Native call capture currently behaves like a **local speaker vs remote audio** attribution path, not a true **multi-party remote speaker diarization** path.

That is acceptable for 1:1 calls and some self/remote separation cases, but it breaks down on multi-party Zoom / Meet calls where the system stem contains several remote participants. In practice, these runs can produce results like `raw_diarization_num_speakers: 2` on a 7-person meeting, with most remote speech effectively collapsed into one bucket.

## Context

This was surfaced by a follow-up comment on closed issue #157 after upgrading to `0.13.3`.

This is **not the same problem** as #159:
- #159 is about `[recording.sources] call = "auto"` preflight blocking native ScreenCaptureKit capture before it starts.
- This issue is about the diarization behavior **after native capture succeeds** and emits `.mov` + per-source stems.

## Why I think this is a real product limitation

The native macOS helper already writes:
- `...call.voice.wav`
- `...call.system.wav`

The pipeline then discovers those stems and prefers stem-based attribution.

But the current stem diarizer in `crates/core/src/diarize.rs` is effectively hardcoded to two logical labels:
- `SPEAKER_0` = local / voice stem
- `SPEAKER_1` = call / system stem

It compares energy between the two stems per time window and decides which side "owns" the window. That means it can distinguish:
- local only
- remote only
- overlap with one side dominant

…but it does **not** cluster multiple remote speakers inside the system stem.

So for a multi-party remote call, `raw_diarization_num_speakers: 2` is consistent with the current implementation, even when capture itself succeeded.

## Current user-visible consequence

On native call recordings of multi-party meetings:
- local speaker attribution can still be decent
- remote participants may all collapse into one anonymous remote bucket
- downstream name mapping then has very little useful structure to work with
- overall diarization quality is materially worse than expected for meetings with many remote speakers

## Proposed direction

For native / source-aware call captures, use a hybrid approach:

1. Keep the **voice stem** as the source of truth for the local speaker.
2. Run **ML diarization on the system stem** to cluster remote participants.
3. Remap the system-stem labels so they do not collide with the local label namespace.
4. Merge the local and remote segment timelines into a single diarization result.
5. Preserve the existing stem-only energy heuristic as a fallback for simple / 1:1 / degraded cases.

[...]

PR #713, opened 2026-08-10
fix(attribution): decline to name a voice-stem label that holds two people (#169)

silverstein:
Partial fix for #169, from akshay7394's hybrid-meeting report. It does not fix the diarization gap; it stops Minutes from putting a confident name on the merge that gap produces.

## The defect

Two people sharing one room and one mic, plus one remote participant. Both in-room speakers collapse into `SPEAKER_0`, and self-attribution then names that label after the recorder at medium confidence. The other person's speech is attributed to the recorder, and derived action items land on the wrong human.

The root assumption: a source-backed label says which **stem** speech came from, never how many people were in front of that microphone. The code treated those as the same fact. This is precisely what the project's own rule exists to prevent, that a wrong name is worse than an anonymous label.

## The fix, using evidence already in hand

That branch already diarizes the voice stem to compare embeddings against enrollments, then kept only the boolean match result. It now also counts speakers in that stem and declines deterministic self-attribution when there is more than one.

**Checked before the enrollment comparison, deliberately.** A match does not make naming safe: it proves the recorder is one of the voices in the stem, while the label still covers all of them. Declining leaves the label to the capped, visibly hedged suggestion paths.

**A stricter bar than the existing helper.** `meaningful_speaker_count_excluding` counts anything over half a second, which is right for describing a result and wrong for gating a safety decision. A speaker now counts only with at least 5 seconds AND at least 10% of the stem's attributed speech. Two colleagues with roughly even talk time over 17 minutes clear both by a wide margin; a cough or one mis-clustered noisy window clears neither. Tests pin both directions, because the regression risk here is real: an over-eager guard would silently stop naming solo recorders.

## Diagnosability, which is why this took a code trace

Diagnosing the report required reading the source to learn two things that should have been in the output: `applied_via: source-backed-stem` covers both "matching ran and missed" and "matching never ran" (only distinguishable via a separate `skipped_reason`), and nothing anywhere reported how many people the local stem held. `voice_stem_speaker_count` is now recorded whether or not it changes the decision.

## What stays open

The local stem is still never clustered. In `diarize.rs`, the hybrid merge that splices ML results into the energy timeline begins `if segment.speaker != "SPEAKER_1"`, so local regions pass through as one speaker no matter how many people are in the room. Those two in-room speakers remain merged in the transcript. That is the real fix and remains #169's scope; akshay's suggestion to cluster the local stem is the right shape for it.

## Verification

- `diarize` suite: 86 passed; `pipeline` suite: 143 passed
- fmt clean, clippy clean
- new tests cover the shared-room case (detected), a brief artifact, a long-but-negligible-share artifact, a too-short clip, and a silent stem

### DIFF
--- a/crates/core/src/diarize.rs
+++ b/crates/core/src/diarize.rs
@@ -2169,6 +2169,44 @@ fn merge_remote_diarization_into_stem_result(
     }
 }
 
+/// Speakers holding enough of a stem's speech to be treated as really present.
+///
+/// Deliberately stricter than `meaningful_speaker_count_excluding`'s
+/// half-second bar, because this gates a safety decision rather than a
+/// descriptive one. An incidental split (a cough picked up as its own cluster,
+/// a diarizer hiccup on one noisy window) must not read as a second person and
+/// suppress self-attribution that would otherwise be correct, so a speaker
+/// counts only when they hold both an absolute and a proportional share of the
+/// attributed speech.
+///
+/// The two people sharing a room in issue #169 held roughly even talk time
+/// across 17 minutes, which clears both bars by a wide margin; the artifacts
+/// this is written to ignore clear neither.
+pub(crate) fn substantial_speaker_count(result: &DiarizationResult) -> usize {
+    const MIN_ABSOLUTE_SECS: f64 = 5.0;
+    const MIN_SHARE_OF_SPEECH: f64 = 0.10;
+
+    let mut speaker_durations: std::collections::HashMap<&str, f64> =
+        std::collections::HashMap::new();
+    for segment in &result.segments {
+        *speaker_durations
+            .entry(segment.speaker.as_str())
+            .or_insert(0.0) += (segment.end - segment.start).max(0.0);
+    }
+
+    let total: f64 = speaker_durations.values().sum();
+    if total <= 0.0 {
+        return 0;
+    }
+
+    speaker_durations
+        .values()
+        .filter(|&&duration| {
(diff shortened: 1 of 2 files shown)


silverstein:
Closing in favor of #714, which keeps the diagnostics and drops the guard. Adversarial review found the guard was both unsafe and largely cosmetic, and I would rather say that plainly than merge it.

**It would fire for solo recorders.** `diarize.rs` already documents that open-speaker setups let the mic acoustically pick up multi-speaker system audio, so a solo recorder on laptop speakers can have remote voices clustered inside their own voice stem. The guard would read that as a shared room and silently stop naming them, removing self-attribution that works today for what is probably a common setup. Distinguishing co-presence from bleed needs the second speaker's speech to be checked against system-stem activity, which is real work and not what that patch did.

**And it was largely cosmetic.** After self-attribution declines, `has_unmapped` becomes true for that label and the LLM mapping fallback runs and can name it anyway. The patch would have traded a medium-confidence deterministic name for a medium-confidence guessed one.

**The threshold was also incoherent.** A 10% share bar means 9 minutes of misattributed speech is tolerated in a 90-minute meeting but 18 seconds is not in a 3-minute standup. There is no safety reason for the tolerated amount of wrong naming to grow with meeting length.

**And the tests did not test it.** They covered only the counting helper; deleting the guard from the pipeline would have left every new test green.

The deeper reason not to invest further here: once the local stem is clustered, that label splits into two, and voiceprint matching names only the cluster that matches. The guard is scaffolding for a problem the real fix dissolves. Design notes are on #169.

PR #715, opened 2026-08-10
Fix the Archive setup screen: broken folder glyph, clipped window, mute back button

silverstein:
Four defects the release operator hit during pilot QA on the installed app. Every fix was verified by driving the installed dev app through its real UI (accessibility API + synthetic clicks) and reading the rendered pixels back, not by inspecting markup.

## What was wrong

- **The approved-location icon rendered as a misaligned sliver.** Two causes, and the first fix only got one of them. The glyph was `U+2302 HOUSE`, absent from the shipped fonts, so the webview substituted a fallback. Underneath that, `.location-copy span` also matched `.location-icon`, and a class+element selector outranks a single class — so its `display: block` overrode the icon box's `place-items: center` and pinned the glyph to the top-left corner. Now an inline SVG folder, with the text rule scoped to the text column.
- **The window opened too short for its own first screen.** 820pt of height meant a default launch required scrolling before the operator could do anything. 960 fits the setup screen with the footer visible (measured, captured, confirmed).
- **The way back to adding folders was unfindable.** From the search screen, "Census report" gave no hint it was also the route to `Change locations`; the operator reported being unable to add folders at all after searching. Now "Census & locations".
- **Singular grammar:** "1 searchable document do not record where a clause ends" → "does not record ... so it was excluded".

## Also

A procedure note in `docs/release/archive-pilot-signing-and-handoff.md`: QA fixtures must not live on an iCloud-synced volume. A run that put them on the Desktop found the canary in `~/Library/Caches/CloudKit/com.apple.bird/` — iCloud's own upload cache, not the application. The app held **zero network sockets** and wrote nothing outside its own space. Keeping fixtures off synced volumes means the leak sweep has one unambiguous answer instead of a distinction to relitigate each run.

## QA verified in this pass

Folder approval, census, index build, exact search (2 provisions with anchors, correct same-clause refusal for the PDF, separately-labeled meaning suggestions), picker cancel, the full navigation loop, window close (process exits, index discarded), zero network sockets, and a clean canary sweep of logs, crash reports, and temp storage.

Dev-build UI only — no Rust changes, no effect on the sealed pilot artifact `b18f007b`.

https://claude.ai/code/session_01MjmKq74A9UFXTYAUbYEjbx

### DIFF
--- a/archive/src-tauri/src/main.rs
+++ b/archive/src-tauri/src/main.rs
@@ -250,6 +250,49 @@ fn reveal_archive_document(
     Ok(())
 }
 
+/// Show an approved location in Finder, named by opaque id.
+///
+/// "Approved location 1" cannot be told from "Approved location 2", and an
+/// owner with three matters approved has no way to confirm he approved the
+/// folders he meant to. The obvious fix -- put the folder's name on the row --
+/// is the wrong one: in a practice the folder name is the sensitive part
+/// ("Smith v. Acme -- privileged" says more than any path does), and sending
+/// it would turn a checkable rule, no filesystem-derived string crosses into
+/// the webview, into a judgement call every later feature has to make again.
+///
+/// So the answer is the one the evidence cards already use. The webview sends
+/// back the id it was given; the path is resolved here, handed to Finder, and
+/// dropped. The owner sees the real folder in its real place, which answers
+/// the question better than a label could, and nothing crosses.
+#[tauri::command]
+fn reveal_archive_location(
+    location_id: u64,
+    state: tauri::State<'_, ArchiveState>,
+) -> Result<(), String> {
+    let session = state.session.lock().map_err(|_| lock_error())?;
+    let location = session
+        .locations
+        .iter()
+        .find(|location| location.id == location_id)
+        .ok_or_else(|| "That location is no longer approved.".to_string())?;
+    let path = location.root.canonical_path();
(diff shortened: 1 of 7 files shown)


silverstein:
Added the approved-location reveal (`93196c15`), which resolves the "Approved location 1 is unidentifiable" problem this PR's label change only partly addressed.

**Why not the folder name.** In a practice the folder name *is* the sensitive part — "Smith v. Acme — privileged" says more than any path does — and the leak vector is the moment the owner shows his screen to someone else. More importantly, sending it would convert a structurally checkable rule (*no filesystem-derived string crosses into the webview*) into a judgement call every later feature has to make again. This codebase has already been burned by that shape once, in the provision-boundaries defect.

**What it does instead.** Extends the road the evidence cards already take: the interface sends back the opaque id it was given, `reveal_archive_location` resolves the path, hands it to Finder, and drops it. The owner sees the real folder in its real place — a better answer than any label — and nothing crosses the boundary. A location since replaced by a symlink, or no longer a directory, is refused rather than followed.

Verified against the installed dev app by driving the real UI: the button renders on the row, and clicking it opened Finder with the approved folder selected.

**Not included:** per-location artifact/byte counts. I claimed those were free and they are not — the census accumulates across all locations, so per-root counts need a change in `archive-core` plus a decision about whether they enter the privacy-reviewed export shape. Worth doing, but it is its own change with its own review.

PR #718, opened 2026-08-10
fix(sdk): stop a never-fed worker's unconfirmed kill from poisoning the process

### LINKED ISSUE #689: Corpus worker: an early-deadline kill of a never-fed worker permanently poisons the process

silverstein:
Fixes #689.

## The defect

An unconfirmed kill (SIGKILL sent, `close` not observed inside the 2s grace) retains the killed lease's memory reservation and refuses every later corpus read with `requires a process restart`. That is correct when the child might still be reading the corpus, and it stays correct.

It is wrong for a child killed **before `begin` reached it**. The protocol is the only way a worker learns which directory to read, and it does nothing before that message, so such a child provably holds no corpus bytes. Charging the process for bytes never read, and locking every later lease until restart, protects nothing.

This is reachable in ordinary use, not just in tests: any short budget that expires while the worker is still starting hits it. On a busy machine, one such moment takes out corpus reads for the rest of an MCP server's lifetime.

## The rule

Poison only when the corpus root was actually disclosed to the child. Everything after `begin` keeps today's behavior exactly.

**The authority is the write itself.** `send` now reports whether the bytes reached the pipe, and it declines to write once the lease has settled, so merely reaching the call site does not count. Buffering means a `true` result proves the bytes were handed to the pipe, not that the child read them. That asymmetry is deliberate: over-reporting disclosure costs one conservative poisoning, while under-reporting would skip one that is required.

## Testing needed a seam, and this one cannot weaken the property

Whether a kill is confirmed inside the grace window is a race no test can force: SIGKILL is untrappable, so no fixture can refuse to die on cue. `workerTerminationGraceMsForTest` makes every kill unconfirmed instead, so the two cases differ only in whether the child was told which corpus to read.

Worth contrasting with the reset seam that adversarial review rejected on #690: that one could *disarm* the fail-closed barrier if reached from production. This one can only make the parent **more** conservative, since a shorter grace means more kills go unconfirmed, and an unconfirmed kill of a worker that knew the corpus still poisons.

## Verification

| | |
|---|---|
| never-fed worker, unconfirmed kill | process stays usable; a following lease succeeds |
| worker that knew the corpus, unconfirmed kill | still poisons; a following lease is refused |

Both fail without the change. Pre-fix, the first test poisons the process and the second then fails too, which is the cascade this fixes, visible in miniature.

They live in the isolated refusal file since they drive process-global state, and the poisoning test asserts the process is still clean when it starts, so its must-stay-last requirement fails loudly rather than silently.

SDK: 148 passed. MCP: 283 passed. Mirror check clean.

### DIFF
--- a/crates/mcp/src/corpus-lease-poisoning.test.ts
+++ b/crates/mcp/src/corpus-lease-poisoning.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { withStableCorpusLease } from "./corpus-lease.js";
+import { retireBoundReadersForProcessShutdown } from "./secure-read.js";
+
+// This file deliberately poisons the worker path and leaves the process
+// unusable, so it must stay a file of exactly one test.
+//
+// An earlier version kept this beside the recovery test with a comment saying
+// it must run last, plus an opening assertion that the process was still
+// clean. That only catches poison arriving from earlier tests; a test added
+// after it would inherit the poison and could pass for the wrong reason, since
+// most assertions here are of the shape "the lease is refused". Vitest runs
+// each file in its own process, so a file boundary enforces what a comment
+// could only request.
+function withCorpus(run: (root: string) => Promise<void>): Promise<void> {
+  const root = mkdtempSync(join(tmpdir(), "minutes-corpus-poisoning-"));
+  return run(root).finally(async () => {
+    await retireBoundReadersForProcessShutdown();
+    rmSync(root, { recursive: true, force: true });
+  });
+}
+
+describe("stable corpus lease poisoning", () => {
+  it("still fails closed after an unconfirmed kill of a worker that knew the corpus", async () => {
+    await withCorpus(async (root) => {
+      writeFileSync(join(root, "meeting.md"), "disclosed canary");
+      // The counterpart to the never-fed recovery case in
+      // corpus-lease-refusal.test.ts, and the regression that fix must not
+      // introduce. A generous budget lets `begin` reach the child, so the
+      // worker knows which directory to read; stalling it forces the kill.
+      //
+      // Once the corpus root has been disclosed, an unconfirmed kill has to
+      // keep failing closed: the child may still be reading, and refusing
+      // every later lease is the intended answer to that.
+      await expect(
+        withStableCorpusLease(root, () => "unreachable", {
+          timeoutMs: 2_000,
+          workerStallPhaseForTest: "before-baseline",
+          forceUnconfirmedTerminationForTest: true,
+        })
+      ).rejects.toThrow("stable meeting corpus authorization failed");
+
+      await expect(
+        withStableCorpusLease(root, () => "must not run")
+      ).rejects.toThrow("requires a process restart");
+    });
+  });
+});
--- a/crates/mcp/src/corpus-lease-refusal.test.ts
+++ b/crates/mcp/src/corpus-lease-refusal.test.ts
@@ -68,16 +68,8 @@ describe("stable corpus lease refusal contract", () => {
     });
   });
 
-  // These two share this file's isolation for the same reason as the test
-  // above: they deliberately drive the worker-poisoning path, which is
-  // process-global. Vitest runs each file in its own process, so a poisoned
-  // outcome cannot leak into the main corpus suite.
-  //
-  // Whether a kill is confirmed inside the grace window is a race no test can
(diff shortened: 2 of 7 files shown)


silverstein:
Adversarial review (codex) found three real problems. All fixed; the reasoning matters more than the diff.

**The admission slot was released for a possibly-live child.** Releasing the memory reservation is defensible, since a never-fed child holds no corpus bytes, but the worker cap counts live processes rather than retained bytes, and an unconfirmed kill means the process may still exist. Releasing immediately let repeated failures accumulate unbounded uncharged children and evade a cap the 256 MiB accounting normally enforces. Holding forever is the other extreme and wrong for a child that almost certainly died a moment after the grace expired. The slot is now handed to the termination promise and freed if and when the child is actually reaped, guarded against a double release. That is neither extreme and is what the review argued for.

**The never-fed test could pass without entering the branch it covers.** This is the sharpest finding. A zero-length grace is still a race, and a never-fed child dies almost immediately, so the termination promise usually wins and the unconfirmed path is skipped entirely. The test would have passed while testing nothing. Replaced the millisecond knob with `forceUnconfirmedTerminationForTest`, a boolean that selects the branch outright, which also removes the DoS surface the review flagged: no delay to coerce, nothing to hold cleanup open for 24 days, and no path where a non-numeric value throws inside the cleanup block and skips the releases.

**My red/green claim was wrong.** I wrote that both tests fail without the change. Reverting only the disclosure gate shows what actually happens:

| | gate reverted |
|---|---|
| never-fed recovery | **fails** (detects the change) |
| disclosed still poisons | passes (pins legacy behavior) |

Both were failing in my earlier check only because the first test poisoned the process and the second inherited it, which is the cascade this work removes, not evidence. Corrected in the commit message.

**Ordering is now enforced rather than requested.** The poisoning test moved to its own file. My opening clean-process assertion only catches poison arriving from earlier tests; anything added after it would inherit the poison and could pass for the wrong reason, since these assertions are mostly "the lease is refused". A file boundary enforces what a comment could only ask for.

On the remaining hook-surface concern: the hooks object already contains `workerScriptForTest`, which lets whoever constructs it run an arbitrary binary as the worker. That is a strictly larger capability than a termination boolean, so the trust boundary is already at "whoever builds the hooks object" and this adds nothing to it.

Confirmed by review and worth recording: the shipped worker performs no corpus traversal, open, read, or watcher creation before `begin`; `send` handles backpressure correctly (a `false` from `write()` still means the string was accepted for buffering); there is no await between the begin write and the flag assignment; and every parent write goes through `send`.

SDK 148 passed across 5 files, MCP 283 passed, mirror clean at 6 files.

PR #725, opened 2026-08-10
feat(sherpa): run sherpa from an isolated plugin so it coexists with diarization

### LINKED ISSUE #685: Isolate the sherpa stack behind its own namespace so engine-sherpa and diarize can share a macOS binary

silverstein:
Closes #685. Restores the combination #683 had to forbid, and with it the path back to shipping the SOTA engine.

## The problem, restated

sherpa-onnx bundles its own ONNX Runtime (1.17.1) and its own kaldi-native-fbank. pyannote brings ONNX Runtime 1.22.0 and a different kaldi-native-fbank under the same archive names. One executable holds one definition per symbol, so linking both statically breaks whichever consumer loses, and every resolution loses somewhere:

| ORT winner | kaldi winner | pyannote | sherpa |
|---|---|---|---|
| 1.17.1 | sherpa's | ✗ API 17/22 | ✓ |
| 1.22.0 | sherpa's | ✗ SIGABRT | ✓ |
| 1.22.0 | knf-rs's | ✓ | ✗ SIGABRT |

#686 responded by refusing the combination at compile time, which also meant pulling `engine-sherpa` out of release builds entirely.

## The fix

sherpa moves to `crates/sherpa-plugin`, a cdylib embedding the whole sherpa stack, which `minutes-core` dlopens. A Rust cdylib exports only its own `#[no_mangle]` surface, so the embedded runtime stays internal and macOS two-level namespaces bind each image's calls to its own copy. **The conflict is removed rather than arbitrated**, so the #683 guard is deleted and its CI assertion is replaced by the inverse: the combination must now compile.

I proved the linking hypothesis with a throwaway spike before writing any of this, since the whole design rests on it.

## Verified on macOS arm64, one CLI binary with both features

```
plugin exports OrtGetApiBase: 0
plugin ORT: 1.17.1
cli ORT: 1.22.0
enroll (pyannote) ........................ ok
transcribe via plugin .................... ok
transcribe --diarize, both in one process  ok
```

## Two things beyond linking that had to be right

**The plugin ships separately, so compiled-in no longer means runnable.** My first pass got this wrong: with the plugin absent, `transcribe` hard-errored. That would break a recording, which is the one thing this codebase refuses to do. An unloadable or ABI-mismatched plugin now falls back to Whisper with a warning naming every path tried, exactly like a missing model. Verified on hardware in both directions:

```
WARN engine = "sherpa" selected but unavailable (feature/model/plugin); falling back to whisper
     plugin="... Tried: /nonexistent/x.dylib: not found; ..."
[0:00] Matt and Wesley are reviewing the Minutes Apple Speech benchmark.
```

**The FFI boundary is unwind-safe.** A panic crossing back into the host is undefined behavior, so every entry point catches; an ABI version is checked before any symbol is taken, because a separately-shipped plugin makes a mismatched pair a real possibility rather than a theoretical one.

## A gap this found in my own tooling

The crate is excluded from the workspace, since building it downloads the sherpa bundle and that would land on every `cargo clippy --all`. Excluding it revealed that `check_ci_rust_filter.mjs` knew about workspace members and non-crate directories but not excluded crates, so a PR touching only the plugin would have skipped every Rust job and reported green. The guard covers exclusions now and reported this itself once taught to look.

## Deliberately not here

Release packaging: shipping the dylib as a release asset and having `setup --sherpa` install it. Until that lands, release builds still omit `engine-sherpa` rather than ship an engine that cannot load. That is the follow-up that actually gets sherpa back into your hands in a downloaded build.


silverstein:
Adversarial review (codex) found seven problems. All addressed. Three were serious enough that this was not mergeable as written.

**A desktop sherpa-to-whisper substitution left no trace.** The worst one, because it is the exact behavior #633 exists to prevent. The artifact warning compares the requested engine against `effective_batch_engine`, and sherpa was missing from that function's fallback list, so it always answered `"sherpa"` regardless of what ran. The CLI tracing warning is dropped by the desktop app, which installs no subscriber. So on desktop, a silent substitution. Sherpa now resolves through the same predicate as the dispatch, the warning names an actionable reason, and a test pins it in all three build shapes.

**The fallback only covered preflight.** Passing the checks is not succeeding: `model_files_present` clears a size floor rather than a hash, so a corrupt model loads the plugin and then fails inside it, and an individual decode window can fail too. Those became a failed transcription, which loses a recording. Whisper now answers whenever sherpa cannot, at any stage. My own module doc had claimed this already, which is the second time on this branch that a comment described behavior I had not implemented.

**The CI gate proved nothing about the property it was named for.** `cargo check` type-checks a feature combination; it does not link a binary, build the plugin, dlopen it, or start either runtime. And the plugin's FFI error-path tests were never executed in CI at all, which is pointed given the panic-across-FFI those tests exist for was found by reading rather than by any gate. CI now runs the plugin's tests **on every platform** (Linux and Windows target branches were previously uncompiled while the README claimed they worked), asserts `cargo tree` shows no `sherpa-rs` in the host graph, and asserts the built dylib exports no `OrtGetApiBase`.

Also fixed:

- **The remediation text named a command that does not do the job.** `minutes setup --sherpa` installs the model, not the plugin, so the advertised fix sent someone round a loop. Corrected in both the load failure and the ABI mismatch.
- **Failure was cached forever**, with a comment claiming the answer could not change without a restart. Plainly false for a filesystem probe: a long-lived MCP server could never pick up a plugin installed after start. Only success is cached now.
- **`engine-sherpa` did not imply `whisper`**, so in the no-default build CI advertises, the "fallback" returned a diagnostic placeholder as though it were a transcript. The feature implies whisper now, so the promise holds.

Not changed, deliberately: `sherpa_engine::transcribe_segments` still returns `Err` rather than falling back. That is the raw engine, and it now documents that, with a pointer to the dispatch for callers that must not lose a recording.

Claims the review confirmed: FFI ownership is correct across the boundary, the `Send`/`Sync` shape is sound, no sherpa stack remains in the host image (`cargo tree -i sherpa-rs` finds nothing), the standalone lockfile is committed, and live-transcript and dictation never reach sherpa directly.

PR #727, opened 2026-08-11
fix(macos): Apple Speech worker Swift Concurrency runtime rpath

silverstein:
## What

The Apple Speech worker links the Swift Concurrency bridge (Task, async). Its binary carries a weak `@rpath` load of `libswift_Concurrency`, which on macOS lives only in the dyld shared cache, not on disk, so that `@rpath` resolves only with a runtime rpath into `/usr/lib/swift`. build.rs added `/usr/lib/swift` as a link *search* path but never as a runtime rpath, so the weak load was silently skipped, every Concurrency type descriptor stayed null, and the worker aborted inside `swift_getTypeByMangledName` the instant the analyzer's async types resolved.

## Why this is the real root cause

This is the true cause of every prior signed-acceptance runtime crash. Across the investigation it was misattributed to the hosted VM and to missing Speech assets, because the only thing ever run on real hardware was a clang-linked C-driver harness, and clang links the Swift concurrency rpath automatically. The Rust-linked worker never got it. Running the real Rust-linked worker on hardware is what exposed it.

## The fix

- `crates/cli/build.rs`: add `cargo:rustc-link-arg-bin=minutes-apple-speech-worker=-Wl,-rpath,/usr/lib/swift`. This is the shipping fix; a `cargo:rustc-link-arg` in the minutes-core library build script does NOT reach a downstream binary (that mistake was caught by an `otool` check on the actual binary).
- `crates/core/build.rs`: the same rpath for minutes-core's own test/example targets.
- `check_apple_speech_worker_packaging.mjs`: structural guards for both, proven to fail when either is removed. build.rs is deliberately not golden-hashed so unrelated build edits need no reseal.

## Proof (Apple Silicon, macOS 26.6)

- The real Rust-linked worker path (`process_private_audio_request`, non-echo, real demo audio) transcribes 28 words, `runtimeSupported=true`, where it aborted before.
- `otool -l` confirms the shipping `minutes-apple-speech-worker` binary now carries `path /usr/lib/swift`; the graph worker is unaffected.
- clippy `-D warnings` clean, fmt clean, all packaging guards + self-tests pass.

## Scope

The product gate stays hard false: this makes the still-dormant worker capable of running Speech, it does not activate anything. Unblocks the Apple Speech activation epic (beads minutes-i279, minutes-uyc8; plan PR #726).

### DIFF
--- a/crates/cli/build.rs
+++ b/crates/cli/build.rs
@@ -29,4 +29,18 @@ fn main() {
             plist.display()
         );
     }
+
+    // The Apple Speech worker links the Swift Concurrency bridge (Task, async).
+    // Its binary carries a weak @rpath load of libswift_Concurrency, which on
+    // macOS lives only in the dyld shared cache, not on disk, so that @rpath
+    // resolves only with a runtime rpath into the OS Swift libraries. Without
+    // it the weak load is silently skipped, every Concurrency type descriptor
+    // is null, and the worker aborts inside swift_getTypeByMangledName the
+    // instant the analyzer's async types resolve. A cargo:rustc-link-arg from
+    // the minutes-core build script does NOT reach this downstream binary, so
+    // the rpath must be set here, on the worker bin itself. Proven on Apple
+    // Silicon: without this the real worker aborts; with it it transcribes.
+    println!(
+        "cargo:rustc-link-arg-bin=minutes-apple-speech-worker=-Wl,-rpath,/usr/lib/swift"
+    );
 }
--- a/crates/cli/build.rs
+++ b/crates/cli/build.rs
@@ -40,7 +40,5 @@ fn main() {
     // the minutes-core build script does NOT reach this downstream binary, so
     // the rpath must be set here, on the worker bin itself. Proven on Apple
     // Silicon: without this the real worker aborts; with it it transcribes.
-    println!(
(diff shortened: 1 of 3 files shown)


silverstein:
## Codex adversarial review: PASS (one P2, fixed)

Ran a codex adversarial pass on the four risk areas. No P1. Codex confirmed: rpath placement correct, the resulting `LC_RPATH` is covered by the worker's code signature, `/usr/lib/swift` resolves via the dyld shared cache even under App Sandbox, no dylib-hijack/precedence risk (absolute SIP-controlled system path, not writable or bundle-relative), no non-macOS linker effect (both additions are behind `CARGO_CFG_TARGET_OS == macos` gates), and the signed transport acceptance is not invalidated (packaging re-computes and re-seals the worker CDHash).

One [P2], now fixed: the rpath guard used raw `source.includes`, so a commented-out or dead-coded `println!` would still pass. Hardened to strip comments via the existing `activeCode()` helper and added self-test mutations that comment out and remove both rpaths; a real commented-out worker rpath now fails the guard (verified).

Codex also suggested an `otool -l` LC_RPATH assertion in the macOS CI job as defense-in-depth. Sensible follow-up, left out of this one-line link fix; the shipped worker binary was already confirmed to carry `path /usr/lib/swift` via `otool` during verification.

Issue #645, opened 2026-08-04
Ship the sherpa engine on Linux and Windows (needs dylib bundling)


silverstein:
## Context

#644 enables `engine-sherpa` in the **macOS** release builds. It is deliberately not enabled for Linux or Windows, and this issue tracks closing that gap.

## Why macOS only today

From `crates/core/Cargo.toml`:

- **macOS**: `sherpa-rs` with `features = ["download-binaries", "static"]`. Static linkage means self-contained binaries with no dangling `@rpath` dylib references, which was the #369 packaging blocker.
- **Linux**: `features = ["download-binaries"]` only, with the note that `sherpa-rs-sys`'s `static` "panics in build.rs here unless a global RUSTFLAGS relocation hack is set."
- **Windows**: same dynamic situation.

Dynamic linkage means the binary needs `libsherpa-onnx-c-api` beside it. Confirmed in practice: a Linux build of `minutes` with `engine-sherpa` fails at startup with

```
error while loading shared libraries: libsherpa-onnx-c-api.so: cannot open shared object file
```

and only runs with `LD_LIBRARY_PATH` pointed at `target/<profile>/`. Shipping that as a release artifact would produce binaries that do not start.

## Why it is worth doing

Benchmarked on LibriSpeech test-clean (12 clips, 777s, 2050 reference words, M1):

| Engine | WER | Errors / 2050 | RTF |
|---|---|---|---|
| sherpa parakeet-v3-int8 | 2.29% | 47 | 0.28 |
| parakeet tdt-ctc-110m | 2.39% | 49 | 0.24 |
| whisper small (default) | 3.46% | 71 | 2.86 |

Roughly a third fewer errors than the shipping default and an order of magnitude faster. Leaving Linux and Windows on whisper-small is a real capability split, not a cosmetic one, and it is the kind of split that quietly becomes permanent.

## Options

1. **Make static linkage work off-darwin.** The Cargo.toml note implies a known RUSTFLAGS relocation workaround. If it can be scoped to this crate rather than set globally, this is the cleanest outcome: same self-contained story as macOS.
2. **Bundle the dylibs.** Ship `libsherpa-onnx-c-api` (and its onnxruntime dependency) alongside the binary, with `$ORIGIN` rpath on Linux and the DLL next to the exe on Windows. Works, but changes the "single binary" distribution shape and needs installer and archive changes.
3. **Leave it macOS-only.** Acceptable short term, but should be a stated decision rather than drift.

## Acceptance

[...]

PR #733, opened 2026-08-11
fix(release): ship the sherpa plugin in the Linux archive again

silverstein:
## A release-blocking regression from #725, which I merged

#685 moved sherpa into a dlopened cdylib outside the workspace. That means `-p minutes-cli --features engine-sherpa` now builds only the loader and emits no sherpa shared library at all.

The Linux archive step still copied `*.so` out of the CLI's target directory. It finds nothing, and the step's own presence check then fails:

```
test -f "$out/libsherpa-onnx-c-api.so" || { echo "sherpa shared library missing from the archive"; exit 1; }
```

So the next `v*` tag would have failed the release. Confirmed on Linux rather than reasoned about: `cargo tree -p minutes-cli --features engine-sherpa` contains no sherpa crate, and a real build of that feature combination produces zero `.so` files.

**`Release CLI Binaries` runs only on tags and `workflow_dispatch`**, which is why every gate on #725 was green while the shipping artifact was already broken. That gap is the more important half of this fix.

## What the archive does now

Builds `crates/sherpa-plugin` and ships the plugin beside the executable, which is one of the paths `sherpa_plugin::candidate_paths` searches, together with the sherpa libraries the plugin resolves through its own `$ORIGIN`.

## Three things so it cannot recur quietly

**`scripts/check_sherpa_packaging.mjs`** asserts the packaging contract statically on every PR, wired into Lint workflows next to the existing graph-worker and Apple-speech packaging guards. It fails against the workflow as it stands on main today, with five specific findings, which is the check that matters: a guard that only passes on the fixed version proves nothing. Its `--self-test` rejects four separate regressions (dropping the plugin build, not shipping the plugin, dropping the load check, removing the step).

**`scripts/verify_sherpa_plugin_loads.py`** dlopens the packaged plugin from outside the build tree with `LD_LIBRARY_PATH` unset, and checks it exports `minutes_sherpa_abi_version`. The binary's own `--version` smoke test cannot reach this, because the plugin is loaded lazily and only once sherpa is genuinely selected: a packaged CLI runs `--version` perfectly with an unloadable plugin sitting next to it. Exercised against all four real outcomes (good archive, plugin without its libraries, plugin absent, a non-plugin `.so`).

**`crates/sherpa-plugin/build.rs`** sets the `$ORIGIN` runpath itself instead of relying on a `RUSTFLAGS` the packaging step happened to export. Without it the plugin only resolved its libraries while sitting in the build tree, so the hand-install that `docs/architecture/sherpa-engine.md` documents produced a plugin that could not load. That `RUSTFLAGS` is now removed from the workflow, along with its stale claim that the executable needs sherpa libraries beside it, which stopped being true at #685.

## Verification

The release step itself was extracted from the YAML and run on Linux with `RUSTFLAGS` unset. It produces a 25 MB archive whose plugin dlopens from `/tmp` with `LD_LIBRARY_PATH` unset and reports ABI 1. The rewritten doc instructions were then run verbatim into a scratch `~/.minutes/lib` and load too, which the previous instructions did not.

Plugin tests still pass (4), `cargo fmt` clean, both workflow files parse.

## Deliberately not in this PR

`release-macos.yml` carries two comments saying `engine-sherpa` is held out because it "conflicts with diarize on macOS; see crates/core/src/lib.rs guard". Both are now wrong: #725 removed that guard and dissolved the conflict. I left them because that file is hash-sealed as packaging authority, and breaking the seal for a comment edit is exactly what the seal exists to make deliberate. Worth a follow-up that reseals on purpose. macOS still holds sherpa out for the real reason, which is signing and notarization, tracked on #369.

Windows needs different handling entirely, since it resolves a DLL's dependencies from the loading process's search path rather than the DLL's own directory. That stays #645.


silverstein:
Pushed f18b1b0d after a codex adversarial pass, which found four real defects in the guards I had just added. All four would have let the regression they exist to catch come straight back, so recording them here.

**The static guard was substantially decoration.** It sliced the raw file and matched substrings. It accepted a workflow that deleted the plugin copy, repointed the library source back at the old CLI target directory, commented out the plugin build or the load check while leaving their words visible in the comment, or ran the verifier against the build-tree plugin instead of the packaged one. It also rejected correct alternatives such as `--manifest-path` and `[[ -f ... ]]`, which is its own failure mode: a guard that only accepts one spelling gets worked around rather than satisfied.

It is now Python, parses the workflow with PyYAML (already a CI dependency via `check_signing_secret_scope.py`), strips comment lines before matching, keys on `id: sherpa-archive` rather than the step name, and joins backslash continuations so it inspects the loader's actual argument rather than a window that ran to the end of the script. The self-test now covers eight regressions plus two correct variants that must not be rejected. It still fails against the workflow as it stands on main.

**The load verifier could approve a plugin the host refuses.** It accepted any ABI at or above 1 while `sherpa_plugin.rs` requires exact equality, so bumping the plugin ABI without the host constant would have published an archive whose engine the binary beside it rejects at runtime. It also checked one symbol when the host resolves five. Both the expected ABI and the symbol list now come out of `crates/core/src/sherpa_plugin.rs` rather than being restated, because a restated copy drifts and the drift is invisible until release.

**Two shell holes.** `find -exec cp \;` exits 0 even when every copy failed, verified directly: the same failing command returns 0 with `\;` and 1 with `+`. And `rm -rf "$out" && mkdir` does not trip `errexit` on a failed removal, so a half-removed stale directory could have survived into the archive. `libonnxruntime.so` was also unasserted, which mattered because only `LD_LIBRARY_PATH` is cleared for the dlopen check, so a system copy could have satisfied it.

**The docs had the same class of bug as the code.** The recommended quick start copied a `.dylib` unconditionally, which installs nothing on Linux, so someone following the top of the page got no plugin at all. And the macOS build section still told people to drop `--no-default-features` and give up diarization, for a conflict #685 dissolved. Both fixed, and verified rather than asserted: `cargo check -p minutes-cli --features engine-sherpa` compiles with the default features, and CI already requires `diarize,engine-sherpa` to compile.

The release step was re-extracted from the YAML by its new id and run end to end again after all of this. It still produces an archive whose plugin loads from `/tmp` with `LD_LIBRARY_PATH` unset, now reporting that the ABI matches the host and all five required symbols are exported.

Issue #736, opened 2026-08-11
Automate the Homebrew cask bump in the release pipeline (cask lagged at 0.18.2 while releases reached 0.24.0)


erikcw:
The current release to 0.24.0.  The Minutes homebrew cask is still at version 0.18.2.  Perhaps updating homebrew can be added to the CI process.

```bash
brew update                                                                                                                                                                                             
brew upgrade --cask silverstein/tap/minutes
brew info --cask silverstein/tap/minutes   # check version again
==> Updating Homebrew...
Already up-to-date.
Warning: Calling string comparison format for `depends_on macos:` is deprecated! Use `depends_on macos: :big_sur` instead.
Please report this issue to the silverstein/homebrew-tap tap (not Homebrew/* repositories), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/silverstein/homebrew-tap/Casks/minutes.rb:11

Warning: Not upgrading minutes, the latest version is already installed
Warning: Calling string comparison format for `depends_on macos:` is deprecated! Use `depends_on macos: :big_sur` instead.
Please report this issue to the silverstein/homebrew-tap tap (not Homebrew/* repositories), or even better, submit a PR to fix it:
  /opt/homebrew/Library/Taps/silverstein/homebrew-tap/Casks/minutes.rb:11

==> minutes ✔ (Minutes): 0.18.2
Privacy-first conversation memory — record, transcribe, search meetings locally
https://useminutes.app/
Installed (on request)
/opt/homebrew/Caskroom/minutes/0.18.2 (48MB)
  Installed on 2026-08-11 at 12:00:56
```

silverstein:
Thanks for the report, and for including the full brew output, which made this a one-look diagnosis.

You found a real process gap. The tap's CLI formula was current at 0.24.0, but the cask stopped getting bumped at 0.18.2, so desktop installs fell six releases behind while every one of those releases had a perfectly good DMG sitting on GitHub.

Fixed as of a few minutes ago. The cask now points at the 0.24.0 DMG (sha256 computed from the release asset itself), and the deprecated `depends_on macos:` string comparison you were warned about is corrected to the symbol form in the same commit. This should get you current:

```bash
brew update
brew upgrade --cask silverstein/tap/minutes
brew info --cask silverstein/tap/minutes
```

You should see 0.24.0 and no deprecation warning.

[...]

PR #737, opened 2026-08-11
feat(release): bump the Homebrew tap from the release event

silverstein:
Refs #736. Deliberately not an auto-close: the issue stays open until the secret exists, since a merged-but-inert workflow has not actually fixed anything yet.

## Why

The tap's cask sat at 0.18.2 while releases reached 0.24.0. Nothing was broken; bumping it was simply a step someone had to remember, and release-time steps rot invisibly between releases. erikcw found it, we didn't.

`Update Homebrew Tap` fires on `release: published` and points both tap files at the version. Publishing rather than tagging is the trigger deliberately: the release procedure creates a draft, waits for assets, then un-drafts. Publishing is the first moment the DMG the cask has to hash is guaranteed to exist. A tag push fires while those assets are still building.

## One thing needed from you

The default `GITHUB_TOKEN` cannot write to another repository, so this needs a secret. Until it exists the workflow warns and exits 0 rather than turning a good release red, so **merging this is safe and inert**.

1. https://github.com/settings/personal-access-tokens/new
2. Resource owner `silverstein`, repository access **Only select repositories** → `silverstein/homebrew-tap`
3. Repository permissions → **Contents: Read and write**. Nothing else.
4. Copy the token, then:

```bash
gh secret set HOMEBREW_TAP_TOKEN --repo silverstein/minutes
```

To confirm it works without waiting for a release, run the workflow manually with dry run left on. It will report what it would change and write nothing.

## Why the logic is a script, not inline YAML

`scripts/bump_homebrew_tap.py` can be exercised; a `run:` block can only be read.

- Edits are substitutions that must match **exactly once**. A restructured tap file makes it refuse rather than guess, because a wrong `sha256` is worse than a stale one: stale installs an old version, wrong fails every install with a checksum mismatch.
- The cask hash is computed from the DMG bytes, not read from `SHA256SUMS.txt`, which does not list the DMG at all.
- Idempotent, so re-running a release workflow is a no-op that exits 0.
- `--self-test` covers the transforms, three refusal cases, and idempotence, and runs as the first step of the job so broken logic cannot reach the token.

## Token handling

The job holds a credential that can write to another repository, so it takes nothing else it does not need: `permissions: contents: read` for the default token, `persist-credentials: false` on checkout, no third-party actions in the job at all, and the release tag reaches the shell through `env` rather than expression interpolation so a crafted release name cannot become code.

## Verification

Against the live tap and real releases:

- Dry run at 0.24.0 correctly reports the tap is already current and no-ops.
- Dry run at 0.23.0 exercises the real edit path and computes `70a61db0c11e...`, which matches an independent `sha256sum` of the downloaded DMG.
- Version resolution tested for all four input shapes (`release` event tag with and without `v`, `workflow_dispatch` input with and without `v`).
- actionlint clean, checked by exit code. Its first pass caught an unquoted expansion, and fixing that also caught a real bug of mine: the verification step gated on a token variable it never declared, so it always skipped. It now gates on whether a write actually happened and fails on mismatch.

The manual path stays documented for when the workflow is unavailable, and now points at the same script rather than hand-editing.

### DIFF
--- a/.github/workflows/homebrew-tap.yml
+++ b/.github/workflows/homebrew-tap.yml
@@ -28,16 +28,36 @@ on:
 permissions:
   contents: read
 
+# Two runs must not interleave. Each writes a single commit, but run A can
+# still read the tap, run B commit, and run A then commit over a base it never
+# saw. The non-forced ref update in the script makes that fail rather than
+# silently win, and this keeps it from arising at all.
+concurrency:
+  group: homebrew-tap
+  cancel-in-progress: false
+
 jobs:
   bump:
     name: Point the tap at the release
     runs-on: ubuntu-latest
+    # Creating a release or dispatching a workflow needs only write access, so
+    # without a gate the trust boundary for a cross-repository token is every
+    # write collaborator rather than the maintainer. Adding required reviewers
+    # to this environment in repository settings closes that; the environment
+    # exists either way, so this is safe before it is configured.
+    environment: homebrew-tap
     steps:
       - uses: actions/checkout@v7
         with:
           persist-credentials: false
+          # The default branch, never the event ref. A release is published
+          # from a tag, and `workflow_dispatch` can select any ref, so taking
+          # the event ref would run whatever version of the bump script that
+          # ref happens to contain while the tap token is in the environment.
+          ref: ${{ github.event.repository.default_branch }}
 
       - name: Verify the bump logic before using it
+        # First, so broken logic cannot reach the token.
         run: python3 scripts/bump_homebrew_tap.py --self-test
 
       - name: Resolve version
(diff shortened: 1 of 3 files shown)


silverstein:
Pushed a63923b6 after a codex adversarial pass, which found five defects, four high severity. Every one of them could have published a tap that **breaks installs** rather than merely lagging, which is the outcome this whole PR exists to avoid.

**A truncated download could produce a wrong hash.** A sized read on a connection that closes early does not necessarily raise, so hashing until the stream ends can digest a partial file and write that hash into the cask. Reproduced, both by the reviewer and again by me against the fixed code. The asset must now be in state `uploaded`, the downloaded byte count must equal the declared size, and the computed hash must equal the digest GitHub itself records. Any disagreement refuses. GitHub does publish a `digest` field, and it matches what I compute for the current DMG, so this is a real cross-check rather than a formality.

**A matching version suppressed hash repair.** If a cask carried the right version and a wrong `sha256`, the script reported "already current" and the workflow verified only version strings. That combination made a corrupt hash permanently unfixable by this tool and invisible to its own check. The cask is now current only when version and hash both agree, and the postcondition re-runs the bump in dry-run mode against the published tap, so it re-derives the expected hash instead of comparing strings.

**The two writes were not atomic, and validation happened after the first one.** A typo'd version bumped the formula, then failed the release lookup, leaving the tap half-updated with the verification step skipped. Everything is now read, downloaded and checked before anything is written, and both files land in a single commit through the Git Data API with a non-forced ref update, so a concurrent run fails rather than interleaving a formula and cask from different versions. A `concurrency` group keeps that from arising at all.

**The job checked out the event ref while holding a cross-repository token.** Publishing a release or dispatching a workflow needs only write access, so the effective trust boundary was every write collaborator, not the maintainer: push a tag carrying a modified `bump_homebrew_tap.py`, publish a release for it, and the script runs with the tap token. It now checks out the default branch explicitly, and the job runs in a `homebrew-tap` environment you can give required reviewers in repository settings. The environment exists implicitly, so this is safe before you configure it.

**The token was forwarded across the asset redirect.** Release asset downloads 302 to object storage, and the stdlib redirect handler copies `Authorization` to the new host. I confirmed that against the installed runtime rather than taking it on faith. A handler now strips it on a host change, and the self-test exercises both the same-host and cross-host cases.

Prereleases are also skipped explicitly now, rather than only being refused by the version guard further in.

Verification: the self-test grew from 5 checks to 15, including five rejected version shapes and both redirect cases. The truncation and digest-mismatch paths were exercised directly against fabricated assets. Live dry-runs still behave: 0.24.0 is a true no-op that now confirms the hash as well as the version, and 0.23.0 exercises the real edit path and plans a single two-file commit. actionlint clean, checked by exit code.

PR #754, opened 2026-08-13
feat(mcp): add an opt-in Streamable HTTP transport

erikcw:
Adds an opt-in Streamable HTTP transport to the MCP server. stdio stays the default, and with no flag nothing changes: that path is byte-for-byte what it is today.

```bash
minutes-mcp --transport http --port 7373
```

Six commits, each of which compiles on its own.

## Why a server factory had to come first

The first commit looks like unrelated churn, so it is worth saying why it is there. `crates/mcp/src/index.ts` builds one module-level `McpServer` and registers every tool, resource and handler onto it at module scope, across about 9000 lines. HTTP cannot reuse that. Two constraints in `@modelcontextprotocol/sdk` 1.30.0, both read out of the installed source rather than assumed:

- `Protocol.connect()` throws "Already connected to a transport" when a second transport is attached to the same server.
- `WebStandardStreamableHTTPServerTransport.handleRequest()` throws "Stateless transport cannot be reused across requests" once `sessionIdGenerator` is undefined and the transport has served one request. The SDK's own comment gives the reason: message id collisions between clients.

So HTTP needs one `McpServer` plus one transport per session, and there is no shared-transport shortcut. Rather than indent 4600 lines into a factory function, registrations now record into a builder list that `createMinutesServer()` replays onto a fresh instance. Handlers still close over module state, so instances share the CLI bridge, caches and leases; only protocol state is per-instance.

## Security

HTTP mode has no authentication. It binds `127.0.0.1` and rejects any request whose `Host` or `Origin` header is not loopback, plus a content-type check. Binding to loopback is not enough on its own: a page already open in your browser can POST to a local port, so the header checks are the part that actually stops that. The README says the same thing next to the config example.

## Tests

New under vitest: `httpTransport.test.ts` and `serverFactory.test.ts`. New as integration tests: `mcp_http_transport_test.mjs` (12 checks) and `surface_parity_test.mjs` (16). CI runs `npm run test:unit`, so the vitest half is enforced there. The `.mjs` tests are local-only, which is how `mcp_tools_test.mjs` and `copilot_mcp_contract.mjs` already work here, so this follows existing practice rather than adding a gap.

`serverFactory.test.ts` is the one that earns its keep. It compares a factory instance's surface against the stdio singleton, so "registered on the singleton, invisible over HTTP" fails a test instead of shipping.

The parity helper in `test/lib/surface-parity.mjs` is bigger than it looks like it should be, and the reason is worth knowing before you read it. Comparing whole `tools/list` name lists between two separately spawned servers is not a deterministic assertion: each one runs `probeCapabilitiesSync` at module load on a 2s `execFileSync` budget, and a cold or contended CLI blows that budget and silently drops eight tools and two resources. So I split the check instead of loosening it. The ungated core is compared exactly across processes. The capability-gated names are compared per process against that process's own probe outcome, read from the `cli-capabilities-*` line each process writes under its own pid. That channel reports what a process decided rather than what it advertised, and from the advertised surface alone, "the probe timed out" and "the factory dropped the gated tools" look identical. The second one is the defect the test exists to catch. A missing witness fails rather than quietly degrading to a weaker assertion.

## Three things I would like your read on

1. Default port 7373. Arbitrary, but unassigned by IANA and clear of 3000/5173/8000/8080. Trivial to change if you would rather it were something else.
2. `--help` used to launch stdio and now prints usage and exits 0. That is the branch's one behaviour change to an existing invocation. I judged it safe because no MCP host passes `--help`, but it is not a strict zero-delta and I would rather flag it than have you find it.
3. Shared process state, which is the thing the diff does not show. Under stdio every client got its own process, so its own `cliAvailable` cache, corpus lease state, copilot subscription set and `MINUTES_BIN`. Under HTTP all sessions share one module scope. For recording that is arguably more correct, since it is machine-global behind a PID file and flock either way. But the concurrency test only calls `listTools` from two sessions, which touches no shared mutable state, so overlapping tool execution against shared leases and caches is not exercised in this branch. I would rather say that plainly than imply it was covered.

One thing I left out on purpose: running stdio and HTTP in the same process is cheap now that the factory exists, but a process owns one stdin, so it would only ever mean "HTTP alongside a host-spawned stdio session". Nobody has asked for that.

## What comes after this

Two more branches build on this one. I will open them as separate PRs once this lands rather than piling them in here: an MCP prompts primitive, since `prompts/list` currently answers Method not found, and an opt-in tool-search mode that cuts the `tools/list` payload by about 94%. Both need per-instance wiring through the builder list this PR adds, which is the only reason they are sequenced behind it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,33 @@ Minutes exposes a standard MCP server. Point any MCP-compatible client at it:
 }
 ```
 
+#### Transports
+
+`minutes-mcp` speaks stdio by default, which is what the config above uses: the client spawns the server as a subprocess and owns its lifetime. Nothing changes for existing setups.
+
+Pass `--transport http` to run one long-lived Streamable HTTP server instead, so several clients share a single process rather than each spawning their own:
+
+```bash
+minutes-mcp --transport http --port 7373
+```
+
+Then point clients at the endpoint:
+
+```json
+{
+  "mcpServers": {
+    "minutes": {
+      "type": "http",
+      "url": "http://127.0.0.1:7373/mcp"
+    }
+  }
+}
+```
+
+Flags: `--host` (default `127.0.0.1`), `--port` (default `7373`; `0` asks the OS for a free port and the chosen one is printed to stderr), `--max-sessions` (default 16). Each accepts an environment variable instead: `MINUTES_MCP_TRANSPORT`, `MINUTES_MCP_HOST`, `MINUTES_MCP_PORT`, `MINUTES_MCP_MAX_SESSIONS`. `GET /health` reports live session count.
+
+HTTP mode has no authentication. It binds `127.0.0.1`, so only this machine can reach it, and it rejects requests whose `Host` or `Origin` is not loopback. That second check matters: binding to loopback does not by itself stop a web page you have open from POSTing to a local port. Do not bind a public address or forward the port.
+
 Canonical MCP reference now lives at:
 
 - <https://useminutes.app/docs/mcp/tools>
--- a/README.md
+++ b/README.md
@@ -519,6 +519,8 @@ Then point clients at the endpoint:
 
 Flags: `--port` (default `7373`; `0` asks the OS for a free port and the chosen one is printed to stderr), `--max-sessions` (default 16). Each accepts an environment variable instead: `MINUTES_MCP_TRANSPORT`, `MINUTES_MCP_PORT`, `MINUTES_MCP_MAX_SESSIONS`. `GET /health` reports live session count.
 
+Request bodies are JSON-RPC envelopes, so they are capped at 4 MiB; anything larger gets a JSON-RPC error with HTTP 413. A client that disconnects without sending `DELETE /mcp` leaves its session behind, so a session with no request in flight and no open stream is reclaimed after 30 minutes idle and answers 404 after that. A client holding an open stream is never reclaimed, however quiet it is.
+
 HTTP mode has no authentication, so the bind address is always `127.0.0.1` and there is no flag to change it. Only this machine can reach the endpoint, and it rejects requests whose `Host` or `Origin` is not loopback. That second check matters: binding to loopback does not by itself stop a web page you have open from POSTing to a local port.
 
 To reach it from another machine, front it with a reverse proxy and put authentication there. The proxy has to rewrite `Host` to the upstream address, since the original name would fail the loopback check. In Caddy:
--- a/README.md
+++ b/README.md
@@ -517,9 +517,19 @@ Then point clients at the endpoint:
 }
 ```
 
-Flags: `--host` (default `127.0.0.1`), `--port` (default `7373`; `0` asks the OS for a free port and the chosen one is printed to stderr), `--max-sessions` (default 16). Each accepts an environment variable instead: `MINUTES_MCP_TRANSPORT`, `MINUTES_MCP_HOST`, `MINUTES_MCP_PORT`, `MINUTES_MCP_MAX_SESSIONS`. `GET /health` reports live session count.
+Flags: `--port` (default `7373`; `0` asks the OS for a free port and the chosen one is printed to stderr), `--max-sessions` (default 16). Each accepts an environment variable instead: `MINUTES_MCP_TRANSPORT`, `MINUTES_MCP_PORT`, `MINUTES_MCP_MAX_SESSIONS`. `GET /health` reports live session count.
 
-HTTP mode has no authentication. It binds `127.0.0.1`, so only this machine can reach it, and it rejects requests whose `Host` or `Origin` is not loopback. That second check matters: binding to loopback does not by itself stop a web page you have open from POSTing to a local port. Do not bind a public address or forward the port.
+HTTP mode has no authentication, so the bind address is always `127.0.0.1` and there is no flag to change it. Only this machine can reach the endpoint, and it rejects requests whose `Host` or `Origin` is not loopback. That second check matters: binding to loopback does not by itself stop a web page you have open from POSTing to a local port.
+
+To reach it from another machine, front it with a reverse proxy and put authentication there. The proxy has to rewrite `Host` to the upstream address, since the original name would fail the loopback check. In Caddy:
+
+```
+reverse_proxy 127.0.0.1:7373 {
+  header_up Host {upstream_hostport}
+}
+```
+
+Native MCP clients send no `Origin`, so nothing else is needed for them. A browser-based client would also need its `Origin` handled at the proxy, which gives up the CSRF defense described above.
 
 Canonical MCP reference now lives at:
 
--- a/crates/mcp/package.json
+++ b/crates/mcp/package.json
@@ -32,11 +32,12 @@
   "scripts": {
     "build": "tsc && vite build",
     "build:ui": "vite build",
-    "test": "npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs",
+    "test": "npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs && node test/mcp_http_transport_test.mjs",
     "test:unit": "vitest run",
     "test:host-compat": "node test/live_events_host_compat.mjs",
     "test:copilot": "node test/copilot_mcp_contract.mjs",
-    "test:all": "vitest run && npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs",
+    "test:http": "node test/mcp_http_transport_test.mjs",
+    "test:all": "vitest run && npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs && node test/mcp_http_transport_test.mjs",
     "start": "node dist/index.js",
     "dev": "tsx src/index.ts"
   },
--- a/crates/mcp/package.json
+++ b/crates/mcp/package.json
@@ -32,12 +32,13 @@
   "scripts": {
     "build": "tsc && vite build",
     "build:ui": "vite build",
-    "test": "npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs && node test/mcp_http_transport_test.mjs",
+    "test": "npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs && node test/mcp_http_transport_test.mjs && node test/surface_parity_test.mjs",
     "test:unit": "vitest run",
     "test:host-compat": "node test/live_events_host_compat.mjs",
     "test:copilot": "node test/copilot_mcp_contract.mjs",
     "test:http": "node test/mcp_http_transport_test.mjs",
-    "test:all": "vitest run && npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs && node test/mcp_http_transport_test.mjs",
+    "test:surface-parity": "node test/surface_parity_test.mjs",
+    "test:all": "vitest run && npm run build && node test/mcp_tools_test.mjs && node test/copilot_mcp_contract.mjs && node test/mcp_http_transport_test.mjs && node test/surface_parity_test.mjs",
     "start": "node dist/index.js",
     "dev": "tsx src/index.ts"
   },
--- a/crates/mcp/src/httpTransport.test.ts
+++ b/crates/mcp/src/httpTransport.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, it } from "vitest";
 
 import {
-  DEFAULT_HTTP_HOST,
+  HTTP_BIND_HOST,
   DEFAULT_HTTP_PORT,
   isAllowedContentType,
   isAllowedHostHeader,
@@ -9,36 +9,38 @@ import {
 } from "./httpTransport.js";
 
 describe("http transport request guards", () => {
-  it("binds loopback by default", () => {
-    expect(DEFAULT_HTTP_HOST).toBe("127.0.0.1");
+  it("binds loopback, with nothing to override it", () => {
+    expect(HTTP_BIND_HOST).toBe("127.0.0.1");
     expect(DEFAULT_HTTP_PORT).toBeGreaterThan(1024);
   });
 
   describe("Host header", () => {
-    it("accepts loopback names and the bound address", () => {
+    it("accepts loopback names", () => {
       for (const host of [
         "127.0.0.1:7373",
         "localhost:7373",
         "localhost",
         "[::1]:7373",
       ]) {
-        expect(isAllowedHostHeader(host, "127.0.0.1")).toBe(true);
+        expect(isAllowedHostHeader(host)).toBe(true);
       }
     });
 
-    it("accepts the explicitly bound non-loopback address", () => {
-      expect(isAllowedHostHeader("192.168.1.10:7373", "192.168.1.10")).toBe(true);
+    // The allowlist is fixed, so a LAN address never passes. Before --host was
+    // removed, binding one added it here, which is the hole that removal closed.
+    it("rejects a non-loopback address", () => {
+      expect(isAllowedHostHeader("192.168.1.10:7373")).toBe(false);
+      expect(isAllowedHostHeader("0.0.0.0:7373")).toBe(false);
     });
 
     // A rebound hostname resolves to 127.0.0.1 but still sends its own name.
     it("rejects a rebound hostname", () => {
-      expect(isAllowedHostHeader("attacker.example:7373", "127.0.0.1")).toBe(false);
-      expect(isAllowedHostHeader("192.168.1.10:7373", "127.0.0.1")).toBe(false);
+      expect(isAllowedHostHeader("attacker.example:7373")).toBe(false);
     });
 
     it("rejects a missing Host header", () => {
-      expect(isAllowedHostHeader(undefined, "127.0.0.1")).toBe(false);
-      expect(isAllowedHostHeader("", "127.0.0.1")).toBe(false);
+      expect(isAllowedHostHeader(undefined)).toBe(false);
+      expect(isAllowedHostHeader("")).toBe(false);
     });
   });
 
--- a/crates/mcp/src/httpTransport.test.ts
+++ b/crates/mcp/src/httpTransport.test.ts
@@ -195,6 +195,78 @@ describe("session reclamation", () => {
     const tools = await client.listTools();
     expect(tools.tools.map((t) => t.name)).toContain("ping");
   });
+
+  // `isInitializeRequest` validates method and params but not `id`, so an
+  // initialize sent as a JSON-RPC notification passes it. Opening a session for
+  // one strands the slot: the SDK answers 202 with no `Mcp-Session-Id`, so the
+  // client has no handle to use it or to DELETE it, and only the reaper frees
+  // it. Repeat it and the session limit is exhausted by requests that never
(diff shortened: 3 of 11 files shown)


silverstein (commented):
Thank you for this. It is an unusually strong first code contribution (and good to see you again after the Homebrew cask report in #736), and I want to name the specifics rather than hand-wave: you read the installed SDK source to establish the two constraints (one transport per server, no stateless transport reuse) instead of guessing; you treated "binds loopback" as insufficient on its own and added Host/Origin validation with an explicit DNS-rebinding test; and the surface-parity test is exactly the right guard for a factory refactor, since silently dropping a registration is the one failure mode that change could introduce. The PR body flagging its own scariest-looking commit and explaining why it exists is the kind of writing we try to hold ourselves to here.

The five red checks are not your code failing. They are our own guard scripts failing to read it, and you could not have known about them. Two scripts in this repo parse `crates/mcp/src/index.ts` with regexes anchored on `server.resource(` to build derived artifacts: `scripts/generate_llms_txt.mjs` (feeds the Agent Docs llms.txt check and Agents Skill Sync) and `scripts/sync_site_release_version.mjs` (feeds Version Sync and Site Release Link Consistency, and throws the "Found no MCP resources" error you can see in the logs). Your rename of the call sites to `registerResource(` is the whole story; CI Gate is just the aggregate of the others.

Suggested path, all inside this PR: update the regexes in those two scripts to match the new call names (`registerResource(`, and please check the tool patterns too since the same scripts derive the tool surface), then regenerate the derived docs with `node scripts/generate_llms_txt.mjs` and commit the result. Running those two scripts locally reproduces all five failures without waiting on CI. If you would rather not touch our tooling, say so and I will push the script updates to your branch instead, but you are clearly capable of it and it keeps the PR self-contained.

One design question before this merges, and it is the only substantive one I have. `--host` (and `MINUTES_MCP_HOST`) allows binding a non-loopback address, and `isAllowedHostHeader` then accepts the bound address, so a LAN bind gives every device on the network unauthenticated access to the full tool surface: meeting transcripts, recording control, all of it. The warning in the README is honest, but for a privacy-first tool I think the flag invites a mistake that documentation cannot walk back. My preference is to drop `--host` from this PR entirely and ship loopback-only, which also makes your "no authentication" story unconditionally true. If you have a concrete use case for LAN binding, I would rather hear it and design an explicit opt-in (a scarier flag name, or a shared-secret token) as a follow-up than block this PR on it. Which way would you like to go?

Happy to iterate here. stdio-unchanged plus opt-in was exactly the right call, and I would like to land this.

erikcw:
Both done, pushed as two commits.

## The five red checks

Your read of the cause was right, and it was one cause rather than five. `registerResource(` is most of it, plus one shape change the diff summary hides: `registerAppResource(server, ...)` became `registerAppResourceOnEveryServer(...)` and lost its `server` first argument, so swapping the name in the regex would still have missed that one.

I matched both spellings rather than replacing one with the other, so the scripts work whichever call shape a site uses and don't depend on which branch they run against. I also checked that extraction is unchanged rather than merely non-zero: old and new patterns both find the same 11 resources, on `main` and here. Worth checking, because a regex that matches *something* clears the "Found no MCP resources" throw and then fails Version Sync on a count instead, which is a more annoying failure to read.

Four scripts, five checks, since CI Gate is the aggregate as you said. All four reproduce the failure locally and all four pass now: `sync_site_release_version.mjs --check`, `generate_llms_txt.mjs --check`, `check_version_sync.mjs`, `check-agents-sync.sh`.

Two notes on the path you suggested:

- The tool patterns didn't need touching. Both scripts read the tool surface from `manifest.json` (`manifest.tools`) instead of parsing `index.ts`, so the rename couldn't reach them, and `registerTool` already had that name on `main` anyway.
- I didn't commit regenerated docs, because there was nothing to regenerate. Running `node scripts/generate_llms_txt.mjs` changes exactly one thing across the six generated files: the `Last generated:` date. Content is byte-identical, which is why `--check` reports up to date. Committing it would have been six files of churn. Say the word if you'd rather that date move on every PR that touches the generators and I'll include it.

## Dropping `--host`

Going your way. Reading the code again while removing it, the case is a bit stronger than the README warning suggested. The problem isn't only that `--host` opens the port. `isAllowedHostHeader` accepted `hostname === bindHost`, so binding a LAN address also widened the header check itself to trust that address. The check that exists to reject requests arriving under some other name would have quietly been extended to cover the new one.

So `--host` and `MINUTES_MCP_HOST` are gone. `DEFAULT_HTTP_HOST` became `HTTP_BIND_HOST`, since "default" implies an override that no longer exists, and `isAllowedHostHeader` dropped its `bindHost` parameter for a fixed loopback allowlist. Same tests, one fewer degree of freedom, and "no authentication" is now unconditionally true instead of true by default.

Nothing is lost capability-wise, which is what made this easy to agree to. A reverse proxy still gets you LAN access, and it's a better home for it since that's where authentication can go. One gotcha, now in the README: the proxy has to rewrite `Host` to the upstream address or the loopback check rejects it. In Caddy, `header_up Host {upstream_hostport}`. Native MCP clients send no `Origin`, so nothing else is needed for them; a browser client would also need `Origin` handled at the proxy, and that does give up the CSRF check.

On tests, I replaced the "accepts the explicitly bound non-loopback address" case with its inverse, so a non-loopback `Host` being rejected is asserted rather than incidental, and added a `parseTransportConfig` case pinning `--host` and `MINUTES_MCP_HOST` to the unknown-flag path.

Local run: 13/13 tools, copilot contract PASS, 12/12 HTTP, 16/16 parity, 308 vitest across 16 files.

I left the shared process state item alone since you didn't flag it as blocking. Still happy to add an overlapping-tool-execution test if you want it before merge rather than after.

silverstein (commented):
Verified everything you pushed, and it all checks out: `--host` and `MINUTES_MCP_HOST` gone with the header check reduced to a fixed loopback allowlist, both regex spellings matching (including the `registerAppResourceOnEveryServer` shape you caught that my suggestion would have missed), and both guard scripts pass locally on your branch. You were right on both pushbacks too: the tool surface comes from `manifest.json` so those patterns never needed touching, and regeneration would have moved only the date line, so skipping the six-file churn was the correct call. CI is fully green. Your point about `--host` silently widening the header allowlist itself is a sharper argument than the one I gave you, and it is the reason I am glad it is gone entirely.

On your open question: the overlapping-tool-execution test is welcome after merge, no need to hold this PR for it.

One more round before merge, and then I am done asking. We run an adversarial review pass on everything before it lands here, our own work included, and it came back with three findings on the transport. I verified each against your branch rather than relaying them blind:

1. Sessions that end without a DELETE are never reclaimed. `sessions.set` happens in `onsessioninitialized`, and removal only happens via `onsessionclosed` (which the SDK fires on explicit DELETE) or `transport.onclose` (server-side shutdown). A client that crashes, loses its network, or calls the SDK's plain `Client.close()` leaves its entry forever, and the transport spans requests so no socket event fires either. After 16 such disconnects the server 503s every new client while zero are active, and only a restart recovers. Some idle expiry or a reap of dead sessions is the fix; this is the one I would consider blocking, since routine client churn bricks a long-lived server.

2. The 413 for oversized bodies never reaches the client. In the body reader, `req.destroy()` on crossing 4 MiB tears down the socket before the catch block can write the JSON-RPC 413, so callers see a connection reset instead of the documented response. Draining the remainder (or ending the request without destroying the response side) preserves the contract your README states.

3. `--max-sessions 0` is accepted and starts a server that rejects every initialize, since `sessions.size >= 0` is always true. Port 0 is documented as special; the session limit should reject zero with the same style of parse error you added for the other flags.

Same offer as before: if you would rather I push any of these to your branch, say so. Otherwise this merges as soon as the session reclamation lands; 2 and 3 I am happy to take as fast follows if you prefer to keep this PR tight.

erikcw:
Added the overlapping-execution test. Writing it turned up an error in my own description of the shared state in the PR body, so that first, since it makes the remaining risk smaller than I claimed.

## The shared-state list in the PR body overstates it

I listed the copilot subscription set as shared across sessions. It isn't. `registerLiveEventsSubscriptionHandlers` creates `eventSubscriptions` and `copilotSubscriptions` as locals inside the function, it runs per target server through the builder list, and the controller it returns is torn down per session. Subscription state is per-instance, so one session cannot see another's live-event notifications. That was the item with any real privacy weight and it doesn't apply.

What is actually process-global, having gone back through it:

- `cliAvailable` / `cliCheckedAt`, and `MINUTES_BIN`
- the corpus-lease admission counters (`activeCorpusWorkerCount`, the reserved-memory total), and the capability probe result

## The test

`overlapping tool execution from two sessions stays independent`, in `mcp_http_transport_test.mjs`, which is 13 tests there now. It drives the corpus tool from both sessions at once, since `dispatchStableCorpusLease` is the deepest shared path a tool call reaches.

The assertion is exact equality against baselines captured single-session, before the second client connects. That's deliberate. The failure worth catching is a colliding JSON-RPC id handing a plausible response to the wrong session, and an "it returned content" check sails right through that. I confirmed the test can actually fail by mutating one call's tool to simulate the swap: it fails and names the session that got the wrong answer.

The corpus tool is denied on my machine, which makes this more useful rather than less, because the denial path is the branch with the interesting release logic.

## What it doesn't prove, since the repetition looks like it should

Admission is capped at min(MAX_ACTIVE_WATCHERS, maxWatcherCount) = 64 slots, and a concurrent pair costs about 1.2s here. Exhausting the cap would take roughly 33 pairs and 39 seconds even if every single call leaked, and much worse on a machine where the corpus actually resolves. So the three rounds catch state that degrades quickly, not slot exhaustion. That's written into the test comment rather than left for someone to work out.

I did read the release path before building on it: `withStableCorpusLease` releases in a `finally`, with explicit handling for the "worker died before authorization" case and the reasoning commented. Leaks don't look like the likely defect there. Proving it either way wants a unit test against `corpus-lease.ts` and its `beforeSentinelCreate` probe hook, not an HTTP round trip. I can write that separately if you want it, though I don't think it belongs in this PR.

`cliAvailable` poisoning is also untested, where one session's transient CLI failure gates every other session for the cache TTL. That's sharing rather than concurrency, so it needs a different setup to provoke.

## One local note, not about the branch

`src/corpus-lease.test.ts` is red on my machine at the moment under heavy system load (load average 36/85/126 over 1/5/15 min). The 15s authorization deadline blows, which exhausts the process-global memory reservation, and everything after it fails closed. The branch has a zero diff against that file and the suite was green earlier today, so I'm flagging it as an environment observation rather than a regression. Mentioning it in case it's a known flake at your end too.

erikcw:
The red check on 537ab01c is a flake, and I can't clear it myself, so here is the evidence rather than a shrug.

`Agent Corpus + MCP Lockfile (macos-latest)`, one test: `never removes a replacement at a unique live sentinel name`, `AssertionError: expected [ ...(2) ] to include 'REPLACEMENT_MUST_SURVIVE'`, at 5009ms. That number is `DEFAULT_FENCE_TIMEOUT_MS`, so it timed out rather than failed on logic.

Four things say it isn't the commit:

- `ubuntu-latest` and `windows-latest` pass the same job on the same commit. Only the macOS runner went red.
- That job runs `src/index.test.ts src/corpus-lease.test.ts src/secure-read.test.ts src/paths.test.ts`. It never runs `test/*.mjs`, and `test/mcp_http_transport_test.mjs` is the only file the commit touches.
- The branch has a zero diff against `corpus-lease.ts` and `corpus-lease.test.ts`.
- I ran that exact test five times locally, no code change between runs: one failure, four passes.

It also reproduces the same way on my machine under load. The early tests in that file blow the 15s authorization deadline, the timed-out lease defers its release to a termination promise that never settles inside the run, so `reservedCorpusMemoryBytes` never comes back down and twenty-odd later tests fail fast on "meeting corpus retained snapshots exceeded their process budget". Only the unhandled-rejection block at the bottom of the vitest output names the real cause, which is what makes it look like a broad breakage instead of one slow test. Might be worth a longer fence timeout on macOS runners, though that's your call and not something I'd put in this PR.

A re-run should clear it. I don't have rights to trigger one, so it needs you or a fresh push from me, whichever you prefer.

One correction while I'm here, unrelated to CI. A sibling investigation on my side concluded that the corpus tools failing closed was a regression in this branch. It isn't, at least not in this one. Same shell, same PATH, `list_meetings` over stdio: this branch's `dist/index.js` and stock `npx -y minutes-mcp` return a byte-identical denial. `qmd_available` is false on my machine and `qmd` isn't on PATH, which explains it without involving the branch. I'd assumed the environment was at fault, then checked rather than kept assuming, and the check held.

silverstein (commented):
Re-ran it, and thank you for the evidence rather than a shrug. I verified your reasoning independently before triggering it: the branch has a zero diff against `corpus-lease.ts` and `corpus-lease.test.ts` and touches none of the four test files that job runs, so it cannot be the cause.

Your diagnosis is better than anything we had on this, so I have recorded it against our existing tracking item rather than let it evaporate in a PR thread. Two details you could not have known that corroborate it:

- It is not macOS-specific, it is load-sensitive. The same test file failed the same way on `windows-latest` on our own merge to `main` yesterday, while other platforms passed the same commit. Your "ubuntu and windows pass, only macOS went red" observation is the same phenomenon with the platforms permuted.
- The 5009ms-equals-`DEFAULT_FENCE_TIMEOUT_MS` catch is the part that unlocks it. We had been reading this as a broad breakage; a timeout deferring a lease release that never settles, so `reservedCorpusMemoryBytes` never comes back down and twenty-odd later tests fail fast on the budget, explains every symptom we have seen including why only the unhandled-rejection block at the bottom names the real cause. The longer macOS fence timeout is a reasonable mitigation and I agree it does not belong in this PR.

On the correction: noted, and the way you handled it is the part worth saying out loud. Assuming the environment was at fault, then checking rather than continuing to assume, and reporting the check when it cleared your own branch, is exactly the standard we try to hold here. Byte-identical denial against stock `npx -y minutes-mcp` with `qmd` absent from PATH is conclusive.

Where that leaves us: the only thing between this and merge is the session reclamation from my last review, the P1 where a client that crashes or calls plain `Client.close()` leaves its map entry forever and the server 503s everyone after 16 such disconnects. I see `537ab01c` added the overlapping-execution test, which I had said could wait, so I want to check I have not left you guessing about sequencing: that test is genuinely optional for merge, the reclamation is not. The 413 destroy and the `--max-sessions 0` validation I am happy to take as fast follows if you would rather keep this PR tight.

Say the word if you would like me to push the reclamation instead. Otherwise I will watch for it and merge.

silverstein (commented):
All three landed and I verified each against the finding rather than the commit message. The reclamation is the one I cared about, and you went further than the report: idle timeout plus a bounded sweep, GET-stream closes stamping immediately, and keep-alive probing so a vanished peer surfaces as a socket close instead of waiting out the full timeout. `OVERSIZE_DRAIN_BUDGET_BYTES` is the right shape for the 413 too, since draining without a ceiling would just move the denial-of-service one level up. And `parsePositiveInt` with the comment explaining why zero is meaningful for a port but not a count is better than the one-line validation I had in mind.

The Windows red is the same flake, now with the platforms permuted again: `corpus-lease.test.ts`, load-sensitive, nothing to do with your branch. Re-ran it. Your diagnosis is now recorded on our tracking item for it, and it is the best account anyone has produced of a failure we have been living with for weeks, so thank you for that too.

One last finding from the adversarial pass on the new code, and then I think we are done. `isInitializeRequest(body)` is true for an `initialize` sent as a JSON-RPC *notification*, meaning valid method and params but no `id`. That opens a session, and the SDK then answers 202 with no `Mcp-Session-Id`, so the client can never reach or delete it and the slot sits occupied until the reaper runs. A buggy client, or a loop that forgets the id, can exhaust `--max-sessions` in a handful of requests. Requiring a real request with an `id` before `openSession` closes it.

I recognise this is the third round and each one has been my findings rather than your defects, so: this is the last thing I will hold it for, and if you would rather I push it myself and merge, say the word and I will. Otherwise send it up and I will merge on green. Genuinely good work throughout, and the reasoning you have shown in these replies is the reason I have been comfortable reviewing this at the depth I have.

erikcw:
Fixed in `2cd99d4d`, pushed.

You're right, and in one respect it's worse than the report: it doesn't need a buggy client, just a reachable endpoint. Reproduced before changing anything, against `--max-sessions 2`:

```
attempt 0: status=202  mcp-session-id=null  sessionCount=1
attempt 1: status=202  mcp-session-id=null  sessionCount=2
attempt 2: status=503  mcp-session-id=null  sessionCount=2
```

Three requests, not one usable session out of any of them, and the server already refusing everyone.

The fix gates `openSession` on `isJSONRPCRequest(body)` as well. That's the predicate that actually looks at `id`, and it also catches `id: null`, which I checked separately rather than assuming it behaved like a missing `id` — it takes a different route through the schema. `isInitializeRequest` returns true for both:

```
                    isJSONRPCRequest   isInitializeRequest
with id: 1                true                true
no id                     false               true
id: null                  false               true
```

Two tests. The first sends four notification-initializes at a server with `maxSessions: 2`, asserts 400 each time, asserts the count is still zero, then connects a real client to show the limit was never consumed. The 400 on its own wouldn't have caught this, since a session opened and then refused still leaks, so the count is the assertion that matters. The second pins `id: null`. Both fail on the old code, the first with `expected 202 to be 400`.

Local: 17/17 `httpTransport.test.ts`, 11/11 `serverFactory.test.ts`, 13/13 tools, 16/16 parity, 13/13 HTTP, copilot contract PASS, all four guard scripts, `tsc --noEmit` clean.

On the rounds, I wouldn't frame it that way. These were defects in code I wrote and your review found them, which is the outcome I wanted from it. Far rather they turn up here than in somebody's logs.

Merge on green whenever you're ready.

silverstein:
Merged. Verified the gate on `main` before saying so.

The reproduction is the part I want to call out: three requests, no session ID on any of them, server already refusing everyone. I had described it as something a buggy client could do; you showed it is what a correct-looking client does by accident, which is a worse bug than the one I reported. Checking `id: null` separately rather than assuming it behaved like a missing `id`, and putting the three-way table in the comment, is exactly the habit that makes a reviewer trust the rest of the diff. And asserting the session count rather than only the 400 is the right instinct, since a session opened and then refused still leaks.

Thank you for this. An opt-in HTTP transport that leaves stdio byte-for-byte unchanged, a factory refactor with a parity test guarding the one failure mode it could introduce, loopback-only with the header checks that actually stop a browser, and every finding across four rounds closed with a test that fails on the old code. You also fixed a flaky test diagnosis we had been living with for weeks and corrected your own earlier report unprompted. That is a higher standard than most of what lands here.

Two follow-ups on our side, neither yours: the `corpus-lease.test.ts` flake now has a fix in progress that uses your diagnosis, and I will note the HTTP transport in the release notes when the next release goes out.

silverstein:
This shipped in [v0.25.0](https://github.com/silverstein/minutes/releases/tag/v0.25.0), credited to you in the notes.

Also shipped in the same release, from your work on this PR: the `corpus-lease.test.ts` flake fix. Your diagnosis was the whole thing. Once you pointed out that the failure time was exactly `DEFAULT_FENCE_TIMEOUT_MS`, the rest followed, and measuring it confirmed your chain: one stranded child holds about 14.6 MB of a capped process budget, awaiting while it is unreaped clears nothing, awaiting after the reap clears it to zero. So a case now waits for the cleanup it started. That failure had blocked three people on three platforms in two days and had been misread as flaky infrastructure for weeks.

Thanks for the transport, for the four rounds without a single defensive reply, and for fixing a problem that was not yours while you were blocked by it.

Issue #496, opened 2026-07-17
[Epic] First-class voice enrollment (active + passive + on-device)


silverstein:
Tracking issue for making voice enrollment a first-class feature. Design + phasing: PR #495 (`docs/plans/voice-enrollment.md`).

**Problem:** the engine exists (embeddings, `voices.db`, blended profiles, L2/L3 attribution) but there's no product around it — no desktop UI, no MCP tool, no passive learning, no backfill. Surfaced by first-run testing:
- #490 — Recall panel text not copyable
- #491 — configured "Your Name" ignored on solo captures
- #492 — onboarding "test recording" reads like voice enrollment but isn't; no desktop enroll path

**Approach (see PR #495 for the full design + codex adversarial review):**

Phase 1 (next release) — safe foundation + active enrollment:
- [ ] Immutable provenance-rich `voice_samples` schema; one active profile per (person, model); never blend across models (fixes model-version-blind matching)
- [ ] Quality-producing `embed_solo_clip` primitive (16kHz, multi-window, reject inconsistent clips)
- [ ] Matching returns structured evidence + abstains on ambiguity; per-model calibrated thresholds
- [ ] Desktop active enrollment (multi-sample, quality-gated) + status/rename/delete + "delete all voice data"
- [ ] Granular privacy settings (store embeddings / restricted eligibility default off / retention)
- [ ] Onboarding rework (#492) + self-anchoring for solo captures (#491)
- [ ] Ship diarization on-by-default in the desktop bundle (~34MB first-run download)

Phase 2 (fast-follow, gated on an adversarial FAR/FRR eval):
- [ ] Passive candidate capture → explicit promotion (default off, multi-meeting agreement, no restricted/degraded)
- [ ] Confirmed-label backfill from `.embeddings` sidecars (dry-run report; no clustering-to-guess-self)
- [ ] MCP enroll tool; household/multi-user; merge duplicate profiles
- [ ] Adversarial FAR/FRR eval harness as the release gate

**Sequencing:** start after the in-flight conversation-trust work merges to avoid colliding on `commands.rs`/`pipeline.rs`.

silverstein:
Phase 1 progress: the CLI-first slice is merged in #518 (builds on the #505 schema and #506 privacy config).

[...]

PR #759, opened 2026-08-13
fix(attribution): require consistent segment voice evidence

### LINKED ISSUE #753: Voice enrollment can mark a mixed diarization cluster High confidence from centroid-only matching
## Summary

The meeting pipeline can assign a High-confidence enrolled identity to a diarization cluster that contains more than one real speaker.

The pipeline matches each cluster's single averaged embedding against enrolled profiles. If that centroid clears the configured threshold, the entire cluster is stamped `High / Enrollment` and its transcript labels are rewritten. A cluster can clear that threshold when it is dominated by the enrolled speaker even though it also contains substantial speech from somebody else.

This violates the project's standing rule that a wrong rewrite is worse than an anonymous label.

## Sanitized production reproduction

Observed with the released Minutes v0.24.0:

- 78m52s ordinary CPAL WAV
- `diarization_from_stems: false`
- 235 diarization segments
- `raw_diarization_num_speakers: 2`
- effective labels: `SPEAKER_1`, `SPEAKER_2`, and `UNKNOWN`
- one self-enrolled CAM++ v0.3 profile, built from one manual sample
- configured voice-match threshold: `0.65`
- `SPEAKER_1` centroid similarity to that profile: `0.7552`
- `SPEAKER_2` centroid similarity: `0.1310`

The pipeline therefore mapped `SPEAKER_1` to the enrolled profile at `confidence: high`, `source: enrollment`.

After rewriting, that enrolled name appeared on 369 transcript lines. Inspection of alternating dialogue establishes that those lines contain both real participants. Only one line retained `SPEAKER_2`; 57 were `UNKNOWN`.

No private transcript text or audio is included here.

## Current-main verification

Verified against `main` at `28d9704575c16c19465501ec039cfb4649e23cc8`.

`match_speakers_by_voice` in `crates/core/src/pipeline.rs` still:

1. loads legacy centroid profiles with `load_all_with_embeddings`;
2. calls `match_embedding` once per diarized cluster;
3. marks any above-threshold winner `Confidence::High` with `AttributionSource::Enrollment`.

The newer structured `match_active_profiles` path is used by the CLI voice-test surface, not by meeting attribution. Migrating that caller alone would not prove this case safe: with exactly one enrolled profile there is no runner-up, so inter-profile margin cannot reveal that the input cluster itself contains two people.

The diarizer computes embeddings for individual speech segments while clustering, but `DiarizationResult` retains only the final per-label averaged embedding. The segment-level identity evidence is discarded before attribution.

## Relationship to existing work

- Related to #169, but that issue is centered on source/stem-aware local-versus-remote clustering. This reproduction is the ordinary non-stem CPAL path.
- Related to #496, but this is intra-cluster inconsistency rather than ambiguity between enrolled profiles.
- #713 established the important principle that an enrollment match can prove the enrolled person is among the voices without proving that the label contains only that person. That PR was correctly closed: counting speakers in a source stem can misfire for a solo recorder whose microphone captures remote-speaker bleed.
- #714 and #731 improve diagnostics only and explicitly make no attribution behavior change.

## Required behavior

A cluster centroid crossing the enrollment threshold must not, by itself, justify a High-confidence transcript rewrite.

Before applying `High / Enrollment`, meeting attribution needs evidence that the winning identity is consistently supported within the cluster. A likely direction is to preserve or evaluate segment/window-level match evidence before it is averaged away, then abstain or cap confidence when that evidence materially conflicts.

This issue intentionally does not prescribe a fixed supermajority percentage. That threshold needs calibration rather than being inferred from one meeting.

## Acceptance criteria

- A regression fixture represents a non-stem meeting where one diarized label contains substantial speech from an enrolled person and a second person.
- A centroid dominated by the enrolled person cannot cause the mixed label to be rewritten at High confidence.
- Conflicting segment/window evidence leaves the label anonymous or below rewrite confidence.
- A clean, consistently matching cluster can still receive High-confidence enrollment attribution.
- Coverage includes the solo-recorder/open-speaker bleed shape that invalidated #713; the fix must not reduce that case to a naive speaker-count guard.
- The attribution diagnostic records enough aggregate, privacy-safe evidence to explain why a match was accepted or refused.
- Existing one-to-one and model-compatibility work from #496 is composed with, but does not substitute for, the intra-cluster consistency gate.

silverstein:
## Summary

- retain reliable per-segment speaker embeddings through diarization label remapping and stem-result merging
- route meeting attribution through model-scoped active voice profiles and structured match evidence
- require the centroid and every fully covered reliable segment to uniquely support the same enrolled profile before assigning High / Enrollment
- fail closed when a merged speech region is longer than the audio its capped embedding actually evaluated
- make legacy voice deletion revoke migrated immutable samples and clear the active matching cache
- synchronize changed legacy enrollment centroids into the immutable sample architecture
- abstain on missing, conflicting, ambiguous, duplicate-profile, non-finite, or incompletely covered evidence
- add privacy-safe aggregate diagnostics and synthetic regressions for mixed clusters, long merged regions, clean enrollment, model compatibility, deletion, repeat enrollment, duplicate claims, and open-speaker bleed

## Root cause

Meeting attribution used a final cluster centroid as sufficient evidence for a High-confidence enrollment match. A diarization cluster containing two humans could average into an enrolled profile even though the cluster's individual segments did not consistently support that identity.

The first implementation also exposed two migration gaps during release-gate review: capped embeddings could represent longer merged speech regions than they actually heard, and legacy delete/re-enroll operations did not update the new immutable sample path.

This change preserves the individual segment evidence and its exact audio coverage, then composes it with the structured active-profile matcher. It does not add a speaker-count guard or a new percentage threshold.

## Validation

- `cargo test -p minutes-core --features diarize pipeline::tests:: -- --test-threads=1` — 114 passed, 1 ignored
- `cargo test -p minutes-core voice::tests:: -- --test-threads=1` — 29 passed
- `cargo test -p minutes-core --features diarize diarize::tests:: -- --test-threads=1` — 83 passed, 1 ignored
- broad no-default lib suite — 1724 passed, 1 ignored; four worker-harness tests required the documented worker binary prebuild and all four passed independently afterward
- `cargo clippy --all --no-default-features -- -D warnings`
- `cargo clippy -p minutes-core --features diarize -- -D warnings`
- private local reproduction, aggregate-only: old policy accepted one cluster; revised policy produced zero High-confidence attributions across two diarized speakers and 182 reliable segment embeddings; three incompletely covered long merged regions were detected and rejected

All repository fixtures are synthetic. No private audio, transcript, embeddings, names, or paths are included.

Closes #753

### DIFF
--- a/crates/core/src/diarize.rs
+++ b/crates/core/src/diarize.rs
@@ -173,6 +173,16 @@ pub struct SpeakerSegment {
     pub end: f64,
 }
 
+/// One reliable, normalized segment embedding retained for identity safety.
+///
+/// This stays in memory only. Meeting diagnostics summarize counts and
+/// decisions; they never serialize the embedding itself.
+#[derive(Debug, Clone)]
+pub struct SpeakerEmbeddingSegment {
+    pub embedding: Vec<f32>,
+    pub speech_seconds: f64,
+}
+
 #[derive(Debug, Clone)]
 pub struct DiarizationResult {
     pub segments: Vec<SpeakerSegment>,
@@ -191,6 +201,13 @@ pub struct DiarizationResult {
     /// Per-speaker averaged embeddings (for Level 3 confirmed learning).
     /// Empty when using the Python subprocess engine.
     pub speaker_embeddings: std::collections::HashMap<String, Vec<f32>>,
+    /// Reliable segment embeddings grouped by their final speaker label.
+    ///
+    /// A cluster centroid alone cannot establish that every segment in the
+    /// cluster belongs to the same enrolled person (issue #753). Retaining the
+    /// inputs that formed it lets attribution require consistent evidence and
+    /// abstain before a wrong High-confidence rewrite.
+    pub speaker_embedding_segments: std::collections::HashMap<String, Vec<SpeakerEmbeddingSegment>>,
 }
 
 impl Default for DiarizationResult {
@@ -204,6 +221,7 @@ impl Default for DiarizationResult {
             from_stems: false,
             source_aware: false,
             speaker_embeddings: std::collections::HashMap::new(),
+            speaker_embedding_segments: std::collections::HashMap::new(),
         }
     }
 }
@@ -1817,6 +1835,7 @@ fn diarization_from_energy_windows(
             from_stems: true,
             source_aware: true,
             speaker_embeddings: std::collections::HashMap::new(),
+            speaker_embedding_segments: std::collections::HashMap::new(),
         });
     }
 
@@ -1880,6 +1899,7 @@ fn diarization_from_energy_windows(
             from_stems: true,
             source_aware: true,
             speaker_embeddings: std::collections::HashMap::new(),
+            speaker_embedding_segments: std::collections::HashMap::new(),
         })
     }
 }
@@ -2080,6 +2100,17 @@ fn remap_diarization_labels(
         }
     }
 
+    let mut speaker_embedding_segments = std::collections::HashMap::new();
+    let mut evidence_keys: Vec<String> =
+        result.speaker_embedding_segments.keys().cloned().collect();
+    evidence_keys.sort();
+    for raw_label in evidence_keys {
+        let remapped_label = remap_label(&raw_label);
+        if let Some(evidence) = result.speaker_embedding_segments.get(&raw_label) {
+            speaker_embedding_segments.insert(remapped_label, evidence.clone());
+        }
+    }
+
     DiarizationResult {
         segments,
         num_speakers: label_map.len(),
@@ -2089,6 +2120,7 @@ fn remap_diarization_labels(
         from_stems: result.from_stems,
         source_aware: result.source_aware,
         speaker_embeddings,
+        speaker_embedding_segments,
     }
 }
 
@@ -2156,6 +2188,12 @@ fn merge_remote_diarization_into_stem_result(
         .filter(|(label, _)| present_labels.contains(*label))
         .map(|(label, embedding)| (label.clone(), embedding.clone()))
         .collect();
+    let speaker_embedding_segments = remote_result
+        .speaker_embedding_segments
+        .iter()
+        .filter(|(label, _)| present_labels.contains(*label))
+        .map(|(label, evidence)| (label.clone(), evidence.clone()))
+        .collect();
 
     DiarizationResult {
         num_speakers: present_labels.len(),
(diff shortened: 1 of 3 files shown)


silverstein (commented):
Reviewed this as the release gate, since a wrong name is worse than none and this is the last thing I would want shipping unresolved. The design is right and it avoids the trap that killed #713: requiring the centroid *and* every reliable segment to agree, with no speaker-count guard and no new percentage threshold, is exactly the discriminator that separates a genuinely mixed cluster from a solo recorder picking up bleed. Routing through the structured active-profile matcher also finishes the #496 migration rather than bolting a second policy beside it.

Adversarial pass found three, and I verified each against the diff rather than relaying them:

1. [P1] The gate can still pass a mixed cluster, so the reported bug is not fully closed. `merge_short_segments` runs at diarize.rs:3253, *before* embeddings are computed, so `speech_segments` here are already merged regions and can exceed `MAX_EMBEDDING_SECONDS` (60). The embedding is computed from the capped prefix (3274-3277) while `speech_seconds` records the region's full length (3454), so a 120s merged region holding speaker A for 60s and speaker B afterwards yields exactly one evidence item, unanimously A. The consistency check sees no disagreement and assigns High/Enrollment to a region that is half B. That is the original failure with a longer segment, and Mat's repro was a 78-minute two-person interview, which is where long merged regions are most likely. Either retain bounded embeddings across the whole region or withhold High for audio no embedding covers.

2. [P1] Legacy deletion stops working, and this one is a consent problem rather than an accuracy one. Once a profile is migrated and matching reads `voice_active_profiles`, `minutes voices --delete` still calls `delete_profile` (cli/src/main.rs:16948), which only removes the `voice_profiles` row (voice.rs:1382-1386). The command reports success, and the person keeps being identified in later meetings from the migrated biometric copy. For this project that is the worst shape a bug can take: we told someone their voiceprint was gone and it was not. Route that command through `revoke_voice_person`, or remove the migrated samples and cache alongside.

3. [P2] Repeated `minutes enroll` runs stop improving anything. Later enrollments go through `save_profile_blended` and keep the original `enrolled_at`, so `migrate_legacy_profiles` dedupes on the unchanged timestamp and never updates the sample. The CLI reports a higher sample count and promises better accuracy while matching still uses the first centroid.

None of these are reasons to change the approach; the approach is right. 1 and 2 I would fix before merge, 2 especially. 3 can follow.

PR #760, opened 2026-08-14
fix(sdk,mcp): make a refused bound read diagnosable instead of a hang

### LINKED ISSUE #617: Flaky: corpus-lease process budget exceeded on windows-latest in CI

silverstein:
Refs #617. Takes the thread that #758 left open.

## Why this and not a fix

#758 closed a real corpus-lease contamination path. The file still failed on Windows afterwards: run `31753865023`, head `a29f4472`, which `git merge-base --is-ancestor` confirms contains the #758 merge commit. It timed out at 15s in `fails closed when every bounded reader at the cap is active`.

Note the symptom changed. #617 opened on a **budget** error; this is a **timeout**. And the CI log said nothing else, so there is no way to tell from it which resource was exhausted or even which read failed. That is the actual blocker on #617 right now, more than the flake itself.

## Two reasons it was undiagnosable

**The test turns any first-read failure into an infinite hang.** `markStarted` runs inside `afterFirstRead`, so a read refused before that point leaves the gate unresolved and the suite waits out its timeout rather than reporting the rejection. Racing the read against the gate surfaces the real error.

Verified with a throwaway probe rather than asserted: with a first read forced to fail, the old shape hung (confirmed by a 1.5s race against a sentinel) and the new shape rejected with the underlying error. Both cases passed, then the probe was deleted.

**The refusal itself said nothing.** Bound-read admission can be refused by a per-reader in-flight cap, a global in-flight cap, or a global byte reservation, and all three threw the identical sentence. Because those counters are **process-global**, the read that gets refused is usually not the one at fault, which is exactly the situation where the reason matters.

The reason now goes to stderr naming which limit tripped and its current values:

```
[bound-reader] refused: global bytes 402653184 + 33554432 requested > 402653184
```

The thrown message is deliberately unchanged. Callers still must not learn why a read was refused; the detail is operator-visible only. That split follows the denial diagnostics already in `corpus-lease.ts`, including deferring the write with `setImmediate` and never letting a broken stderr escalate a clean refusal into a crash.

## Honest scope

**This does not fix the flake and does not claim to.** It makes the next occurrence say which resource was exhausted instead of leaving another anonymous 15s timeout. Given the flake's rate, that is the difference between diagnosing it on the next hit and waiting for a dozen more.

I would not close #617 on this.

## Verification

SDK 153 passed, MCP 314 passed, both `tsc --noEmit` clean, and the sdk/mcp copies of both changed files are byte-identical.

### DIFF
--- a/crates/mcp/src/corpus-lease.test.ts
+++ b/crates/mcp/src/corpus-lease.test.ts
@@ -132,7 +132,12 @@ describe("stable corpus lease", () => {
             },
           })
         );
-        await didStart;
+        // Racing the read surfaces its rejection. Awaiting `didStart` alone
+        // hangs to the suite timeout whenever the first read is refused,
+        // because `markStarted` runs inside the read that just failed. That
+        // turned a named capacity refusal into an anonymous 15s timeout on
+        // Windows, which is what made #617 undiagnosable from CI logs.
+        await Promise.race([didStart, activeReads[activeReads.length - 1]]);
       }
 
       const thirdParent = join(root, "active-1");
--- a/crates/mcp/src/corpus-lease.ts
+++ b/crates/mcp/src/corpus-lease.ts
@@ -23,6 +23,7 @@ import {
   fingerprintTextFileFromBoundParent,
   readTextFileWithRevisionFromBoundParent,
   type BoundFileRevision,
+  writeOperatorDiagnostic,
 } from "./secure-read.js";
 
 const MAX_AUTHORIZATION_ATTEMPTS = 2;
@@ -2018,7 +2019,11 @@ async function dispatchStableCorpusLease<T>(
             const budget = overran
               ? `authorization budget overrun by ${magnitudeMs}ms`
               : `${magnitudeMs}ms of authorization budget remained`;
-            process.stderr.write(`[corpus-lease] denied: ${message} (${budget})\n`);
+            // Shared with the bound-reader refusal path, because a plain
+            // try/catch here never covered an asynchronous EPIPE: that
+            // arrives as an 'error' event on a later tick and is fatal
+            // without a listener.
+            writeOperatorDiagnostic(`[corpus-lease] denied: ${message} (${budget})\n`);
           } catch {
             // A broken stderr must never turn a clean denial into a crash.
           }
--- a/crates/mcp/src/secure-read.ts
+++ b/crates/mcp/src/secure-read.ts
@@ -78,6 +78,39 @@ export type BoundTextFileRead = {
   revision: BoundFileRevision;
 };
 
+/**
+ * Absorbed once, then never again: an 'error' event on stderr with no listener
+ * is fatal to the host process.
+ */
+let stderrErrorsAbsorbed = false;
+
+/**
+ * Write one operator-visible diagnostic line, without letting a broken stderr
+ * become fatal.
+ *
+ * A try/catch around the write is not sufficient. When the consumer of the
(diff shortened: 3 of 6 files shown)


silverstein:
Pushed after a codex pass, which found that the diagnostic I had just added could **kill the host process** — worse than the missing diagnostic it replaced.

A `try/catch` around `process.stderr.write` does not cover an asynchronous EPIPE. When the consumer of the pipe is gone, the failure arrives as an `'error'` event on a later tick, outside any surrounding catch, and an `'error'` event with no listener is fatal. My comment said a broken stderr "must not escalate a refusal into a failure"; the code did not actually deliver that, which is the same overclaiming the comment was warning against.

Reproduced in both directions rather than argued:

```
old shape (try/catch only):  exit=1
guarded shape (this PR):     SURVIVED, exit=0
```

The fix is the call this repository already makes for a killed worker's stdin (`child.stdin?.on("error", () => {})` at `corpus-lease.ts`, with a comment describing this exact hazard). So rather than hand-rolling a second copy of a pattern that was already subtly wrong, it is now one shared `writeOperatorDiagnostic` helper.

Worth noting: **`corpus-lease.ts`'s existing denial diagnostic had the identical hazard**, in shipped code, not just in my new line. It now routes through the same helper, so that path is fixed too.

Codex confirmed the rest clean, and two of its checks are worth recording because they were the ones I most wanted independently verified:

- **Behavioral equivalence of the admission rewrite.** Turning four OR'd conditions into a ternary chain is exactly where a security admission path could silently change. It compared old and new decisions across 100,000,000 combinations including `NaN`, `Infinity`, `-0`, and non-integers, with zero mismatches.
- **The diagnostic names the right limit.** All four branches print the values actually compared, and when several limits are exceeded the reported one is the first true predicate, matching the old short-circuit order.

One observation it made that I am leaving alone: despite its name, the test proves refusal by the **global byte reservation** rather than by every reader slot being active, because the loop holds a single read. That predates this commit and the race does not change what it proves, but it is worth knowing when someone returns to #617.

SDK 153 passed, MCP 314 passed, both tsc clean, mirror guard reports all six files identical.

Issue #761, opened 2026-08-14
Would you take a pull-request security scan workflow?


ralyodio:
Would a pull-request security scan be useful here, or is this already covered?

One workflow. On each pull request it scans the checked-out repository for
hardcoded credentials, injection, SSRF and unsafe deserialisation, and writes
findings to the Security tab. Report-only — findings never fail the build.

Two files under `.github/`, a pinned `@profullstack/threatcrush@0.11.0` whose tarball is hashed before
install, and `pull_request` rather than `pull_request_target`.

A pull request is open alongside this with the diff, if reading it is easier
than discussing it.

Disclosure: I maintain [ThreatCrush](https://github.com/profullstack/threatcrush);
MIT and free. Written with AI assistance. Closing this is a fine answer and I
will not ask again.

silverstein:
Answering here too, since this is where you asked, and thank you for asking before sending code.

Short answer: not this one, but the question was a good one.

Credential scanning is already in place and acts earlier than a pull-request scan would. This repository has GitHub secret scanning with **push protection** enabled, so a hardcoded credential is refused at push time rather than reported after it has landed. CI additionally runs `scripts/check_signing_secret_scope.py`, which asserts signing secrets stay behind an approved environment gate.

For the part that genuinely is not covered, injection and SSRF and unsafe deserialization, the right answer for this repository is CodeQL: first party, free for public repositories, understands the TypeScript here, and adds nothing new to the dependency surface. Your issue is a reasonable prompt to set that up, and I have written it down as such.

The full reasoning, including what I thought was well done about the implementation, is on #762.

PR #762, opened 2026-08-14
ci: scan pull requests for credentials and injection with ThreatCrush

ralyodio:
Adds one workflow. On each pull request it scans the checked-out repository for
hardcoded credentials, injection, SSRF and unsafe deserialisation, and writes
findings to the Security tab and a comment.

- `.github/workflows/threatcrush-scan.yml`
- `.github/scripts/threatcrush-to-sarif.py` — SARIF shim for older CLI versions

**Report-only.** `failOn` is empty, so findings never fail the build. An install
or scan failure does fail the job: a scanner that reports clean when it did not
run is worse than no scanner.

**Scope:** it scans the whole checked-out repository, not only the diff.

**Supply chain.** Pinned to `@profullstack/threatcrush@0.11.0`; the tarball is hashed and checked against
a value in the workflow before install (`npm view` it yourself), installed with
`--ignore-scripts`, actions pinned to commit SHAs, and it runs on `pull_request`
rather than `pull_request_target`.

Asked first in https://github.com/silverstein/minutes/issues/761.

Disclosure: I maintain [ThreatCrush](https://github.com/profullstack/threatcrush);
MIT and free. Written with AI assistance. Closing this is a fine answer and I
will not send another.

### DIFF
--- a/.github/scripts/threatcrush-to-sarif.py
+++ b/.github/scripts/threatcrush-to-sarif.py
@@ -1,0 +1,235 @@
+#!/usr/bin/env python3
+"""Convert ThreatCrush terminal output to SARIF 2.1.0.
+
+Compatibility shim for CLI versions older than native ``--format sarif``.
+When the CLI can emit SARIF itself the workflow uses that and never runs this
+file; parsing a human-readable stream is strictly worse and exists only so a
+repository is not left unscanned while waiting for a release.
+
+It **fails closed**. If it cannot recognise the output it exits non-zero and
+dumps what it saw. Emitting empty SARIF instead would report "0 findings",
+which is indistinguishable from a clean scan and is the single most expensive
+thing a security tool can get wrong.
+
+Three details of the format, each of which is load-bearing:
+
+* Severity is bare for ``CRITICAL`` and bracketed for ``[HIGH]``/``[MEDIUM]``/
+  ``[LOW]``. One regex shape misses half the findings.
+* ``File:`` paths are relative to the scan root, not the repository root. Left
+  unprefixed, every finding resolves to nothing in the consumer's view of the
+  repo. Hence ``--path-prefix``.
+* Whole-file findings report line ``:0``. SARIF requires ``startLine >= 1``.
+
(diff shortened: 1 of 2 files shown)


silverstein:
Thank you for this, and genuinely for how you sent it. You asked first in #761, you disclosed that you maintain ThreatCrush without being asked, and you said up front that closing it is a fine answer. That is a better process than most unsolicited contributions follow, and the workflow itself shows care: report-only, tarball hash pinned, `--ignore-scripts`, actions pinned to SHAs, `pull_request` rather than `pull_request_target`, and failing loudly when the scan does not run instead of reporting clean. The line about a scanner that reports clean when it did not run being worse than no scanner is exactly right, and it is the same principle several of this repo's own guards are built on.

I am going to decline, and I want to give you the actual reasons rather than a polite non-answer.

**The credential half is already covered, and covered earlier in the lifecycle.** This repository has GitHub secret scanning enabled with **push protection**, so a hardcoded credential is blocked at push time rather than reported after it has landed in a branch. CI also runs `scripts/check_signing_secret_scope.py`, which asserts that signing secrets stay behind an approved environment gate. A scan that reports a credential on a pull request would, for this repo, mostly be telling us about something push protection should already have refused.

**For the half that is not covered, injection and SSRF and unsafe deserialization, the better fit is CodeQL.** It is first party, free for public repositories, understands the TypeScript here, and adds no new dependency to CI. If we want static analysis on pull requests, that is the version we should adopt, and your issue is a fair prompt to do it.

**And the cost is a standing third-party dependency in CI.** The package is four months old, and the workflow would run on every pull request indefinitely. Your pinning and hash check meaningfully reduce that risk rather than hand-waving it, but they do not remove it, and this project has spent a lot of effort recently reducing what its release and CI paths depend on.

None of that is a judgment about ThreatCrush. It is about this repository already having the cheaper answer for one half and a first-party answer available for the other.

Closing per your own suggestion. #761 gets the same answer so the question is answered where you asked it. Thanks again for asking first; it made this an easy conversation to have honestly.

PR #763, opened 2026-08-14
release: bump version to 0.25.0

silverstein:
Step 1 of the release, plus the fix that made it possible to run at all.

## The bump script was broken, and would have blocked the release

`node scripts/bump-version.mjs --dry-run 0.25.0` failed outright:

```
bump-version: cargo update changed Cargo.lock beyond the version fields
for minutes-core, minutes-cli, and minutes-reader
```

The script hardcoded those three as the packages whose `Cargo.lock` version tracks the workspace version. The archive crates arrived later and inherit that version too, so `cargo update` moves **eight** packages while the expectation still described three. The equality check then failed and no version bump could run.

Worth noting how it presented: the error names a `Cargo.lock` problem, and there is nothing wrong with `Cargo.lock`. Reproducing it took three attempts, because the failure only appears when both manifests are bumped first, which is what the script does internally and what a manual `cargo update` does not.

Now derived from `version.workspace = true` in each workspace member, which is exactly the property that makes a member track the workspace version. The eight it selects match the eight `cargo update` actually moves, and `whisper-guard` plus the two Tauri apps are correctly left on their independent versions. The `cargo update -p` argument list comes from the same source, so the two can no longer disagree.

This is the same shape as the CI rust filter and the tap cask: a hand-maintained list that silently stops describing reality, and only bites at release time.

## The bump

132 pull requests since v0.24.0 on 2026-07-27. All Domain 1 sources move together, verified individually after applying:

| source | value |
|---|---|
| workspace `Cargo.toml` | 0.25.0 |
| `crates/cli` minutes-core pin | 0.25.0 |
| `tauri.conf.json` | 0.25.0 |
| `crates/mcp/package.json` | 0.25.0 |
| `crates/sdk/package.json` | 0.25.0 |
| `manifest.json` | 0.25.0 |
| `MCP_SERVER_VERSION` | 0.25.0 |

Domain 2 (plugin lockstep) stays at 0.21.0, untouched.

`node scripts/check_version_sync.mjs` passes. `--release` mode still reports the `minutes-sdk` dependency at 0.24.0, which is correct at this point: pinning it is **Phase 2**, after this is on main and CI is green.

## Not done here, deliberately

No tag, no draft release, no registry publish. `minutes:1` is holding the release until the dictation work in `minutes:0` finishes, and I agree with that hold. This commit does not force it: more PRs can merge after a bump without invalidating it, since the version strings stay correct. The step that must come last is the tag.

The generated site docs and `llms.txt` move with the version, which is why the diff is wider than the version files themselves.


silverstein:
Verified a release-day assumption while waiting on CI, since the Homebrew automation has never run for a real release and this will be its first.

`Update Homebrew Tap` triggers on `release: published`, and the procedure creates the release as a **draft** first, attaches assets, then un-drafts it. Whether that un-drafting counts as "published" decides whether the tap updates at all. GitHub's docs are explicit in both directions:

> Workflows are not triggered for the `created`, `edited`, or `deleted` activity types for draft releases.

> The `prereleased` type will not trigger for pre-releases published from draft releases, but the `published` type will trigger.

So both halves hold. Nothing fires while the release is still a draft, which matters because the cask hashes a DMG that does not exist yet at that point, and un-drafting does fire, which is the moment the assets are attached and the hash is computable.

Expect one manual step on release day: the job runs in the `homebrew-tap` environment, which now has required reviewers, so it will pause for an approval before touching the tap.

Source: [Events that trigger workflows](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows)

PR #791, opened 2026-08-18
fix(knowledge): stop blocking agent tools when no QMD registration exists (#788)

### LINKED ISSUE #788: MCP tools blocked with "readiness could not be verified safely" — qmd cleanup fails because qmd is unavailable and persistence is disabled anyway
**Environment**

Minutes desktop app: v0.25.1
Minutes CLI: v0.25.1 (symlinked to the app-bundled binary at /Applications/Minutes.app/Contents/MacOS/minutes)
macOS, Apple Silicon (M-series MacBook Air)
Claude Desktop, connected via the minutes-mcp MCP server

**What happened**

Almost every MCP tool call (list_meetings, search_meetings, get_meeting_insights, track_commitments, list_processing_jobs, etc.) fails with:

Run `minutes qmd cleanup`, repair or reinstall qmd if requested, then restart Minutes.

Following that instruction and running the CLI command directly:

$ /Users/xxx/.local/bin/minutes qmd cleanup

produces:

Persistent QMD indexing is disabled; Minutes search uses a process-private live projection.
Legacy QMD registry cleanup could not be attested because qmd is unavailable.
{
  "schema_version": 2,
  "persistence_enabled": false,
  "cleanup_only": true,
  "cleanup_confirmed": false,
  "qmd_available": false,
  "output_dir": "/Users/xxx/Documents/Meetings",
  "policy_mirror": "/Users/xxx/.minutes/qmd-policy-mirror",
  "target_collection": "minutes",
  "registered": false,
  "physically_registered": false,
  "attested": false,
  "matching_collections": [],
  "unsafe_registrations": [],
  "preserved_uninspectable_registrations": [],
  "config_engine": "builtin",
  "config_collection": null
}
Error: Persistent QMD indexing is disabled, but legacy cleanup could not be fully confirmed

**The core issue**

persistence_enabled is false and qmd_available is false — QMD isn't installed and isn't even configured to be used. But the app still requires the legacy-cleanup confirmation step to succeed before it will let MCP tools proceed, and that confirmation can never succeed if qmd itself can't run. This traps users who never opted into QMD in a state where core MCP functionality is unavailable.

This may be related to the broader class of QMD reliability issues on macOS Apple Silicon (Metal GPU cleanup causing non-zero exit codes even when the underlying operation succeeded) — see tobi/qmd#368 and tobi/qmd#674 — though in this case qmd isn't installed at all (qmd_available: false), so it looks more like a logic gap in how Minutes handles the "QMD not present" case.

**Steps to reproduce**

Have a Minutes install where QMD was never set up / isn't installed.
Update Minutes app to v0.25.1 (and sync the CLI symlink to match, since the CLI on PATH isn't auto-updated by the app).
Restart Minutes and Claude Desktop.
Call any MCP meeting-retrieval tool → fails with the readiness message above.
Run minutes qmd cleanup as instructed → fails with the error above, qmd_available: false.

**Expected behavior**

If persistence_enabled is false and QMD was never configured/installed, the readiness check should not block core MCP functionality on a QMD cleanup confirmation. At minimum, minutes qmd cleanup should be able to self-resolve (e.g. "no legacy QMD registration found, nothing to clean up, proceeding") when there's genuinely nothing to attest.

**Workaround found**

None via the CLI/MCP surface. As a temporary workaround I'm reading meeting markdown files directly from the output directory (~/Documents/Meetings/) instead of using the MCP tools.

silverstein:
Fixes #788.

## The bug

Agent trust readiness failed closed on any QMD retirement audit failure, including on machines that had never registered a collection. A reporter who never opted into QMD lost every content-bearing MCP tool, and the remediation we printed ran the same audit and failed the same way, so there was no route out.

The blanket block was deliberate, on the reasoning that Minutes cannot prove a negative from its own state. What makes it indefensible in practice is that the audit also fails closed on a QMD that is installed but broken: a native-binding load failure exits non-zero, so `qmd collection list` refuses to answer and `minutes qmd cleanup` bails with "QMD registry could not be safely listed". Both states were reproduced on a real machine, which also disproved the earlier advice on the issue that installing qmd unblocks you.

## The rule now

Readiness passes only when both of these hold:

- no qmd command ever came back successful, so the registry told us nothing at all, and
- nothing on the machine says Minutes ever registered a collection, meaning no ownership marker, no policy mirror, no leftover mirror transaction directory, and no QMD reference in config.

Everything else still fails closed, unchanged. A registry that answers and leaves a collection we cannot inspect still blocks. A parseable-but-surprising listing still blocks. Any one of the four evidence signals still blocks.

The pending retirement marker is still written on the passing path. Readiness is recomputed on every agent call, so a qmd that later appears or starts answering is audited again and can still block. This defers an unanswerable question rather than answering it.

## What review found, in order

The first version of this fix was wrong twice and incomplete twice more. Recording that here because the shape of the mistakes is the useful part: each one traded the reported failure for a quieter one in the opposite direction.

**Evidence was read after the retraction.** The retraction deletes the mirror directories on its way through, so the question was being asked of a machine we had already cleaned. Caught by an existing test.

**The gate keyed on "the audit returned Err".** That also let through a working qmd holding a collection we could not inspect, which is exactly the exposure the block exists for. Caught by an existing test.

**"Unreachable" was decided from the last failure rather than the whole run.** The retraction re-audits after its removals and any of those commands can fail with an I/O error, so a registry that listed fine and then broke looked identical to one never installed. On a machine whose marker had been deleted, that reported ready while a real Minutes-owned collection could still hold the meeting text. It now tracks the positive fact: `registry_answered` is set the first time any qmd command succeeds, and once it has, everything after fails closed.

**The evidence scan could answer "clean" without having looked.** It read a bounded 64-entry prefix of the parent directory and treated a directory it could not open as clean. readdir order is arbitrary, so whether leftover mirror plaintext counted as evidence depended on the order the filesystem returned entries in. It now walks to a bound far above any plausible `~/.minutes` and answers "there is residue" for every case where it could not finish: unopenable directory, unreadable entry, or more entries than it will walk.

**The CLI and the gate disagreed about the same machine.** `minutes qmd cleanup` computed its own weaker version of the evidence question and never looked at the ownership marker, so with no qmd and a surviving marker it reported a confirmed cleanup while readiness went on blocking on that same marker. That is the #788 failure shape one layer up: told your remediation worked, still blocked. Both now read one snapshot, taken before the retraction starts deleting the things that constitute the evidence.

## Tests

Existing tests that encoded the old policy are updated with the rationale written into the test body rather than quietly deleted.

- `agent_startup_blocks_missing_and_transient_qmd_spawn_failures_without_local_evidence` becomes `agent_startup_blocks_unusable_qmd_only_when_a_registration_may_exist` and asserts both directions.
- `agent_readiness_revalidation_does_not_reuse_a_prior_ready_result` keeps its intent but drives the second call with a malformed registry rather than an absent one, so it tests recomputation instead of the retirement policy.
- The CLI status test asserted that cleanup fails with no qmd on PATH, which is the #788 symptom itself. It now asserts cleanup succeeds there, and the unconditional refusal to register moved to the register path.

New, each mutation-verified rather than merely green:

- `a_mirror_or_configured_collection_also_counts_as_registration_evidence`
- `a_registry_that_answered_once_never_counts_as_unreachable`
- `residue_evidence_answers_yes_when_it_cannot_finish_looking`
- `an_unreadable_mirror_parent_counts_as_residue`
- `a_retraction_reports_the_evidence_it_saw_before_it_started`
- `qmd_cleanup_is_not_confirmed_while_an_ownership_marker_survives`

One note on that: the first draft of the residue test created 65 entries and passed under the mutation it was written to catch, because whether it caught anything depended on readdir order, which is the defect it is testing for. It now asserts on the real bound.

## Message fixes

The remediation text no longer instructs a command that cannot succeed. It names the precondition, that qmd has to run, before naming the command. The CLI no longer reports a strong "cleanup is confirmed" when the confirmation came from absence rather than from reading the registry.

## Verification

- qmd absent: `agent-readiness --json` returns `ready: true`. Previously `false`.
- qmd installed but failing to load its native bindings: `ready: true`. Previously `false`.
- Core knowledge suite 169 passed, CLI 115 passed, `cargo fmt --all --check` and `cargo clippy --all --no-default-features -- -D warnings` clean.
- The `index.ts` change is a single string literal, typechecked by the MCP Server job rather than locally, since that worktree has no `node_modules`.

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -8145,14 +8145,21 @@ fn qmd_cleanup_status_report(collection: &str, config: &Config) -> Result<QmdSta
         std::fs::symlink_metadata(&report.policy_mirror),
         Err(error) if error.kind() == std::io::ErrorKind::NotFound
     );
-    // `qmd_available` is deliberately NOT required here. When qmd is not
-    // installed there is no registry to inspect and nothing that could serve a
-    // stale registration, so demanding an inspection that cannot happen left
-    // users who never opted into QMD permanently blocked, with the remediation
-    // we printed failing for the same reason (#788). The retraction itself
-    // still has to succeed, and every other condition below still applies.
+    // `qmd_available` is deliberately NOT required here. When the registry
+    // never answered there is nothing that could serve a stale registration,
+    // so demanding an inspection that cannot happen left users who never opted
+    // into QMD permanently blocked, with the remediation we printed failing for
+    // the same reason (#788).
+    //
+    // `registration_evidence` is required to be false alongside it, and comes
+    // from the same snapshot agent readiness uses. Without that this reports a
+    // confirmed cleanup on a machine whose ownership marker readiness is still
+    // blocking on, which tells the user their remediation worked while they
+    // stay blocked. The retraction itself still has to succeed, and every other
+    // condition below still applies.
     report.cleanup_confirmed = (persistent_cleanup.confirmed
-        || persistent_cleanup.registry_never_answered)
+        || (persistent_cleanup.registry_never_answered
+            && !persistent_cleanup.registration_evidence))
         && !report.registered
         && !report.physically_registered
         && report.matching_collections.is_empty()
@@ -9345,6 +9352,43 @@ life (qmd://life/)
             assert!(error.contains("Persistent QMD collections are disabled"));
             assert!(!error.contains("engine ="));
             assert!(!error.contains("qmd_collection"));
+
+            if let Some(path) = original_path {
+                std::env::set_var("PATH", path);
+            } else {
+                std::env::remove_var("PATH");
+            }
+        });
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn qmd_cleanup_is_not_confirmed_while_an_ownership_marker_survives() {
+        // Counterpart to the no-qmd case above. With no qmd on PATH but an
+        // ownership marker still on disk, agent readiness blocks, so cleanup
+        // must not tell the user it succeeded. Reporting confirmed here sends
+        // someone away believing they are unblocked when they are not.
+        with_temp_home(|dir| {
+            let empty_bin = dir.join("empty-bin");
+            std::fs::create_dir_all(&empty_bin).unwrap();
+            let original_path = std::env::var_os("PATH");
+            std::env::set_var("PATH", &empty_bin);
+
+            let config = Config::default();
+            config.ensure_dirs().unwrap();
+            std::fs::write(
+                minutes_core::knowledge::qmd_owned_target_path_for_tests(),
+                br#"{"schema":1,"target":"minutes"}"#,
+            )
+            .unwrap();
+
+            let report = qmd_cleanup_status_report("minutes", &config).unwrap();
+            assert!(!report.qmd_available);
+            assert!(
+                !report.cleanup_confirmed,
+                "a surviving ownership marker must not report a confirmed cleanup"
+            );
+            assert!(cmd_qmd("cleanup", "minutes").is_err());
 
             if let Some(path) = original_path {
                 std::env::set_var("PATH", path);
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -8135,7 +8135,8 @@ fn qmd_status_report(
 
 fn qmd_cleanup_status_report(collection: &str, config: &Config) -> Result<QmdStatusReport> {
     config.ensure_dirs()?;
-    let persistent_cleanup = minutes_core::knowledge::ensure_qmd_persistence_disabled(config);
+    let persistent_cleanup =
+        minutes_core::knowledge::ensure_qmd_persistence_disabled_status(config);
     // The configured/persisted target is retracted by the core transaction.
     // Audit the explicitly requested legacy name as well, removing it if it
     // was created by an older CLI configuration.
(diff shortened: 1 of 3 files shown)


silverstein:
Follow-up commit, from reviewing my own gate before asking anyone else to.

The pass condition was "the last qmd failure was a spawn-level I/O error", which is not the same thing as "qmd told us nothing". The retraction re-audits the registry after its removals, and any of those commands can fail with an I/O error of its own, so a registry that listed successfully and then broke looked identical to one that was never installed. On a machine whose ownership marker and mirror had been deleted, that reported ready while a real Minutes-owned collection could still hold the meeting text. That is precisely the exposure the block exists for, so the first commit traded one wrong answer for another.

It now tracks the positive fact: `registry_answered` is set the first time any qmd command returns successful, and the pass requires that it never was. A registry that answers even once keeps failing closed for the rest of the run. That also removed the hand-rolled flag on a non-success listing, since an unsuccessful command never marks the registry as having answered in the first place.

New test `a_registry_that_answered_once_never_counts_as_unreachable` lists successfully once and then returns NotFound for everything after. It is mutation-verified rather than just green: reintroducing the old "any I/O error means unreachable" rule makes it fail, and removing that rule makes it pass again.

Both #788 states re-verified on a real machine after the change, qmd absent and qmd installed but failing to load its native bindings. Both report ready.

Issue #792, opened 2026-08-18
macOS dual-source call capture: remote-audio stem lost 5 different ways (zero-stuffing, 7936-byte finalization, 0.25.1 regressions)


MrFreekour:
## Summary

On macOS, dual-source call capture loses the remote-audio (system) stem in several
distinct ways. Across five recordings on two app versions, **not one produced a usable
remote-audio track**, while the microphone stem was fine every time.

Environment:
- macOS 26.6.1 (25G76), Apple Silicon
- Minutes.app 0.24.0 and 0.25.1
- CLI 0.18.2 and 0.25.1
- `capture_backend = "core-audio-tap"`, `[recording.sources] call = "auto"`, `allow_degraded_call_capture = false`

Positive note first: `core-audio-tap` itself works well. I verified it captures on built-in
speakers, an HDMI display, and a virtual device, and it keeps capturing across a mid-recording
Bluetooth headphone connection. It also correctly taps pre-hardware-volume on built-in output.
The problems are all downstream of capture.

---

## Bug 1: system stem written as 7,936 bytes at finalization (native call path)

The remote stem is captured correctly for the entire session, then discarded on stop.

**Repro:** start a recording with call intent while a Teams call is active, run ~30 min, stop.

**During recording**, `~/.minutes/native-captures/<id>.system.wav` (`pcm_f32le`, 48 kHz, mono)
grows normally. I sampled the trailing 30 s every ~100 s for 32 min, decoding to s16le/16 kHz:

| Elapsed | rms | nonzero | distinct 16-bit values |
|---|---|---|---|
| 241 s | 0.07316 | 99.7% | 22,431 |
| 843 s | 0.06751 | 99.7% | 21,677 |
| 1345 s | 0.06858 | 99.7% | 21,594 |
| 1784 s | 0.07019 | — | 21,403 |

15 consecutive healthy samples. The paired `.voice.wav` was 48 kHz in one run and 24 kHz in
another (following the input device), and Pearson correlation between the two stems was
`-0.0000 / -0.0013 / +0.0010`, confirming two independent sources.

**After stop:**
```
jobs/<id>.system.wav        7,936 bytes   = 0.020 s
jobs/<id>.voice.wav   180,342,016 bytes   (intact)
```
The `native-captures/` source is deleted, so the audio is unrecoverable. `job.json` sets
`audio_path` to `.voice.wav`, so only the microphone is transcribed.

Minutes detects this itself and downgrades it to an inferred warning:
```json
[...]

PR #794, opened 2026-08-18
fix(capture): emit whole 100ms chunks from the Core Audio tap (#792)

silverstein:
Fixes bugs 2 and 3 of #792. They are one bug.

## Root cause

A consumer treats one `AudioChunk` as one 100 ms slot however many samples it actually carries. `padded_slot` in `capture.rs` builds a fixed 1600 sample slot and copies the chunk into the front of it:

```rust
fn padded_slot(samples: Option<Vec<f32>>) -> Vec<f32> {
    let mut padded = vec![0.0f32; DUAL_SOURCE_SLOT_SAMPLES];
    ...
}
```

So a short chunk is not a short slot. It is a full slot mostly filled with silence, and the file advances by a full slot regardless. The microphone path never trips this because `streaming.rs` emits exactly 1600 samples per chunk, which is the contract the dual-source writer is built around and which its doc comment states.

The Core Audio tap sent whatever one callback produced. A 512 frame callback at 48 kHz resamples to about 171 samples, or 10.7% of a slot.

## Why this is the cause and not a plausible story

That predicts a stem which is 10.7% nonzero and inflated 9.4x. The report measured 7.9% to 10.7% nonzero, and inflation of 4.9x to 8.9x varying by device buffer size. The reporter independently noticed the two numbers are reciprocals of each other, which is what one cause looks like observed from two directions.

## The fix

The tap buffers resampled output and emits only whole chunks, keeping any partial tail for the next callback. The split is a plain function over a buffer, `drain_whole_chunks`, so the invariant is testable without a live tap.

Three tests: short callbacks must accumulate rather than emit, emitted chunks must carry their samples unchanged and in order, and a zero size must not loop forever. Mutation-verified by restoring the old behavior of emitting whatever arrived, which fails two of the three.

## What this does not explain

One recording in the report had a correct duration header and was still zero stuffed. That row is not accounted for by this fix, and the issue should stay open on it. It is also the row that makes duration useless as a health signal, which the reporter called out.

## Verification

- Confirmed the changed code is actually compiled rather than sitting behind an inactive `cfg`, by introducing a deliberate typo in the edited block and checking that it broke the build.
- `cargo clippy --all --no-default-features -- -D warnings` clean, `cargo fmt --all --check` clean.
- Not verified end to end against a live tap. That needs a real dual-source capture, and the reporter has scripted repros.

### DIFF
--- a/crates/core/src/system_audio_backend.rs
+++ b/crates/core/src/system_audio_backend.rs
@@ -445,7 +445,41 @@ struct CoreAudioTapCallbackContext {
     interleaved: bool,
     resample_pos: f64,
     input_samples: Vec<f32>,
+    /// Resampled samples not yet emitted, held until a whole chunk is ready.
+    ///
+    /// Consumers treat one AudioChunk as one 100 ms slot regardless of how
+    /// many samples it carries, so emitting a short chunk does not shorten the
+    /// slot, it zero-fills the rest of it and advances time by a full slot
+    /// anyway. A Core Audio callback delivering 512 frames at 48 kHz yields
+    /// about 171 samples, which is a slot that is 89% silence and 9.4x longer
+    /// than the audio in it (issue #792, bugs 2 and 3).
+    pending_output: Vec<f32>,
     chunk_index: u64,
+}
+
+/// Samples in one emitted chunk: 100 ms at 16 kHz.
+///
+/// This matches the microphone path in `streaming.rs`, which is the contract
+/// the dual-source slot writer is built around.
+#[cfg(feature = "streaming")]
+const TAP_CHUNK_SAMPLES: usize = 1600;
+
+/// Take whole `size`-sample chunks off the front of `pending`, leaving any
+/// partial tail behind for the next callback to complete.
+///
+/// Returning the tail rather than padding it is the whole point. A consumer
+/// treats one chunk as one slot however many samples it holds, so a short
+/// chunk is not a short slot, it is a full slot mostly filled with silence.
+#[cfg(feature = "streaming")]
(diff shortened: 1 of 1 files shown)


silverstein:
Closing this in favor of #797, which fixes the same bug better. Apologies to @MrFreekour for having a competing patch open at all: I pushed this before yours appeared and should have said on the issue that I was writing it, so you could have skipped the duplicate work.

Three things yours does that this one does not:

It removes the duplicate constant instead of adding another one. This PR introduced a second chunk-size constant next to the tap, which leaves exactly the same trap for the next producer. #797 makes `DUAL_SOURCE_SLOT_SAMPLES` derive from `streaming::CHUNK_SAMPLES`, so the mixer slot and the producer contract cannot drift apart by construction.

It gives producers something to use rather than something to remember. `ChunkAccumulator` is shared, documented at the point of use, and the existing microphone path is refactored onto it, so the hand-rolled chunking that the tap failed to copy no longer exists to be copied wrongly.

It found the reason none of this was caught. The unit test step runs `--no-default-features`, so every `#[cfg(test)]` module inside a streaming-gated file was compiled and never executed. That is why a chunking invariant could break without a red build, and it also means the tests I added in this PR would have been dead on arrival. Adding the streaming test step is worth more than either fix.

The measurement in your header dump also confirms the arithmetic independently: a 4,096 byte aligned header plus 960 frames at 4 bytes is exactly the 7,936 bytes you measured, and 960 frames at 48 kHz is the 0.020 s you reported.

#793 stays open and does not overlap: it stops the finished stem being truncated in the first place, in the Swift helper, which is upstream of the reporting in #798.

PR #795, opened 2026-08-18
feat(capture): report a source that stops delivering mid-recording (#792)

silverstein:
Reports bug 6 of #792. It does not fix it.

## The problem

`padded_slot` fills any slot a source did not supply, so a source that stops delivering is written as silence, slot after slot. The stem comes out the right length and at exact digital zero, and nothing complains.

That is how bug 6 reached its reporter. The microphone delivered normally, a default input device change killed it three minutes in, it never recovered, and the only signal was a stem of pure digital silence. Their measurement at the time, while the stem read exact zero:

```
built-in mic  (avfoundation :2): rms 0.002839, 97.7% nonzero
Bluetooth mic (avfoundation :0): rms 0.000242
```

Both devices were delivering audio. Ours was not recording it, and reported success.

## Why the existing counters miss it

The mixer already counted slots per source and logged the totals. Totals cannot see this failure, because most of that recording really was fine: 600 healthy slots followed by 600 dead ones still reads as a mostly-healthy recording.

What is diagnostic is the longest unbroken run with nothing from a source. `DualSlotStats` now tracks that and names any source absent for 30 seconds or more, on stderr and in the structured log.

Absence is not quietness. A quiet microphone still delivers chunks, they just carry low-amplitude samples, so these counters only move when a source produces nothing at all. That distinction is what keeps this from firing on people who simply do not talk much.

## Scope

This makes the failure loud. It does not change the reconnect behavior that lets the microphone die after a device change, so bug 6 stays open. Splitting it this way means the diagnosis ships even if the reconnect fix takes longer, and anyone hitting it gets told rather than discovering it in a transcript.

## Tests

Four, each mutation-verified rather than merely green:

- a source dying partway is caught even though the totals look healthy, which is exactly the case the existing counters miss
- brief absences that recover are not faults however many there are, so long recordings do not cry wolf
- the threshold is the boundary it claims to be, tested at and just under
- both sources can fault and each is named

Verified by reintroducing cumulative counting, which the recovery test kills, and by moving the threshold off by one, which two tests kill.

### DIFF
--- a/crates/core/src/capture.rs
+++ b/crates/core/src/capture.rs
@@ -1119,12 +1119,73 @@ fn dual_source_slot_for_chunk(base_slot: u64, chunk: &crate::streaming::AudioChu
     base_slot + chunk.index
 }
 
+/// Consecutive slots a source may be absent for before we call it a fault
+/// rather than jitter. 300 slots is 30 seconds.
+///
+/// A source that stops delivering is not the same as a source that is quiet.
+/// A quiet microphone still delivers chunks, they just contain low-amplitude
+/// samples, so these counters move only when a source produces nothing at all.
+#[cfg(feature = "streaming")]
+const SOURCE_GAP_FAULT_SLOTS: u64 = 300;
+
 #[cfg(feature = "streaming")]
 #[derive(Default)]
 struct DualSlotStats {
     both: u64,
     voice_only: u64,
     system_only: u64,
+    voice_gap: u64,
+    voice_gap_longest: u64,
+    system_gap: u64,
+    system_gap_longest: u64,
+}
+
+#[cfg(feature = "streaming")]
+impl DualSlotStats {
+    /// Record one slot and track how long each source has been absent.
+    ///
+    /// The totals alone cannot see the failure in issue #792 bug 6, where the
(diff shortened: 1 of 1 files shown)


silverstein:
Note for whoever merges this after #797: the tests here live in a `#[cfg(all(test, feature = "streaming"))]` module, and until #797 lands they do not execute in CI at all. `default = ["whisper"]`, so the existing `cargo test -p minutes-core --no-default-features --lib` step leaves `streaming` off. Verified directly on main: filtering that command to `streaming::` reports `0 passed; 1763 filtered out`.

So this PR is green today partly because its tests never ran. They pass locally with the feature on, and #797 adds the CI step that will actually execute them. Worth merging #797 first so this one is genuinely covered rather than nominally green.

silverstein:
Merging this ahead of #797 after all, reversing what I said above. #797 is blocked on an unrelated pre-existing Windows failure it happens to have surfaced (#800), and holding a working fix behind that helps nobody. The caveat stands: until #797 lands, the tests in this PR compile and do not execute in CI, so their value here is local and on review rather than enforced. Once #797 is in, they start running with no further change.

PR #797, opened 2026-08-18
fix(core): chunk system-audio tap output to whole mixer slots (#792)

MrFreekour:
Fixes the zero-stuffed system stem reported in #792 (Bug 2 there), and the duration inflation filed alongside it as Bug 3, which turns out to be the same defect.

## Root cause

`AudioStream::start` accumulates microphone samples and emits an `AudioChunk` only once it holds exactly 1600 of them. The Core Audio process tap does not: `CoreAudioTapCallbackContext::ingest` sends one `AudioChunk` per IO callback, carrying however many samples that callback produced.

A tap on a 48 kHz output device delivers 512 frames per callback, which is ~171 samples once decimated to 16 kHz.

Dual-source capture keys its mixer slots off `AudioChunk::index` and pads every slot out to `DUAL_SOURCE_SLOT_SAMPLES` (1600) via `padded_slot`. So each ~171-sample tap chunk claimed a whole slot and was zero-filled for the remaining ~1429 samples, and the system stream's slot counter advanced ~9.4x faster than the microphone's.

That single mismatch produces both reported symptoms:

- **Zero-stuffing.** The system stem measures 8-11% nonzero samples where ~99.7% is expected. Extracting only the nonzero samples and replaying them reconstructs intelligible speech, so the audio was captured and then diluted on write. Decoded as written, the remote side of a call transcribes as `[tapping]` / `[MUSIC PLAYING]`.
- **Duration inflation.** The declared duration of every stem in the recording scales with the slot count, so it inflates by the same ratio. `1 / 0.106 ≈ 9.4` matches the 4.9x-8.9x inflation observed in #792.

It also explains the case in #792 that looked like a counterexample: a recording with a correct 1.03x duration that was still zero-stuffed. Slots are flushed at `min(max_voice_slot, max_system_slot) - 1`, so while both streams are live the microphone gates the slot count and the duration stays right — the system stem is zero-stuffed regardless.

## Evidence

Measuring nonzero density by position within each 1600-sample slot on an affected recording, over a 20 s window (200 slots):

```
samples=320000  nonzero=7.89%
nonzero density: flat through position 170, exactly 0 from position 171 onward
per-slot leading nonzero run: 171, 171, 170, ...
nonzero samples after the run, in any slot: 0
```

171 ≈ 512/3. A boundary landing on the same offset inside every slot of a 1600-sample grid is the `padded_slot` signature, not an acoustic property of the source.

## The change

Fix the producer rather than the mixer, since the mixer's contract (one chunk per source per slot) is the reasonable one.

- `streaming::ChunkAccumulator` buffers a source's resampled samples and emits exactly `CHUNK_SAMPLES` per chunk with a gapless index.
- The Core Audio tap now goes through it.
- The microphone path, which already did this by hand, now shares the same implementation instead of duplicating it.
- `CHUNK_SAMPLES` is public, and `DUAL_SOURCE_SLOT_SAMPLES` is defined in terms of it so the two cannot drift apart.

Two secondary fixes fall out:

- Oversized callbacks are now split across chunks. Previously `padded_slot`'s `.take(DUAL_SOURCE_SLOT_SAMPLES)` silently discarded anything past a slot boundary.
- `pending_system` no longer accumulates the ~90% of system chunks whose index runs ahead of the microphone's for the length of the recording.

A trailing partial chunk (<100 ms) is dropped when a stream stops. That is what the microphone path has always done; this makes the tap consistent with it.

## Testing

New unit tests in `streaming.rs` covering the accumulator, including a regression test that feeds tap-sized (171-sample) callbacks and asserts the emitted chunks are dense rather than zero-stuffed, and that total emitted samples account for the input with no slot inflation. Also: empty-until-full, exact chunk length across callback sizes, gapless indices, and oversized-callback splitting.

A second commit adds a CI step, because those tests would not otherwise have run. `Unit tests (no whisper)` invokes `cargo test -p minutes-core --no-default-features`, so every `#[cfg(test)]` module inside a streaming-gated file is compiled by `Build core with streaming only` and then never executed — that covered the new tests here and the existing `mic_mute_tests`. The new step runs the same suite with `--features streaming`.

Verified on CI: `Build core with streaming only (downstream consumer path)` passes on `macos-latest`, which is the configuration that compiles both the Core Audio tap and the dual-source mixer.

I have a shell harness that plays a known `say`-generated signal through the default output device, records, and asserts nonzero ratio, slot-density profile, and declared-vs-wall-clock duration on the resulting stems. Happy to contribute it under `scripts/` if that is useful.

---

Related: this is one of two independent PRs for #792. The other reports a truncated native-call stem instead of inferring it; they touch different files and can land in either order.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,15 @@ jobs:
         run: cargo test -p minutes-core --no-default-features --lib -- --test-threads=1 --nocapture
         timeout-minutes: 15
 
+      - name: Unit tests (streaming feature)
+        # The step above runs with streaming off, so every #[cfg(test)] module
+        # inside a streaming-gated file was compiled by "Build core with
+        # streaming only" and then never executed. That covered the dual-source
+        # chunking invariants in streaming.rs and the mic-mute tests. Run the
+        # same suite once more with the feature on.
+        run: cargo test -p minutes-core --no-default-features --features streaming --lib -- --test-threads=1 --nocapture
+        timeout-minutes: 15
+
       - name: Copilot attach, routing, and window-contract tests
         # On Windows this executes the real named-pipe connect/reconnect path.
         # On Unix it also asserts owner-only socket/discovery permissions.
--- a/crates/core/src/capture.rs
+++ b/crates/core/src/capture.rs
@@ -950,8 +950,12 @@ fn try_reconnect(
     }
 }
 
+/// One mixer slot is one `AudioChunk` from each source. Kept equal to
+/// `streaming::CHUNK_SAMPLES` by construction: `padded_slot` pads every slot
+/// to this length, so a producer emitting a different chunk size writes a
+/// zero-stuffed stem (#792).
 #[cfg(feature = "streaming")]
-const DUAL_SOURCE_SLOT_SAMPLES: usize = 1600;
+const DUAL_SOURCE_SLOT_SAMPLES: usize = crate::streaming::CHUNK_SAMPLES;
 
 #[cfg(feature = "streaming")]
 struct DualCaptureWriters {
--- a/crates/core/src/streaming.rs
+++ b/crates/core/src/streaming.rs
@@ -54,6 +54,74 @@ pub struct AudioChunk {
     pub source: SourceRole,
 }
 
+/// Samples in one `AudioChunk`: 100 ms of 16 kHz mono audio.
+///
+/// Every producer of `AudioChunk` must emit exactly this many samples per
+/// chunk. Dual-source capture derives its mixer slots from
+/// `AudioChunk::index` and pads each slot out to this length, so a producer
+/// that forwards whatever its driver callback happened to deliver both
+/// zero-stuffs its own stem and inflates the slot count for every other
+/// source sharing the mixer. Use `ChunkAccumulator` rather than chunking by
+/// hand. See issue #792.
+pub const CHUNK_SAMPLES: usize = 1600;
+
+/// Buffers a producer's resampled 16 kHz mono samples into fixed-size
+/// `CHUNK_SAMPLES` chunks carrying a monotonic per-producer index.
+///
+/// Driver callbacks deliver whatever their IO buffer size dictates — a Core
+/// Audio process tap on a 48 kHz output device hands over 512 frames per
+/// callback, which is 170.67 samples once decimated to 16 kHz. Emitting that
+/// directly as an `AudioChunk` is what produced the zero-stuffed system stems
+/// in #792: the dual-source mixer padded each 171-sample chunk out to a
+/// 1600-sample slot, leaving ~89% zeros, and advanced the slot counter ~9.4x
+/// faster than the microphone did.
+///
+/// The trailing partial chunk stays buffered rather than being emitted short.
+/// At most `CHUNK_SAMPLES - 1` samples (<100 ms) are dropped when a stream
+/// stops, matching what the microphone path has always done.
+pub struct ChunkAccumulator {
+    buf: Vec<f32>,
+    next_index: u64,
+}
+
+impl Default for ChunkAccumulator {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ChunkAccumulator {
+    pub fn new() -> Self {
+        Self {
+            // Two chunks of headroom: a callback can straddle a boundary and
+            // leave a remainder without forcing a realloc.
+            buf: Vec::with_capacity(CHUNK_SAMPLES * 2),
+            next_index: 0,
+        }
+    }
+
+    /// Append `samples`, invoking `emit(index, chunk)` once per complete
+    /// `CHUNK_SAMPLES`-sample chunk. Called from real-time audio callbacks,
+    /// so it allocates only the emitted chunks themselves.
+    pub fn push<F>(&mut self, samples: &[f32], mut emit: F)
+    where
+        F: FnMut(u64, Vec<f32>),
+    {
+        self.buf.extend_from_slice(samples);
+        while self.buf.len() >= CHUNK_SAMPLES {
+            let chunk: Vec<f32> = self.buf.drain(..CHUNK_SAMPLES).collect();
+            let index = self.next_index;
+            self.next_index += 1;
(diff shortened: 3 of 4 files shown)


silverstein (approved):
Reviewed and verified locally. This is better than the patch I had open for the same bug (#794), which I have closed in favor of this one.

What I checked, beyond reading it:

The CI gap is real and I confirmed it independently rather than taking it on trust. `default = ["whisper"]`, so `--no-default-features` leaves `streaming` off, and running the existing CI command filtered to `streaming::` reports `0 passed; 1763 filtered out`. Every test in that module has been compiling and never executing. That is the most valuable part of this PR, and it retroactively explains how a chunking invariant could break without a red build.

Deriving `DUAL_SOURCE_SLOT_SAMPLES` from `streaming::CHUNK_SAMPLES` is the right shape. My patch added a second constant beside the tap, which leaves the same trap for the next producer. One constant that both the mixer slot and the producer contract read cannot drift.

Refactoring the existing microphone path onto `ChunkAccumulator` is what makes this a fix rather than a patch. The hand-rolled chunking the tap failed to copy no longer exists to be copied wrongly.

The borrow split in `ingest` is correct, and the accumulator being recreated with the context on reconnect is fine: the mixer advances `slot_base` past the previous maximum on restart, so a per-producer index restarting at zero does not collide.

`chunk_accumulator_tests` all pass here, as does the wider core suite with the feature on.

Two observations, neither blocking:

The tests cover `ChunkAccumulator` itself rather than the tap using it, so a future producer that bypasses the accumulator would not turn anything red. Given the new CI step and the shared constant, that is a reasonable place to stop, but it is the remaining hole.

`buf` keeps whatever capacity a large callback forced. It is bounded by the largest callback ever seen and never grows unbounded, since the drain loop runs to completion each push, so this is a note rather than a request.

Security review: no new dependencies, no `unsafe`, no network or process execution, no filesystem writes outside tests. Merging on that basis.

silverstein (approved):
Reviewed and verified locally. This is better than the patch I had open for the same bug (#794), which I have closed in favor of this one.

What I checked, beyond reading it:

The CI gap is real and I confirmed it independently rather than taking it on trust. `default = ["whisper"]`, so `--no-default-features` leaves `streaming` off, and running the existing CI command filtered to `streaming::` reports `0 passed; 1763 filtered out`. Every test in that module has been compiling and never executing. That is the most valuable part of this PR, and it retroactively explains how a chunking invariant could break without a red build.

Deriving `DUAL_SOURCE_SLOT_SAMPLES` from `streaming::CHUNK_SAMPLES` is the right shape. My patch added a second constant beside the tap, which leaves the same trap for the next producer. One constant that both the mixer slot and the producer contract read cannot drift.

Refactoring the existing microphone path onto `ChunkAccumulator` is what makes this a fix rather than a patch. The hand-rolled chunking the tap failed to copy no longer exists to be copied wrongly.

The borrow split in `ingest` is correct, and the accumulator being recreated with the context on reconnect is fine: the mixer advances `slot_base` past the previous maximum on restart, so a per-producer index restarting at zero does not collide.

`chunk_accumulator_tests` all pass here, as does the wider core suite with the feature on.

Two observations, neither blocking:

The tests cover `ChunkAccumulator` itself rather than the tap using it, so a future producer that bypasses the accumulator would not turn anything red. Given the new CI step and the shared constant, that is a reasonable place to stop, but it is the remaining hole.

`buf` keeps whatever capacity a large callback forced. It is bounded by the largest callback ever seen and never grows unbounded, since the drain loop runs to completion each push, so this is a note rather than a request.

Security review: no new dependencies, no `unsafe`, no network or process execution, no filesystem writes outside tests. Merging on that basis.

silverstein:
Your two red checks are both mine to deal with, not yours.

**Test (ubuntu-latest)** is `bounded_child::tests::continuously_writing_setsid_escape_cannot_hold_pipe_workers`, which is a race in the test rather than the invariant. It failed on three unrelated PRs today. Fixed in #799, measured before and after under CPU saturation: 6 failures in 25 runs before, 0 in 25 after.

**Test (windows-latest)** is more interesting, and it is a genuine find that your PR is responsible for surfacing but not for causing. Filed as #800.

`policy_fs::tests::restricted_override_audit_child_process_entry` passes in the existing step and fails in the streaming step you added, on the same runner, in the same job:

```
19:35:19  --no-default-features                      ... ok      (13 passes)
19:44:38  --no-default-features --features streaming ... FAILED  (4 failures)
```

with `private correction store DACL is not owner-only` from the owner-only attestation in `overlays.rs`. Both steps run `--test-threads=1`, so it is not cross-test concurrency.

So the configuration you enabled was not merely untested, it was broken, and nobody could have known because `default = ["whisper"]` meant CI never built it. Why the feature flag changes a Windows ACL outcome is not yet understood and needs a real Windows box to answer, which I have.

It fails closed, so this is a bug rather than an exposure: the attestation refuses a store it cannot verify instead of accepting one it cannot vouch for.

I do not want to hold your fix behind an unrelated pre-existing Windows bug. Suggested unblock, if you are happy with it: skip that single test in the new Windows streaming step with a comment pointing at #800, so the rest of the newly enabled Windows coverage still runs rather than disabling the step. Push it to your branch and I will merge on green, or say the word and I will push it to your branch myself.

silverstein:
Updated your branch from main so it picks up two things that were failing it for reasons unrelated to your change:

- #799 fixes the `bounded_child` setsid race that was failing your ubuntu job. Measured under CPU saturation: 6 failures in 25 runs before, 0 in 25 after.
- #796 stops Linux jobs hanging for hours on apt, which is what left every PR today sitting with no verdict at all.

That should leave only the Windows DACL failure in #800, which is pre-existing and yours only in the sense that your CI step is the first thing to have ever run that configuration. Diagnostic for it is in #801: the attestation currently returns one message for seven distinct conditions, so that branch makes it name the one that actually tripped. Once it reports back, we will know whether it is an owner mismatch, an inheritable DACL, or an extra entry, which are three different bugs.

Offer stands on unblocking you: skip that single test in the Windows streaming step with a pointer to #800, so the rest of the newly enabled Windows coverage still runs. Push it and I will merge on green, or say the word and I will push it to your branch.

PR #798, opened 2026-08-18
fix(tauri): report a truncated native-call stem instead of inferring it (#792)

MrFreekour:
Addresses the reporting half of #792 Bug 1: a native call capture whose system stem is lost at finalization queues silently, and the user's only signal is a guess.

## The gap

`native_call_processing_input` refuses a capture only when **both** PCM stems fail `viable_native_call_stem`:

```rust
if !voice_viable && !system_viable {
    return Err(...);
}
```

When one stem finalizes truncated and its sibling is healthy — which is the common case, since the microphone stem survives — the function returns `Ok` with `recovery_health: None`. Nothing marks the capture. The only thing that eventually reaches the user is a downstream warning:

```json
{"kind":{"other":{"code":"native-call-system-stem-recovery"}},
 "source":"voice",
 "message":"Call/remote audio was missing or silent. This transcript contains local microphone audio only.",
 "diagnostic_confidence":"inferred"}
```

"Missing or silent", inferred — indistinguishable from a call where the remote side genuinely said nothing.

## Why the stem is not merely "missing"

#792 reports the system stem finalizing at exactly 7,936 bytes across multiple sessions, including one reproduced on demand and one from ten weeks earlier on the same machine. The header on such a file explains it:

```
RIFF f80f0000  -> declares 4,096 bytes
WAVE
JUNK (28 bytes)
fmt  format=3 (IEEE float) channels=1 rate=48000 byte_rate=192000 bits=32
FLLR (4,008 bytes of Apple page-alignment filler)
...4,096-byte aligned header, then ~0.02 s of audio
```

The RIFF size field still declares the 4,096-byte page-aligned header block the writer laid down before any audio was appended. The file was never finalized — but capture did run, and the `native-captures/` stem was verified growing normally with ~99.7% nonzero samples for the whole session before this.

So a stem that is present on disk and below the viability threshold is *observed* evidence of a finalization failure, not something to infer.

## The change

In `native_call_processing_input`, after the existing both-stems-failed check, detect a stem that exists but finalized below `MIN_VIABLE_STEM_DURATION_SECS` and record it as a `High`-confidence `native-call-invalid-stem` warning carrying the measured byte count and approximate duration, plus an error-level log line.

`health::append_native_call_invalid_stem_warning` already models exactly this case and already uses `DiagnosticConfidence::High`. Nothing on the truncated-stem path was reaching it.

`truncated_native_call_stem_reason` returns `None` for an absent stem, so single-stem captures — a supported outcome with its own recovery reporting — are not relabelled as finalization failures.

## What this deliberately does not do

The issue asks for a hard error when a 0.02 s stem appears after preflight reported `system_audio_ready: true`. I did not make it fail the job: the surviving sibling is typically a full-length microphone recording, and refusing to queue would destroy usable audio in order to punish a partial capture. The recording still processes; it is now labelled accurately and at the right confidence, which is what makes the failure actionable.

This is reporting only. It does not stop the stem from being truncated — that is upstream in the native call helper's finalize path, and worth a separate fix.

## Testing

Four new unit tests, all confirmed executing on `macos-latest` CI: truncated system stem produces a `High`-confidence warning attributed to `System` and carrying the measured size; the mirrored voice case; a healthy pair produces no warning (this path runs on every successful capture, so a false positive would mark every recording); and an absent stem is not reported as truncated.

---

Related: this is one of two independent PRs for #792. The other fixes the zero-stuffed system stem at its root; they touch different files and can land in either order.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -2981,18 +2981,47 @@ fn viable_native_call_stem(path: &Path) -> bool {
         .is_ok_and(|metadata| metadata.len() >= min_viable_stem_bytes(path))
 }
 
-fn native_call_capture_warning_health(
-    message: impl Into<String>,
-) -> minutes_core::markdown::RecordingHealth {
-    let mut health = minutes_core::markdown::RecordingHealth {
+fn empty_native_call_health() -> minutes_core::markdown::RecordingHealth {
+    minutes_core::markdown::RecordingHealth {
         voice_stem_active_ratio: None,
         system_stem_active_ratio: None,
         system_dominant_ratio: None,
         capture_warnings: Vec::new(),
         diarization_path: Some(minutes_core::markdown::DiarizationPath::None),
-    };
+    }
+}
+
+fn native_call_capture_warning_health(
+    message: impl Into<String>,
+) -> minutes_core::markdown::RecordingHealth {
+    let mut health = empty_native_call_health();
     minutes_core::health::append_native_call_capture_warning(&mut health, message);
     health
+}
+
+/// Describe a stem that exists on disk but finalized below the viability
+/// threshold, or `None` when the stem is absent or long enough to use.
+///
+/// This is a materially different failure from "the stem was never written":
+/// capture ran for the whole session and the bytes were lost when the file was
+/// closed. Issue #792 reports the system stem repeatedly finalizing at 7,936
+/// bytes -- a page-aligned WAV header plus 0.02 s of audio -- after growing
+/// normally for half an hour.
+fn truncated_native_call_stem_reason(path: &Path) -> Option<String> {
+    let len = path.metadata().ok()?.len();
+    let min_bytes = min_viable_stem_bytes(path);
+    if len >= min_bytes {
+        return None;
+    }
+    let byte_rate = read_wav_byte_rate(path)
+        .unwrap_or(FALLBACK_WAV_BYTE_RATE)
+        .max(1);
+    // Approximate: everything past the header is treated as audio. Good enough
+    // to tell "0.02 s" from "0.5 s" in a diagnostic message.
+    let observed_secs = len as f64 / byte_rate as f64;
+    Some(format!(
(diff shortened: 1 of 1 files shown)


silverstein:
Reviewed and verified locally. All nine `native_call` tests and the four new truncation tests pass here.

This is the right half of bug 1 to fix in this file. `native_call_processing_input` refusing only when both stems fail is exactly the gap: the common case is one healthy stem beside one destroyed one, and that returned `Ok` with nothing recorded. Turning that into a high confidence invalid-stem warning, instead of leaving a downstream "missing or silent" guess to reach the user, is the difference between a diagnosis and a shrug.

Your header dump also independently confirms the arithmetic I posted on the issue: a 4,096 byte aligned header, made up of `JUNK` plus Apple `FLLR` padding, plus 960 frames at 4 bytes, is exactly the 7,936 bytes you measured, and 960 frames at 48 kHz is the 0.020 s. Two different routes to the same number is about as solid as this gets without a live capture.

One defect, cosmetic but user-visible. The message in `truncated_native_call_stem_reason` has a run of ten spaces in it:

```
"stem finalized at {len} bytes (~{observed_secs:.3}s of audio); at least          {MIN_VIABLE_STEM_DURATION_SECS:.1}s is required for a stem to be usable"
```

That string lands in the recording health warning a user reads, so it will render as "at least          0.5s is required". Happy to fix it in a follow-up with credit to you if you would rather not round-trip a whitespace change, or push to the branch yourself and I will merge on green. Your call.

Security review: no new dependencies, no `unsafe`, no network or process execution. The only filesystem writes are test fixtures in `TempDir`. Logging the stem path in the `tracing::error!` is consistent with existing practice in this file and in `capture.rs`; the no-filename rule in this repo is specific to the knowledge and QMD subsystems, which is a different trust boundary.

This pairs with #793, which stops the stem being truncated in the Swift helper in the first place. Yours reports it, mine prevents it, and both are worth having: the prevention can regress, and a capture that fails for some other reason should still say so.

Issue #800, opened 2026-08-18
Windows: owner-only DACL attestation fails only in the streaming build


silverstein:
`policy_fs::tests::restricted_override_audit_child_process_entry` fails on Windows when `minutes-core` is built with the `streaming` feature, and passes on the same runner without it.

Surfaced by #797, which adds the missing CI step that runs the core unit tests with `streaming` enabled. It is not caused by that PR. Until it landed, `default = ["whisper"]` plus `--no-default-features` meant this configuration had never been executed in CI at all.

## Evidence

Same job, same Windows runner, one run:

```
19:35:19  cargo test -p minutes-core --no-default-features --lib
          restricted_override_audit_child_process_entry ... ok      (13 passes)

19:44:38  cargo test -p minutes-core --no-default-features --features streaming --lib
          restricted_override_audit_child_process_entry ... FAILED  (4 failures)
```

The failure, from `crates/core/src/overlays.rs:472`:

```
called `Result::unwrap()` on an `Err` value: Custom { kind: Other,
  error: "private correction store DACL is not owner-only" }
```

Both steps pass `--test-threads=1`, so this is not concurrency between tests.

## What the check is asserting

The owner-only attestation in `overlays.rs` requires all of: the object owner SID equals the current user, a non-null DACL, `SE_DACL_PROTECTED` set so nothing is inherited, and exactly one ACE. Any of those failing produces this error.

That test is a child-process entry point: it returns early unless `MINUTES_TEST_AUDIT_DIR` is set, and it is spawned by `restricted_override_audit_serializes_across_processes` through `current_exe()`. So the failing code path is a real child process attesting a real store, not test scaffolding.

## Why this is not a silent security hole

It fails closed. The attestation returns an error rather than accepting a store whose DACL it could not verify as owner-only, so the production consequence is a visible failure, not a private store that is quietly world-readable. That is the right direction to fail in, and it is why this is a bug to fix rather than an incident.

## What is not yet known

Why the `streaming` feature changes the outcome at all. `policy_fs` and `overlays` do not obviously depend on it. Candidates worth checking, in rough order of likelihood:

- the spawned child inherits something different from the streaming test binary (working directory, environment, or the filter it runs under)
- a dependency pulled in only by `streaming` changes process or handle inheritance on Windows
- the store path resolves differently in that build

Reproducing needs a Windows machine, not CI: run the two commands above and compare the actual DACL on the audit directory between the two builds.

## Unblocking in the meantime

[...]

PR #801, opened 2026-08-18
[diagnostic, do not merge] name which owner-only DACL condition fails (#800)

silverstein:
**Diagnostic only. Not for merge.** Open as a draft so CI runs it on Windows, which is the only place #800 reproduces.

## Why

The owner-only attestation in `overlays.rs` returns one message for seven distinct failure conditions:

```
private correction store DACL is not owner-only
```

That is why #800 currently has no diagnosis. Four of eight concurrent processes failed and the message cannot distinguish "the object is owned by another account" from "the DACL is inheritable" from "the DACL has more than one entry". Those point at completely different bugs.

## What this does

Evaluates each condition separately and names the one that tripped, with the observed control word and entry count. It also carries a temporary copy of the streaming test step from #797, because #797 is not merged yet and the failure does not reproduce without it.

## What happens next

Read the Windows job output, learn which condition actually fails, then close this branch. The improved error message is worth keeping on its own and will be reproposed as a normal PR without the temporary CI step, since an attestation that refuses a store should say what it objected to.

Credit for the CI step to @MrFreekour in #797; it is copied here only so this branch reproduces the failure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -247,6 +247,13 @@ jobs:
         run: cargo test -p minutes-core --no-default-features --lib -- --test-threads=1 --nocapture
         timeout-minutes: 15
 
+      - name: Unit tests (streaming feature)
+        # Temporary, to reproduce #800 where it actually happens. The permanent
+        # version of this step is MrFreekour's in #797; this branch is a
+        # diagnostic and is not for merge.
+        run: cargo test -p minutes-core --no-default-features --features streaming --lib -- --test-threads=1 --nocapture
+        timeout-minutes: 20
+
       - name: Copilot attach, routing, and window-contract tests
         # On Windows this executes the real named-pipe connect/reconnect path.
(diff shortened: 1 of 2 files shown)


silverstein:
Baseline diagnostic result at exact SHA 3dc54f0cf41b72e9d0ffbdb791c960d0b7029c21: the Windows runner completed attempts 1 through 25 and reported "#800 did not reproduce in 25 attempts"; the full job passed. This is honest non-reproduction, not disproof of the race. PR #808 removes the source-proven transient state by attaching the final owner-only protected DACL in SECURITY_ATTRIBUTES at CREATE_NEW; its fresh Windows matrix is running at c652d0cdeb368b45beab0d189c6e0d87addacde0.

silverstein:
Diagnostic work is complete and this draft was never intended to merge. The baseline Windows run completed 25 attempts without reproducing #800; the source-proven create-then-tighten race was removed by #808, which landed on main as d5818be2518106c4084e499505f6443a55055b07 with green post-merge CI.

PR #803, opened 2026-08-19
docs(readme): fix the broken star history chart

Dessalines39394:
The README star history chart is currently broken. The upstream service that produced it now trips GitHub stargazer API restrictions, so the badge renders nothing. This swaps the chart over to a maintained mirror that uses a token-free data source, restoring the chart. The link target and image were both updated, and the wrapping link now uses the hash-based URL format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1365,7 +1365,7 @@ const results = await searchMeetings("~/meetings", "pricing");
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


silverstein (approved):
Verified at exact SHA 125bed6747c3b0dcc77736cd166c25a4e7767879: this is a one-line README-only change, the replacement SVG returns valid image/svg+xml, the linked chart page returns 200, and the full required check set is green. Thanks for repairing this.

PR #805, opened 2026-08-19
fix(desktop): do not stop a recording owned by another process (#804)

### LINKED ISSUE #804: App can terminate a CLI-started recording via stop sentinel when app and CLI versions differ
## Summary

When the desktop app and CLI are on different versions but share `~/.minutes`, the app can terminate
a recording started by the CLI, even when the app itself is not recording anything.

Split out of #792 for separate tracking.

## Environment
- macOS 26.6.1 (25G76), Apple Silicon
- Minutes.app **0.25.1**
- CLI **0.18.2** (a hand-installed binary at `~/bin/minutes-macos-arm64`, not managed by the app)
- Shared state directory: `~/.minutes`

## What happened

A CLI recording started successfully and was capturing correctly (readiness probe passed, both stems
growing, verified real signal in each). It then died after 202 seconds with:

```
INFO stop sentinel detected — stopping dual-source recording
```

The recording did not crash, hit a timeout, or lose its audio source. It was told to stop.

At the same time, the app itself was capturing nothing: no `recording.pid`,
`recording-start.txt` or `recording-meta.json`, no audio file created anywhere under `$HOME`, and
`lsof` on the app pid showed no audio file open. `minutes status` agreed (`recording: false`).

Quitting the app resolved it, and CLI recordings then ran to completion (one ran 82 minutes without
interruption).

## Why it matters

The failure mode is silent and destructive in the same way as the issues in #792: the recording ends
mid-meeting with no user-visible signal, and the loss is only discovered afterwards. For anyone using
the CLI for capture (scripted or automated recording) while the desktop app is installed and running,
a version mismatch can silently end recordings.

## Suggested handling

Some form of guarding around the shared state contract, for example:
- Write a version/owner marker alongside the stop sentinel and ignore sentinels not intended for the
  current recording's owner
- Have the app skip writing a stop sentinel when it has no active recording of its own
- Warn on startup when the app detects a CLI binary with an incompatible version sharing `~/.minutes`

## Note on a related symptom I could not confirm

In the same session, the 0.25.1 app appeared to show a recording in progress while writing nothing to
disk for several minutes. I initially attributed this to a separate "0.25.1 does not record" bug, but
I could not reproduce it afterwards: on a later day the same app version recorded four times in a row
without issue. It may have been UI state showing armed-but-not-recording, or a side effect of the
version skew above. Mentioning it only for completeness, not as a claim.

silverstein:
Fixes #804 (split out of #792 bug 5).

This was written yesterday and then not opened as a PR, which is my mistake: the branch has been sitting pushed and idle while the issue stayed open.

## The bug

`request_stop` signaled whatever process held the recording PID, with no check on whether this app started that recording. It writes the stop sentinel and, on Unix, sends SIGTERM. In the reported case the app was capturing nothing and terminated a working CLI recording after 202 seconds.

The path it was written for is real and worth keeping: run `minutes record` in a terminal, hit Stop in the tray, expect it to stop. What was missing is the distinction between that and stopping a recording the app has confused for its own.

## The rule

Refuse only when the app believes it is recording *and* the recording is not ours. Signal in every other case.

Ownership is asked about via `desktop_app_owns_pid` rather than inferred from the PID, because the app can legitimately own a recording running under a different PID, and refusing that would break stopping our own capture. On platforms without the ownership probe the previous behavior is kept rather than guessed at.

When it refuses it also clears its own recording flag, since that flag is the thing that is actually wrong in that state, and returns a message saying the recording was left running and where to stop it.

Three tests, mutation-verified in both directions: making the decision always signal fails the foreign-recording test, which is the bug; making it refuse whenever the recording is foreign fails the tray test, which is the feature. Only the stated rule passes both.

## Against the suggestions in the issue

> Have the app skip writing a stop sentinel when it has no active recording of its own

This is close to what landed here, but not quite, and the difference matters. Skipping whenever the app has no recording of its own would break the tray-stop path above, which is a deliberate feature. The condition used instead is the incoherent state specifically: the app thinks it is recording, and the recording belongs to somebody else. Those are two different recordings, so stopping that one is not what was asked for.

> Write a version/owner marker alongside the stop sentinel and ignore sentinels not intended for the current recording's owner

This is the stronger design and I would like it in addition, not instead. What is here guards the writer, and only in the state that was observed. A marker guards the reader, which also covers a writer we have not thought of, including older or newer builds that predate any fix. It is a shared-state contract change, so it deserves its own PR rather than being smuggled into this one.

> Warn on startup when the app detects a CLI binary with an incompatible version

Agreed in principle, and worth doing after the marker, since with the marker in place the version skew stops being dangerous and becomes merely worth mentioning.

## On the retracted symptom

The issue notes the "0.25.1 does not record" symptom could not be reproduced and may have been UI state or a side effect of the skew. That matches what I found reading the code: I could not construct a path where the app sets its recording flag without a PID file existing, and I was not willing to guess at a fix in capture. Treating it as withdrawn unless it recurs.

### DIFF
--- a/tauri/src-tauri/src/commands.rs
+++ b/tauri/src-tauri/src/commands.rs
@@ -3873,6 +3873,43 @@ pub fn recording_active(recording: &Arc<AtomicBool>) -> bool {
     recording.load(Ordering::Relaxed) || minutes_core::pid::status().recording
 }
 
+/// Whether to signal a recording owned by some other process.
+///
+/// Stopping an externally started recording from the tray is a feature: the
+/// user runs `minutes record`, then hits Stop in the app, and expects it to
+/// work. What is not a feature is the app stopping a recording it does not own
+/// while believing that recording is its own.
+///
+/// That combination is what issue #792 bug 5 reports. The desktop app showed a
+/// recording in progress while capturing nothing, and the only recording on
+/// the machine belonged to a separate CLI process, which it then terminated
+/// after 202 seconds. The user loses a recording the app never made.
+///
+/// The app can legitimately own a recording under a different PID, which is
+/// why ownership is asked about rather than assumed from the PID alone.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum ExternalStopDecision {
+    /// Signal it. Either the app knows it is not the one recording, so this is
+    /// a deliberate stop of someone else, or the recording is ours under
+    /// another PID.
+    Signal,
+    /// Refuse. The app believes it is recording, yet the recording belongs to
+    /// a process that is not ours, so these are two different recordings and
+    /// stopping this one is not what was asked for.
+    RefuseForeign,
+}
+
+pub(crate) fn external_stop_decision(
+    app_believes_it_is_recording: bool,
+    desktop_app_owns_recording: bool,
+) -> ExternalStopDecision {
+    if app_believes_it_is_recording && !desktop_app_owns_recording {
+        ExternalStopDecision::RefuseForeign
+    } else {
+        ExternalStopDecision::Signal
+    }
+}
+
 pub fn request_stop(
     recording: &Arc<AtomicBool>,
     stop_flag: &Arc<AtomicBool>,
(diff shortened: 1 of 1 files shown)


silverstein:
Superseded by #807. Exact head 7ccb13b6b6e25a2f454592de9a38d02e8d2980c0 was merged into the approved #807 head, and the combined writer-side ownership guard plus addressed sentinel landed on main as a22292026ae3ba73fd086ff87581fa10292a2c4d.

Issue #802, opened 2026-08-18
Timing-sensitive tests block merges on loaded runners (three today)


silverstein:
Three separate tests failed today for timing reasons rather than defects, across five pull requests, and between them they blocked every merge for most of the day. Filing the pattern rather than a fourth individual bug, because the individual fixes are already in and the shape is what matters.

## The three

**`bounded_child::tests::continuously_writing_setsid_escape_cannot_hold_pipe_workers`** (Linux). The launcher exited before the escaped writer was scheduled, so `run` collected an empty pipe from a finished parent and returned `Ok`. Failed on three unrelated PRs including a contributor's. Fixed in #799, measured under CPU saturation: 6 failures in 25 runs before, 0 in 25 after.

**`uses exactly two persistent shared slots across more than 64 corpus lifetimes`** (Windows). 80 worker process lifecycles against a 30s limit, landing at 30024ms. Missed by 24 milliseconds and reported nine further tests in the same file as failures purely because the first one timed out. Fixed in #799.

**`never returns outside bytes across a Windows parent junction swap`** (Windows). Passed in 543ms on #801 and failed at 2420ms on #797 in the same hour, neither of which touches `secure-read`. Not yet fixed.

## What they have in common

Each asserts a real invariant, and each expresses it as "this finishes within N", where N was chosen on an unloaded machine. GitHub's runners are shared and Windows process creation is slow, so N is not a property of the code under test. When the machine is busy, the test reports a defect that is not there.

The cost is not just a red check. It is that a red check stops meaning anything: today I very nearly attributed one of these to a change of my own, and the only thing that disproved it was running the unmodified tree under the same load and watching it fail *more*.

## Suggested direction

Not proposing a blanket timeout increase. Each of these has a better fix available, and the first two show the pattern:

- Where a deadline is incidental to the assertion, make the wait bounded but generous, and keep the strict assertion on the thing actually under test. The corpus test asserts slot reuse, not speed, so its limit was the right thing to relax.
- Where the test races a scheduler, make it synchronize instead of hoping. The setsid test now waits for the escapee to publish a pid before the launcher exits, so there is nothing left to race.
- Where a security invariant is timing sensitive, that is worth understanding before relaxing anything, since it may be telling us the invariant itself is timing dependent. The junction-swap test is in this class and should not be quietly given a longer timeout.

## Worth knowing

[...]

PR #806, opened 2026-08-20
feat(capture): warn when a stem finalizes mostly silence (#792)

silverstein:
Implements the safety net @MrFreekour asked for in #792, and closes the gap between "the cause is fixed" and "the next cause will be caught".

## Why this exists even though #797 fixed the cause

Their follow-up raised the severity of bug 2 considerably. The zero-stuffed stem did not merely produce poor transcripts: for anything over roughly twenty minutes, processing failed outright with `decoded audio exceeds the resource budget`, and they lost three of four meetings in one day. The failure arrived hours after the recording, from a different subsystem, phrased as a resource limit, about audio that no longer existed to re-run.

#797 removes that cause. This catches the next one, at the point where the recording is still on disk.

## The rule

A stem that finalizes with real audio in it but mostly exact zeros is reported on stderr and in the structured log.

Deliberately narrow, because a warning that fires on healthy recordings is worse than no warning:

- **Below 25% nonzero.** The reported corruption measured 7.9% to 10.7%. The healthy microphone stems in the very same recordings measured 97% to 99.7%, because a live input carries dither rather than exact zeros even when nobody is speaking. The threshold sits in a wide empty gap between those, not near either.
- **All-zero stems are excluded.** No nonzero samples at all means a source that never delivered, a different fault already named by the gap counters from #795. Reporting both would hand the user two explanations for one problem, one of them wrong.
- **Measured while writing.** No second pass over hundreds of megabytes, and no reading a file back to ask something we could answer while writing it.

## On their other two points

> nonzero ratio is a reliable health signal; duration is not

Agreed and this relies on exactly that. Their recording with a correct 1.03x duration header that was still ~10% nonzero is why duration is not checked here at all.

> worth a regression test: record 30+ minutes of system audio and assert the job completes

Not practical in CI, but the invariant behind it is what these tests assert, and at an earlier point than the decode budget would.

## Verification

Five tests, mutation-verified in both directions: removing the all-zero exclusion makes the silent-stem test fail, and raising the threshold to 0.999 makes the healthy-microphone test fail.

`cargo clippy --all --no-default-features -- -D warnings` and `cargo fmt --all --check` clean.

Full-suite note: the streaming suite run in parallel on a busy machine fails a rotating set of `knowledge::` tests. Unmodified `main` fails 11 under those conditions and this branch fails 6, so that is #802 rather than anything here. CI runs the suite with `--test-threads=1`, which avoids the contention.

### DIFF
--- a/crates/core/src/capture.rs
+++ b/crates/core/src/capture.rs
@@ -1059,7 +1059,6 @@ fn update_audio_level_from_samples(samples: &[f32]) {
     AUDIO_LEVEL.store(level, Ordering::Relaxed);
 }
 
-#[cfg(feature = "streaming")]
 /// Running tally of how much of a stem is actually audio.
 ///
 /// Counted while the samples are already in hand. Reading the finished file
@@ -1130,20 +1129,6 @@ fn write_samples_to_wav(
             .map_err(|e| CaptureError::Io(std::io::Error::other(format!("WAV write: {}", e))))?;
     }
     content.record(&converted);
-    Ok(())
-}
-
-#[cfg(not(feature = "streaming"))]
-fn write_samples_to_wav(
-    writer: &mut hound::WavWriter<std::io::BufWriter<std::fs::File>>,
-    samples: &[f32],
-) -> Result<(), CaptureError> {
-    for &sample in samples {
-        let s16 = (sample * 32767.0).clamp(-32768.0, 32767.0) as i16;
-        writer
-            .write_sample(s16)
-            .map_err(|e| CaptureError::Io(std::io::Error::other(format!("WAV write: {}", e))))?;
-    }
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


silverstein:
Fixed the deterministic Clippy failure at exact SHA 996345648d6ec2c36f523648092ae39e9fb29963 by removing the newly added non-streaming stem writer, which has no non-streaming callers, plus its duplicate feature gate. Local evidence: the exact failing Parakeet Clippy command is clean; cargo fmt is clean; 5/5 stem-content regressions pass. The full GitHub matrix is rerunning.

silverstein:
Independent capture-path review found and removed three per-slot heap allocations introduced by the tally. The same already-converted samples are now counted in place. Replacement exact SHA: 518de6c4b935c04c0034375e54210a30f03a8cc4. Additional evidence: streaming Clippy with -D warnings is clean and all 5 stem-content regressions still pass.

PR #807, opened 2026-08-20
feat(pid): address the stop sentinel to a recording (#804)

silverstein:
The reader-side guard suggested by @MrFreekour in #804. #805 is the writer-side one; both are worth having, for reasons at the bottom.

## The problem

The stop sentinel is a file in a directory shared by every Minutes process on the machine, and the reader was `path.exists()`. A message with no addressee is answered by whoever reads it first. That is how the desktop app terminated a CLI recording it had never started, while capturing nothing itself.

## The contract now

A sentinel may name the recording it is for. A reader honors it only when it is addressed to itself, or unaddressed.

Three cases, each chosen deliberately rather than falling out of the implementation:

**Unaddressed payloads still stop us.** `minutes stop` genuinely means "stop whatever is recording", and every build predating this writes that form, including the 0.18.2 CLI in the report. Refusing them would leave those users unable to stop a recording at all, which is a worse bug than the one being fixed. An unparseable payload is treated identically, because it still means somebody asked for a stop, and stranding a recording the user is actively trying to end is the wrong way to fail.

**A sentinel addressed to another live process is left in place, not cleared.** Not consuming it matters as much as not obeying it: the process it was meant for still has to receive it.

**A sentinel addressed to a dead process is cleared without stopping anything.** Otherwise it outlives its addressee and stops the next recording that reads it, which would be a new silent-loss bug introduced by the fix for a silent-loss bug.

## Why both this and #805

Addressing does not protect a reader too old to look: a 0.18.2 CLI still stops on existence alone, so the exact pairing in #804 needs the writer guard in #805 as well. Conversely, guarding one writer does not cover a writer nobody has thought of, including future code and builds already in the wild. One guards the sender, the other the receiver.

## Verification

Six tests over the decision function, which is split out from the filesystem precisely so this contract is testable. Mutation-verified in both directions: clearing sentinels addressed elsewhere, which is the reported behavior, fails the foreign-recording test; refusing unaddressed sentinels fails the legacy and unparseable tests.

`minutes-core` pid tests, the CLI suite (115) and the desktop suite (346) all pass. clippy and fmt clean.

Not done here, and worth its own change: the startup warning for a version-skewed CLI sharing `~/.minutes`, which is the third suggestion in #804. With addressing in place the skew is no longer dangerous, so that becomes informational rather than protective.

### DIFF
--- a/crates/core/src/pid.rs
+++ b/crates/core/src/pid.rs
@@ -609,25 +609,119 @@ pub fn stop_sentinel_path() -> PathBuf {
     Config::minutes_dir().join("recording.stop")
 }
 
-/// Write the sentinel file to signal the recording process to stop.
+/// Prefix of an addressed sentinel. Anything else is treated as the legacy
+/// unaddressed form.
+const STOP_SENTINEL_TARGET_PREFIX: &str = "stop-pid:";
+
+/// Write an unaddressed sentinel: stop whoever is recording.
+///
+/// Kept because it is the honest representation of `minutes stop`, which
+/// intends to stop the current recording whatever it is, and because older
+/// builds sharing `~/.minutes` only understand this form.
 pub fn write_stop_sentinel() -> std::io::Result<()> {
+    write_sentinel_bytes("stop")
+}
+
+/// Write a sentinel addressed to one recording.
+///
+/// A sentinel is a file in a directory shared by every Minutes process on the
+/// machine, and the reader used to be "does this file exist". That is how the
+/// desktop app terminated a CLI recording it had never started and was not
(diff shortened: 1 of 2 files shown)


silverstein:
Reconciled #805 into this branch so the writer-side ownership guard and reader-side PID addressing now ship together. New exact SHA: 3202006764d7156fbcfc638b3ff1a150821657a2. Local evidence: cargo fmt clean; 3/3 external-stop decision tests; 15/15 PID/sentinel tests; staged diff check clean. The full GitHub matrix is now rerunning against the combined SHA.

PR #808, opened 2026-08-20
fix(policy): create Windows audit leases owner-only (#800)

### LINKED ISSUE #800: Windows: owner-only DACL attestation fails only in the streaming build

silverstein:
Fixes #800.

## Root cause

The Windows audit lease path used CREATE_NEW through ordinary file options, then applied the protected owner-only DACL through the returned handle. A second process could observe the new filename between those two operations. Its fail-closed attestation correctly rejected the inherited descriptor, which is why the same eight-process test alternated between green and four child failures.

## Fix

Create a new lease through CreateFileW with the final protected owner-only security descriptor supplied in SECURITY_ATTRIBUTES. Windows attaches that descriptor before the directory entry becomes visible. Existing leases are still reopened through the retained no-delete-sharing parent capability and attested without repair.

The parent directory handle already denies delete sharing, so its ambient path cannot be renamed or replaced while the create call resolves it. The exact child handle, lexical metadata, stable identity, and parent chain are still re-attested after creation.

## Verification

- cargo fmt --all -- --check
- cargo test -p minutes-core --no-default-features policy_fs::tests::restricted_override_audit -- --test-threads=1 (7 passed)
- cargo check -p minutes-core --no-default-features --target x86_64-pc-windows-gnu
- Full GitHub Windows matrix pending
- Draft diagnostic PR #801 retains the 25-attempt Windows loop for before/after evidence

This keeps the security boundary fail closed and removes the transient non-canonical state instead of retrying or weakening attestation.


silverstein:
Final source review caught and fixed one preservation issue before landing: the new atomic `CreateFileW` path now retains the existing `FILE_FLAG_WRITE_THROUGH` durability guarantee. Exact head is now `c652d0cdeb368b45beab0d189c6e0d87addacde0`. Fresh `cargo fmt --all -- --check`, Windows GNU cross-target `cargo check -p minutes-core --no-default-features --target x86_64-pc-windows-gnu`, and `git diff --check` pass; the hosted matrix is rerunning on this SHA.

silverstein:
Final exact-head receipt: c652d0cdeb368b45beab0d189c6e0d87addacde0 is mergeable and all 23 reported checks are complete with zero failures. Windows and macOS full unit suites, Windows installer, macOS clippy, Ubuntu, policy, compiler, lint, and dependency checks are green.

PR #815, opened 2026-08-21
Fix Bluetooth HFP voice-stem PCM decoding

### LINKED ISSUE #814: Native call capture (ScreenCaptureKit): Bluetooth HFP mic correctly selected but voice stem captured as near-silence (v0.25.1 & v0.25.2)
## Summary

Native call capture (ScreenCaptureKit `SCRecordingOutput`, triggered via the "Call detected" banner) correctly selects the **configured** microphone by exact name (no `#463`/`#467` fallback-to-default behavior triggered), and the **system/remote** stem is captured perfectly — but the **voice/mic stem is captured as near-total digital silence** when the configured microphone is a **Bluetooth HFP device**. This reproduces identically on **v0.25.1** and **v0.25.2** (i.e. not fixed by the 0.25.2 mixer-slot/zero-stuffing fixes in #792/#793/#797).

The same physical Bluetooth microphone captures cleanly through a traditional CoreAudio/AVFoundation path (`ffmpeg -f avfoundation` / `cpal`), so the mic hardware and Bluetooth link are known-good. A control test with the **built-in MacBook mic** (lid open, otherwise identical setup) captures the mic stem perfectly through this same SCStream native-call-capture path — proving the defect is isolated specifically to **Bluetooth/HFP** microphones, not a general SCK mic-capture bug.

## Environment

- macOS 26.5.2 (25F84), Apple Silicon
- Minutes desktop app **v0.25.1**, then re-tested on **v0.25.2** (brew cask, same result)
- Microphone: **Bang & Olufsen Cisco 980**, connected via **Bluetooth**, no USB dongle available
- `system_profiler SPAudioDataType` reports it with `Input Channels: 1` (HFP/mono mode, not A2DP stereo)
- `[recording]` config:
  ```toml
  device = "Bang & Olufsen Cisco 980"
  capture_backend = "cpal"

  [recording.sources]
  voice = "Bang & Olufsen Cisco 980"
  call = "auto"
  ```
- macOS default input **and** default output both explicitly set to the B&O device (verified with `SwitchAudioSource -c -t input/output` immediately before each test)
- Call app: Webex (in `call_detection.apps`, banner fired, "Record" clicked → `intent: "Call"`, `source: "call_detect"` per structured log)
- Microphone + "System Audio Recording Only" both granted to Minutes.app

### Test 1 — v0.25.1, B&O Bluetooth mic, 35 s Webex call

```
system.wav: pcm_f32le, 48kHz mono, 35.00s — mean_volume -39.0 dB, max_volume -7.2 dB  (real, audible remote speech)
voice.wav:  pcm_f32le, 48kHz mono, 34.96s — mean_volume -75.3 dB, max_volume -38.3 dB (noise floor only)
```
Full-duration stem, but no usable speech signal despite the user speaking normally into the B&O mic for the entire call.

### Test 2 — v0.25.2, same setup, 13 s Webex call

```
system.wav: pcm_f32le, 48kHz mono, 13.14s — mean_volume -28.4 dB, max_volume -1.7 dB (real speech, transcribed correctly)
voice.wav:  pcm_f32le, 16kHz mono, 13.04s — mean_volume -90.3 dB, max_volume -70.3 dB (noise floor only)
```
v0.25.2's own new diagnostics correctly detected and reported this in the meeting frontmatter:
```yaml
status: degraded
processing_warnings:
- step: capture
  reason: microphone_audio_not_captured
  message: Microphone audio was not captured (mic stem silent or missing); transcript reflects only call/remote audio
recording_health:
  capture_warnings:
  - kind: !other
      code: native-call-microphone-stem-recovery
    source: system
    message: Microphone audio was missing or silent; the probable cause is microphone device selection or permission.
```
So the honest-degradation / clear-receipt behavior from #806 is working correctly — the recording is not silently lost, and it correctly avoids blaming the wrong stem. The underlying capture defect (why the mic stem is silent in the first place) persists across both versions.

## Ruled out

- **Device selection / `#463` recurrence**: `system_audio_record --list-microphones` returns an exact match for `"Bang & Olufsen Cisco 980"`; `strings` on the shipped binary confirms it passes `--microphone-name` and emits a `configured_microphone` event. The `"configured mic not found, using default"` fallback message was **not** logged in either test, meaning the configured device was found and selected. This is not the device-selection bug fixed in #467.
- **TCC/permissions**: Microphone and "System Audio Recording Only" are both granted to Minutes.app. System-audio capture (same SCStream session, same process) works perfectly in both tests, which would not be possible if TCC were misconfigured.
- **Hardware / Bluetooth link**: the same physical B&O mic captures clean, audible speech when read via `ffmpeg -f avfoundation -i ":Bang & Olufsen Cisco 980"` (traditional CoreAudio HAL path) — verified independently, outside Minutes, immediately before/after these tests.
- **General SCK mic-capture regression**: ruled out by the built-in-mic control test below — the exact same app version/build, config shape, and call flow captures the mic perfectly when the device is non-Bluetooth.

## Confirming control test — non-Bluetooth mic works perfectly (isolates this to Bluetooth/HFP)

I first tried the built-in **MacBook Pro Microphone** as a control, with the lid closed (clamshell mode, dual external monitors) — that test is **invalid and should be disregarded**: a closed lid physically disables the built-in mic (the exact mechanism described in #463), so of course it produced no signal (`voice.wav` truncated to ~10 ms). That result says nothing about Minutes.

I then repeated the control test properly, **with the lid open** so the built-in mic is actually live, same Webex call flow (banner → "Record" → `intent: "Call"`):

```
system.wav: 19.18s — mean_volume -29.9 dB, max_volume -1.7 dB  (real remote speech)
voice.wav:  19.19s — mean_volume -50.3 dB, max_volume -30.3 dB (real, clearly audible speech)
```

Both stems captured cleanly, `diarization_from_stems: true`, and the transcript correctly captured and attributed my own voice:

```yaml
speaker_map:
- speaker_label: SPEAKER_0
  name: Mirek
  confidence: medium
  source: deterministic
```
```
[SPEAKER_0 0:03] testing the microphone, hope it works.
```

No `microphone_audio_not_captured` warning, no `capture_warnings` at all — this recording is entirely healthy. **This is the same app version, same config shape, same call app, same machine — the only variable changed is the microphone (built-in vs. Bluetooth).** This confirms the defect is genuinely isolated to Bluetooth/HFP microphones in the SCStream native-call-capture path, not a general SCK mic bug.

## Suspected root cause

Unlike the system-audio tap (which is format-agnostic to the input route), SCStream's microphone tap likely negotiates a fixed format (commonly 48 kHz) with the mic capture graph. A Bluetooth HFP device is physically limited to 8 kHz (narrowband) or 16 kHz (mSBC/wideband) mono, and — unlike `AVAudioEngine`/CoreAudio's traditional route negotiation — the SCK mic tap may not renegotiate the stream format to match, resulting in a connected-but-effectively-silent tap (matching the -75 to -90 dB noise floor observed rather than an outright stream error). The built-in-mic control test (48 kHz native, no format negotiation needed) working perfectly is consistent with this: the defect only manifests when the tap has to bridge a mismatched/negotiated format, which is exactly the HFP case.

This looks mechanically related to `#792`'s **Bug 6** ("mic stem dies on mid-recording input-device change", also triggered by a Bluetooth device and a 24 kHz vs 48 kHz mismatch) — but that report is on the `cpal` (`current.voice.wav`) capture path, not the SCStream/native-call-capture path (`voice.wav` via `system_audio_record`). If the two share a root cause, a fix for one might not cover the other, since they're different capture backends.

## Ask

- Log the actual negotiated `AudioStreamBasicDescription` (sample rate, format) of the SCK mic tap's first buffer, compared to what the device's nominal sample rate actually is, to confirm/rule out the format-mismatch hypothesis.
- If confirmed, either (a) force a renegotiation/resample step when the mic tap's native format doesn't match the SCK stream's expected format, or (b) for Bluetooth/HFP-class devices specifically, fall back to a traditional CoreAudio input tap for the mic side while keeping SCStream for system audio (hybrid capture), rather than routing the mic through SCK's built-in mixing.
- Given `#792` Bug 6 and this report both point at Bluetooth + SCK/cpal format handling as a recurring theme, a general "Bluetooth mic capture" regression test (mid-call and full-session) might be worth adding regardless of which specific code path is fixed first.

silverstein:
Fixes #814

## Summary

- decode ScreenCaptureKit linear PCM from the complete negotiated ASBD instead of treating every non-float buffer as packed Int16
- support high-aligned 16-bit samples in 32-bit containers, signed Int32, Float32/64, endian flags, interleaved/non-interleaved layouts, and segmented CMBlockBuffers
- log the exact negotiated source ASBD plus the selected microphone's active-format sample rate, including format changes during capture
- withhold unsupported buffers instead of writing a plausible but empty stem
- add a macOS helper self-test containing the HFP-shaped 16-in-32 regression case

## Why this fits #814

The reported voice stem is already 16 kHz, so ScreenCaptureKit did negotiate the headset's HFP rate. The old code then ignored `mBitsPerChannel`, `mBytesPerFrame`, alignment, integer width, and endianness. A 16-bit sample high-aligned in a 32-bit Core Audio container would therefore be read from its low padding and become the full-duration near-silence reported in #814. This patch handles that layout and also records the actual ASBD so hardware validation can confirm or falsify the diagnosis.

## Verification

- `git diff --check`
- local `cargo test -p minutes-app --no-default-features call_capture::tests -- --test-threads=1` (6 passed on Linux; macOS-only helper case skipped)
- local `cargo clippy -p minutes-app --no-default-features --tests --no-deps` (passes; pre-existing warnings remain)
- GitHub CI run 32490127800: every required check passed on exact head `0f4ae6449251c2c0ac7a0392629028f9412688b1`
- macOS full workspace build passed, compiling the Swift helper
- macOS `native_helper_decodes_hfp_pcm_container_layouts` passed
- macOS desktop suite: 350 passed, 0 failed
- macOS integration, full-workspace clippy, format, and desktop capture-relay checks passed
- broad local core run: 1,730 passed, 17 failed, 1 ignored; the failures were unrelated harness/environment prerequisites (mostly no worker-capable `minutes-cli` beside the test harness, plus existing knowledge fixture failures); the clean GitHub platform suites supersede that local environment limitation

## Remaining validation gate

This remains draft until a real Bluetooth HFP capture from the #814 reporter (or equivalent hardware) confirms the voice stem contains speech. CI proves the helper compiles and the suspected container layout decodes correctly; it cannot prove the reporter's headset actually negotiates that layout.

## Hardening (adversarial review, 2026-08-27)

An adversarial pass on the first revision found three real gaps and one math slip, all fixed in `3c90fa38` / `9611e2b4`:

- **Sample-rate change mid-capture** (AirPods A2DP 48 kHz → HFP 16 kHz when the mic engages) previously left the stem `AVAudioFile` at the first buffer's rate, so every later write threw and the stem truncated silently — pre-existing on `main`, but exactly this hardware. Each stem now keeps one processing format and resamples incoming buffers through a single cached `AVAudioConverter`.
- **Withheld buffers** used to vanish, time-compressing the stem against its sibling with no health signal. Undecodable buffers after the stem exists are now written as duration-preserving silence, counted (`mic_frames_silenced` / `system_frames_silenced` in the helper's health JSON), and surfaced as a persisted capture warning naming the source and full ASBD.
- **Helper stderr** was piped but never drained, so a flood of write errors could block the helper's sample queue. It is now drained on a thread with rate-limited tracing (first 5 lines, then one per 10 s with a suppressed count); the helper also reports each write failure once per (source, error).
- **Unsigned PCM normalization** was off by half an LSB (digital silence decoded to 1/65535).

Tests: 9 `call_capture::tests` on macOS including the real helper self-test (rate-change, silence-fill, unsigned-silence cases), the stderr rate limiter, and warning persistence into the queued job.

## Hardware validation

Pending: a Bluetooth HFP capture on AirPods Pro with this build (stem levels + the negotiated-ASBD log line) — will be attached here before the draft is lifted.

### DIFF
--- a/tauri/src-tauri/src/call_capture.rs
+++ b/tauri/src-tauri/src/call_capture.rs
@@ -645,7 +645,7 @@ fn find_native_call_helper_binary() -> Option<PathBuf> {
 #[cfg(test)]
 mod tests {
     #[cfg(target_os = "macos")]
-    use super::native_call_helper_path;
+    use super::find_native_call_helper_binary;
     use super::{
         finish_microphone_startup_handshake, native_call_helper_args, parse_macos_major_version,
         resolve_microphone_selection, MicrophoneSelection, MicrophoneStartupStatus,
@@ -660,7 +660,7 @@ mod tests {
     #[cfg(target_os = "macos")]
     #[test]
     fn native_helper_decodes_hfp_pcm_container_layouts() {
-        let helper = native_call_helper_path().expect("native call helper should be built");
+        let helper = find_native_call_helper_binary().expect("native call helper should be built");
         let output = std::process::Command::new(helper)
             .arg("--self-test-pcm-decoder")
             .output()
--- a/tauri/src-tauri/src/call_capture.rs
+++ b/tauri/src-tauri/src/call_capture.rs
@@ -531,6 +531,18 @@ pub fn start_native_call_capture(
                         let _ = microphone_status_tx
                             .send(MicrophoneStartupStatus::Fallback(name.to_string()));
                     }
+                    Some("audio_format") => {
+                        tracing::info!(
+                            audio_format = %value,
+                            "native call source audio format negotiated"
+                        );
+                    }
+                    Some("audio_format_unsupported") => {
+                        tracing::warn!(
+                            audio_format = %value,
+                            "native call source PCM layout is unsupported; stem buffer withheld"
+                        );
+                    }
                     _ => {}
                 }
             }
@@ -632,6 +644,8 @@ fn find_native_call_helper_binary() -> Option<PathBuf> {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(target_os = "macos")]
+    use super::native_call_helper_path;
     use super::{
         finish_microphone_startup_handshake, native_call_helper_args, parse_macos_major_version,
         resolve_microphone_selection, MicrophoneSelection, MicrophoneStartupStatus,
@@ -642,6 +656,27 @@ mod tests {
         sync::{mpsc, Arc, Mutex},
         time::Duration,
     };
+
+    #[cfg(target_os = "macos")]
+    #[test]
+    fn native_helper_decodes_hfp_pcm_container_layouts() {
+        let helper = native_call_helper_path().expect("native call helper should be built");
+        let output = std::process::Command::new(helper)
+            .arg("--self-test-pcm-decoder")
+            .output()
+            .expect("PCM decoder self-test should launch");
+
+        assert!(
+            output.status.success(),
+            "PCM decoder self-test failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+        assert!(
+            String::from_utf8_lossy(&output.stdout).contains(r#""event":"pcm_decoder_self_test""#),
+            "PCM decoder self-test did not emit its success receipt: {}",
+            String::from_utf8_lossy(&output.stdout)
+        );
+    }
 
     #[test]
     fn parses_major_version_from_product_version() {
--- a/tauri/src-tauri/src/call_capture.rs
+++ b/tauri/src-tauri/src/call_capture.rs
@@ -22,7 +22,100 @@ pub struct CallSourceHealth {
     pub call_audio_live: bool,
     pub mic_level: u32,
     pub call_audio_level: u32,
+    pub mic_frames_silenced: u64,
+    pub system_frames_silenced: u64,
     pub last_update: String,
+}
+
+const HELPER_STDERR_INITIAL_LOG_LIMIT: usize = 5;
+const HELPER_STDERR_LOG_INTERVAL: Duration = Duration::from_secs(10);
+
+#[derive(Debug)]
+struct HelperStderrRateLimiter {
+    initial_lines_logged: usize,
+    suppressed_lines: u64,
+    last_log_at: Instant,
+}
+
+impl HelperStderrRateLimiter {
+    fn new(now: Instant) -> Self {
+        Self {
+            initial_lines_logged: 0,
+            suppressed_lines: 0,
+            last_log_at: now,
+        }
+    }
+
+    /// Returns the number of lines suppressed before the current line when
+    /// the current line should be logged, or `None` when it should be hidden.
+    fn observe(&mut self, now: Instant) -> Option<u64> {
+        if self.initial_lines_logged < HELPER_STDERR_INITIAL_LOG_LIMIT {
+            self.initial_lines_logged += 1;
+            self.last_log_at = now;
+            return Some(0);
+        }
+        if now.duration_since(self.last_log_at) >= HELPER_STDERR_LOG_INTERVAL {
+            let suppressed = std::mem::take(&mut self.suppressed_lines);
+            self.last_log_at = now;
+            return Some(suppressed);
+        }
+        self.suppressed_lines += 1;
+        None
+    }
+}
+
+fn spawn_helper_stderr_drain(stderr: std::process::ChildStderr) {
+    std::thread::spawn(move || {
+        let reader = BufReader::new(stderr);
+        let mut limiter = HelperStderrRateLimiter::new(Instant::now());
+        for line in reader.lines() {
+            match line {
+                Ok(line) => {
+                    if let Some(suppressed) = limiter.observe(Instant::now()) {
+                        tracing::warn!(
+                            helper_stderr = line,
+                            suppressed_since_last = suppressed,
+                            "native call helper stderr"
+                        );
+                    }
+                }
+                Err(error) => {
+                    tracing::warn!(error = %error, "failed to drain native call helper stderr");
+                    break;
+                }
+            }
+        }
+        if limiter.suppressed_lines > 0 {
+            tracing::warn!(
+                suppressed = limiter.suppressed_lines,
+                "native call helper stderr lines suppressed before stream closed"
+            );
+        }
+    });
+}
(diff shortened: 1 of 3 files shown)


silverstein:
## Hardware validation (2026-08-28, dev build `9611e2b4` signed with the stable team identity)

Two native call captures via the desktop "Call detected" banner against a Zoom room, ~15 s of continuous speech each, stems measured with `ffmpeg volumedetect`:

| mic | voice stem | system stem |
|---|---|---|
| MacBook Pro Microphone | 48 kHz Float32 mono, 14.3 s, mean −28.5 dB / peak −5.7 dB | 48 kHz Float32 mono, mean −37.3 dB / peak −7.2 dB |
| Shure MV7+ (USB) | 48 kHz Float32 mono, 14.9 s, mean −31.7 dB / peak −14.8 dB | 48 kHz Float32 mono, mean −37.4 dB / peak −7.2 dB |

Both voice stems are real speech (the #814 failure was ≈ −91 dB, i.e. digital silence). So the hardened decoder + per-stem converter + silence-fill path is regression-free on the two layouts ScreenCaptureKit hands us for built-in and USB mics.

**What this does not prove:** neither device negotiates the Bluetooth HFP layout (mono 16 kHz, 16-in-32) that #814 is about — the AirPods I planned to use went missing. That case rests on the helper self-test (`native_helper_decodes_hfp_pcm_container_layouts`, which encodes the suspected container exactly) plus the new negotiated-format logging that will confirm or falsify the diagnosis on the reporter's headset. I'll ask @miras85 to verify on the release that carries this.

CI on the hardened head: 22/22.

PR #856, opened 2026-08-26
feat(title): use calendar event title as the meeting title (opt-in)

agrawal2674-maker:
## What

Adds an opt-in config toggle so that when a recording overlaps a scheduled calendar event, the event's title is used as the meeting title instead of the AI-generated one.

New key: `[calendar] use_event_title_for_meeting_title` (default `false`).

## Why

When title generation falls back — the LLM is unavailable/offline, or the transcript-first-words heuristic fires — scheduled meetings can end up titled from the first words spoken ("Beep", "Phone Beeps", a greeting). For a meeting that's on the calendar, the real title is already known. This lets users prefer that reliable source.

Note: `auto_title` already preferred a matched calendar event over the transcript-derived title, but the LLM refine step ran afterward and overrode it — so today the calendar title only survives when the LLM path is disabled/fails. This change lets the calendar title win when the toggle is on.

## Behavior

- **Toggle off (default):** unchanged.
- **Toggle on + a matched calendar event** (no explicit `--title`, speech captured): the event title is used verbatim (whitespace-normalized, length-capped) and the LLM title refine is skipped. Logged at `step:"title_generation"` as `outcome:"calendar"`, `detail:"calendar-event-title"`.
- **Explicit `--title`, no matched event, or no speech:** falls through to existing behavior.

## Implementation

- `crates/core/src/config.rs`: new `CalendarConfig::use_event_title_for_meeting_title` (serde default `false`).
- `crates/core/src/pipeline.rs`: `auto_title` uses the event title verbatim under the toggle (bypassing the `title_from_context` heuristic that can otherwise reject a valid event title); a small, testable `title_generation_decision` wrapper short-circuits `maybe_refine_title_with_llm` when the calendar title should win. Existing `select_calendar_event` time-overlap matching is reused as-is.
- `docs/architecture/config.md`: documents the new key.

## Tests

5 new unit tests: calendar wins when enabled; ignored when disabled; skipped without a matched event; explicit title beats calendar; skipped when no speech.

```
cargo test -p minutes-core --no-default-features   # green
cargo fmt --all -- --check                         # clean
cargo clippy --all --no-default-features -- -D warnings  # clean
```

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -827,11 +827,18 @@ pub struct AssistantConfig {
 #[serde(default)]
 pub struct CalendarConfig {
     pub enabled: bool,
+    /// When true, and a recording overlaps a scheduled calendar event, use that
+    /// event's title as the meeting title (overriding the AI-generated title).
+    /// Opt-in; defaults to false to preserve existing behavior.
+    pub use_event_title_for_meeting_title: bool,
 }
 
 impl Default for CalendarConfig {
     fn default() -> Self {
-        Self { enabled: true }
+        Self {
+            enabled: true,
+            use_event_title_for_meeting_title: false,
+        }
     }
 }
 
--- a/crates/core/src/pipeline.rs
+++ b/crates/core/src/pipeline.rs
@@ -4782,10 +4782,7 @@ fn write_transcript_artifact_with_authority(
         if status == Some(OutputStatus::NoSpeech) {
             "Untitled Recording".into()
         } else {
-            calendar_event_title
-                .as_deref()
-                .and_then(title_from_context)
-                .map(finalize_title)
+            calendar_meeting_title(config, calendar_event_title.as_deref())
                 .unwrap_or_else(|| generate_title(&transcript, context.pre_context.as_deref()))
         }
     });
@@ -5374,9 +5371,15 @@ where
         .iter()
         .map(|entity| entity.label.clone())
         .collect();
-    let title_generation = maybe_refine_title_with_llm(
+    // `status == NoSpeech` returned early above, so speech was captured here.
+    // The calendar event matched in phase 1 travels on the artifact's
+    // frontmatter, which is what lets the opt-in calendar title survive the
+    // refine step on the desktop two-phase path.
+    let title_generation = title_generation_decision(
         &artifact.frontmatter.title,
         context.requested_title.as_deref(),
+        artifact.frontmatter.calendar_event.as_deref(),
+        true,
         summary.as_deref(),
         raw_summary.as_ref(),
         &entities,
@@ -6044,24 +6047,8 @@ where
         if status == Some(OutputStatus::NoSpeech) {
             "Untitled Recording".into()
         } else {
-            let calendar_title = if config.calendar.use_event_title_for_meeting_title {
-                // Opt-in: use the calendar event's title verbatim (whitespace-
-                // normalized + length-capped), bypassing the context heuristic
-                // that can otherwise reject an otherwise-valid event title.
-                calendar_event_title
-                    .as_deref()
-                    .map(normalize_space)
-                    .filter(|s| !s.is_empty())
-                    .map(finalize_title)
-            } else {
-                // Default: derive a title from the calendar event via the context
-                // heuristic (which may reject generic names).
-                calendar_event_title
-                    .as_deref()
-                    .and_then(title_from_context)
-                    .map(finalize_title)
-            };
-            calendar_title.unwrap_or_else(|| generate_title(&transcript, pre_context.as_deref()))
+            calendar_meeting_title(config, calendar_event_title.as_deref())
+                .unwrap_or_else(|| generate_title(&transcript, pre_context.as_deref()))
         }
     });
     let entities_started = std::time::Instant::now();
@@ -6524,6 +6511,26 @@ where
     }
 }
 
+/// Resolve the title a matched calendar event contributes, if any.
+///
+/// With `[calendar] use_event_title_for_meeting_title` enabled the event title
+/// is used verbatim (whitespace-normalized, length-capped), bypassing the
+/// `title_from_context` heuristic that rejects generic names like "Meeting".
+/// Otherwise the heuristic decides, matching the long-standing default. Both
+/// the single-pass `process` path and the desktop two-phase path
+/// (`write_transcript_artifact` → `enrich_transcript_artifact`) go through
+/// here so the toggle means the same thing everywhere.
+fn calendar_meeting_title(config: &Config, calendar_event_title: Option<&str>) -> Option<String> {
+    let event_title = calendar_event_title?;
+    if config.calendar.use_event_title_for_meeting_title {
+        Some(normalize_space(event_title))
+            .filter(|title| !title.is_empty())
+            .map(finalize_title)
+    } else {
+        title_from_context(event_title).map(finalize_title)
+    }
+}
+
 /// Decide the meeting title, honoring the opt-in "calendar event title wins"
 /// behavior before falling back to the LLM refine + deterministic paths.
 ///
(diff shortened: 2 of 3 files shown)


silverstein (changes requested):
Thanks for this — clean change, good tests, and the opt-in default is the right call.

One thing blocks it though: as written, the toggle doesn't affect desktop app recordings.

The PR patches `process_with_progress_and_sidecar` (`pipeline.rs:6044` / `:6092`), which is the `minutes process` CLI path and the desktop's retry-recovery path. Normal desktop recordings go through the two-phase job worker instead:

- `transcribe_to_artifact` → `write_transcript_artifact_with_authority` — auto-title at `pipeline.rs:4785`, still the `title_from_context` heuristic
- `enrich_transcript_artifact` — `maybe_refine_title_with_llm` at `pipeline.rs:5377`, still runs and overrides the calendar title

That second call site is exactly where the "LLM refine overrides the calendar title" behavior you describe lives, so the toggle needs to apply there too. The enrich phase has what it needs: `frontmatter.calendar_event` is persisted on the artifact (`markdown.rs:1116`), so `title_generation_decision` can take that as `calendar_event_title`. A test for the two-phase path would be great alongside the existing five.

Small nit: the verbatim branch gates on `calendar_event_title.is_some()`, so an event whose title normalizes to empty falls through to `generate_title` but still logs `outcome: "calendar"` and skips the refine. Gating on the non-empty normalized title (or on whether `auto_title` actually came from the calendar) would keep the log honest.

Happy to re-review once the two-phase path is covered.

agrawal2674-maker:
thanks for the response dudr. i just want to let you know that this peoject
has made me 10x productive. there are so much improvement i would want to
do. i am not a coder. so this is all claude. but i have tons of ideas

silverstein:
Ha — that's genuinely great to hear, and honestly one of the coolest things about this project: a non-coder shipping a clean, tested change to the pipeline through their agent is exactly the kind of contribution I want more of. Keep the ideas coming. The best way to get them in: open an issue describing the idea first (what you want, why), and I'll tell you which part of the pipeline it touches before your agent starts — that would have caught the one gap here.

Speaking of which, I finished this one for you rather than bounce it back — pushed `21fd4b87` onto your branch:

- one shared `calendar_meeting_title()` helper so both auto-title sites (single-pass `process` and the desktop's `write_transcript_artifact`) treat the toggle the same way
- the desktop enrich phase now goes through `title_generation_decision`, reading the matched event from `frontmatter.calendar_event`, so the LLM refine no longer overrides the calendar title on real app recordings
- a blank event title no longer logs as a `calendar` outcome
- a two-phase integration test plus helper/blank-title tests on top of your five

Your commit stays as the first one on the PR; once CI is green I'll merge it.

Out of curiosity, since you've clearly built a real workflow around it: what are you using Minutes for, and how does the day-to-day look — desktop app, CLI, both? Which agent/setup is driving it? Knowing how people actually use it shapes what I build next, and "10x productive" is the kind of claim I'd love to understand better.

agrawal2674-maker:
Hi Mat,

Thanks for finishing up the PR and merging it!

To answer your questions about how I use Minutes and what my workflow looks
like, here is a bit more context:

My Setup & Workflow

1.  Digital Brain: I am trying to build a personal digital brain. Minutes
records every conversation, and I use Claude Code via the CLI to search
through the minutes and extract context. Because of this, features like the
built-in meeting recap aren't as useful to me since I just ask Claude Code
for it, though I imagine it's great for GUI users.
2.  Local & Private: My company doesn't allow corporate AI assistants due
to privacy and regulatory concerns, so I needed something that runs
completely locally on my laptop.
3.  Cost & UX: It is cheap, and the UX is highly intuitive, though I
exclusively use the CLI now instead of the GUI.

Key Challenges & Enhancements I've Ran Into

1.  API Secret Management: There currently isn't a way to save my Anthropic
API secret directly in Minutes. I had to write a custom shell script to
launch Minutes with the secret key injected; otherwise, the CLI isn't
authenticated and summarization fails.
2.  Auto-Record/Pause: Minutes doesn't automatically record yet. I would
love for it to auto-record and auto-pause by detecting Microsoft Teams
usage.
3.  Coach Functionality: I still haven't been able to get the coach feature
to work for me.
4.  Performance: Even on an M5 Mac, I notice my laptop hitting maximum
resource usage during summarization.

Product & Target Persona Thoughts
As a Product Manager, writing out these enhancements makes me curious about
your target persona. Why would someone choose free software when companies
often provide paid tools like Granola to employees for free? I would love
to brainstorm who you see your core users being so I can tailor my
suggestions accordingly.

What are your thoughts on building a local, open-source orchestration layer
that acts as a true personal assistant for employees? Right now, most
companies are building individual tools rather than a digital assistant
that runs locally and orchestrates everything. By moving to this setup,
I’ve stopped paying for Wispr Flow, Granola, and other collaboration tools.
I think there is a massive opportunity to open-source the entire workplace
stack.

Let me know your thoughts!

Best,
Atul Agrawal

silverstein:
Thanks — that's a really useful picture, and the "regulated workplace, local-only, CLI + agent as the interface" persona is one I care a lot about.

I turned your four pain points into tracked issues so they have a home:

- #864 — storing an Anthropic API key in the app for `engine = "claude"` (today it's env-only; the openai-compatible engine already has a Keychain path, so this is extending that)
- #867 — opt-in auto-record / auto-stop on detected calls, Teams desktop first (Teams is already detected; what's missing is acting on it)
- #865 — Coach failing silently; it needs actionable setup errors and a preflight. If you can add what you actually saw when you tried it (and whether Ollama was running), that would help a lot.
- #866 — summarization pegging the machine. Could you paste your `[summarization]` and `[transcription]` config there? Whether it's Whisper or a local LLM changes the fix.

On the persona and the "local orchestration layer" question — that one deserves a real answer from me rather than a quick reply, so I'll come back to it separately.

agrawal2674-maker:
another key issue with using claude code. while i have told it multiple
times to use minutes skill. most of the time it does a file search using
shell command in minutes folder instead of using cli command. so it doesn’t
take into account the recent minutes as its not on the disk yet apparently

silverstein:
That last note is important, thanks. Two separate things in it, if I read it right:

1. **Claude Code greps the meetings folder instead of calling the Minutes CLI/skill.** Could you paste (a) how the skill is installed for you (`~/.claude/skills/minutes`, the plugin, or something else) and (b) one prompt where it went to `grep`/`find` instead of `minutes search` / `list_meetings`? Whether the skill is even loaded vs. loaded-but-ignored changes the fix.
2. **Recent meetings aren't on disk yet.** After a recording stops, transcription + summarization run in a background job, so the markdown file appears a bit later. If the agent searches during that window it will miss the newest meeting. `minutes status` (or the MCP `get_status`) shows the queue — a skill that checks that first is a reasonable ask.

I'll open an issue for (1) once I know the setup; (2) I'll track as "agent should surface in-flight processing".

silverstein:
On your persona question — you asked it at the right moment, and your answer plus another daily user's (#861) landed within a day of each other saying the same thing without knowing it: Minutes is the **capture layer** of a memory you own, an **agent is the interface** (Claude Code for you, Cursor over MCP for them), the destination is a **knowledge base you built** (your digital brain, their Obsidian wiki), and neither of you reads the built-in recap — you ask the agent. That's Persona A. There's a second group that uses the desktop app as a complete notetaker and reads results there — heavily Windows today, where capture parity is still missing. I've written both down in #881 (`docs/personas.md`) with a triage rule: capture reliability serves both and goes first; the MCP/CLI/skill surface is *your* product, so an agent bypassing the skill or missing a meeting still in the processing queue is a product bug, not an integration nit; summarization stays optional everywhere.

Why free software when Granola is handed out for free: Granola is a notetaker you sit in whose memory lives on their servers in their format. For someone in a regulated workplace who can't use a cloud assistant at all, and who wants the record to be theirs and queryable by whatever agent they choose, that's not the same product — it's the thing you can't use. That's the bet.

On the "local orchestration layer for employees" idea — I think the honest version of Minutes' role in that is narrower than the whole stack: be the conversation memory that any orchestrator can read, with policy-bounded access, rather than try to be the orchestrator. The MCP surface and skills are how it plugs in. If you write up what you'd want from that layer concretely (a Discussion is perfect), I'll engage with it properly — you're describing the workflow you actually run, which is worth more than a roadmap slide.

PR #858, opened 2026-08-27
fix: match Safari exactly so the tab probe cannot launch it

### LINKED ISSUE #857: Call detection launches Safari when it is not running (substring match on Safari background agents)
## What happens

With browser meeting detection enabled (`google-meet` or `teams-web` in `call_detection.apps`), Safari opens by itself whenever the tab probe runs, even though Safari is not the default browser and was never opened by the user.

## Why

`detect_browser_meeting` in `tauri/src-tauri/src/call_detect.rs` decides whether Safari is running with a substring match on the process list:

```rust
("safari", "Safari", BrowserKind::Safari, false),
```

macOS always has Safari-named background agents running even when Safari itself is closed. On my machine, with Safari closed:

```
$ ps -axco command | grep -i safari
SafariBookmarksSyncAgent
SafariConfigurationSubscriber
SafariLaunchAgent
SafariLinkExtension
SafariNotificationAgent
SafariWidgetExtension
com.apple.Safari.History
com.apple.Safari.SafeBrowsing.Service
com.apple.SafariPlatformSupport.Helper
```

The substring hit makes the detector treat Safari as running, and `query_browser_tabs` then executes `tell application "Safari"` via osascript, which launches the app when it is not already running.

The Arc entry in the same table already uses an exact match for exactly this reason (substring would catch `searchpartyd` etc.). Safari's binary is exactly `Safari`, so it can get the same treatment.

## Environment

- macOS 26 (Darwin 25.5.0), Minutes 0.25.3
- Default browser is Brave, which is why the launch is so visible: a browser the user never uses pops up mid-work

## Fix

One-line flag change plus a regression test. PR follows shortly; verified locally (with only the background agents running the probe no longer selects Safari, and a real `Safari` process still matches). If you'd rather solve it differently, feel free to close both and I can help with whatever direction you prefer.

adriannoes:
## Summary

- **Bug**: with `google-meet`/`teams-web` detection enabled, the browser tab probe launched Safari even when the user never opened it (`tell application "Safari"` starts the app when it is not running)
- **Cause**: the process check used a substring match, and macOS always runs Safari-named agents (`SafariBookmarksSyncAgent`, `SafariLaunchAgent`, `com.apple.Safari.History`, ...) even with Safari closed
- **Fix**: the Safari entry now requires an exact process-name match, same as the existing Arc entry (`Safari` is the exact binary name)
- The browser table moved into a pure `browsers_to_probe()` helper so the selection is unit-testable; three regression tests cover the agents-only case, the real-Safari case, and the intentional substring behavior for Chrome helper processes
- No behavior change for the ChromeLike entries

Closes #857

## Test plan

- [x] `cargo test -p minutes-app` (376 passed, includes the 3 new regression tests)
- [x] `cargo clippy -p minutes-app --no-default-features -- -D warnings` (clean)
- [x] `cargo fmt -p minutes-app -- --check` (clean)
- [x] Manual on macOS 26: rebuilt and installed the app; with Safari closed and its background agents running, the probe no longer opens Safari

Tested locally and running on my machine, so no urgency on my side. If you'd prefer a different approach, feel free to close this and #857 and I'm glad to rework it.

### DIFF
--- a/tauri/src-tauri/src/call_detect.rs
+++ b/tauri/src-tauri/src/call_detect.rs
@@ -1036,41 +1036,15 @@ impl CallDetector {
         want_teams: bool,
     ) -> BrowserMeetProbe {
         let running_lower: Vec<String> = running.iter().map(|s| s.to_lowercase()).collect();
-        let mut saw_browser = false;
+        let matched = browsers_to_probe(&running_lower);
+        let saw_browser = !matched.is_empty();
 
-        for (proc_fragment, app_name, kind, exact) in &[
-            (
-                "google chrome",
-                "Google Chrome",
-                BrowserKind::ChromeLike,
-                false,
-            ),
-            (
-                "chrome canary",
-                "Google Chrome Canary",
-                BrowserKind::ChromeLike,
-                false,
-            ),
-            ("chromium", "Chromium", BrowserKind::ChromeLike, false),
-            // Arc's binary is exactly "Arc"; substring match would catch
-            // searchpartyd / searchpartyuseragent / TrialArchivingService.
-            ("arc", "Arc", BrowserKind::ChromeLike, true),
-            ("safari", "Safari", BrowserKind::Safari, false),
-        ] {
-            let proc_match = if *exact {
-                running_lower.iter().any(|p| p == proc_fragment)
-            } else {
-                running_lower.iter().any(|p| p.contains(proc_fragment))
-            };
-            if !proc_match {
-                continue;
-            }
-            saw_browser = true;
+        for (app_name, kind) in matched {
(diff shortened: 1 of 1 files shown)


silverstein:
Thanks — clean fix, and the `browsers_to_probe()` split is a nice touch for testability. Verified the assumption locally too: with Safari running, `ps -eo comm=` basenames to exactly `Safari` alongside seven Safari-named agents, so exact match is right. Merging.

adriannoes:
Thank you for the quick review and merge, and for double-checking the process name assumption on your side. I use Minutes daily, so happy to help with future fixes or issues whenever an extra pair of hands is useful. Contributing to open source tools I actually depend on is the fun kind of work :)

PR #522, opened 2026-07-23
feat(summarize): add Cursor Agent CLI as agent_command backend

adriannoes:
## Summary
- Add Cursor Agent CLI (`agent`) as an opt-in `agent_command` backend for post-meeting summarization, title refinement, and L1 speaker mapping (**Fixes https://github.com/silverstein/minutes/issues/520**).
- Headless invocation: `agent -p --mode ask --output-format text --trust <prompt>` — ask-mode stays read-only; `--trust` avoids interactive workspace-trust hangs (same class as Codex/Gemini); no hardcoded `--model`.
- Append `agent` **last** in `detect_agent_cli` so `engine = "auto"` does not change for users who already have another CLI.
- Desktop `WELL_KNOWN_AGENTS` + docs (`cursor-agent.md`, config, agent-integrations, apple-fm auto list).

### Spike notes
- `agent --version` OK (`2026.05.27-fe9a6e2`).
- Headless ask validated: `agent -p --mode ask --output-format text --trust "Reply with exactly: SPIKE_OK"` → `SPIKE_OK`, exit 0; sentinel file untouched (no write side effects).
- Prompt delivery: trailing positional arg (CLI contract). Stdin not used in v1.
- Auth: CLI login / `CURSOR_API_KEY` required for `-p` (IDE login alone may not be enough).

## Test plan
- [x] `cargo test -p minutes-core --no-default-features prepare_agent_invocation`
- [x] `screen_instructions_and_invocation_delivery_stay_in_sync` includes `agent`
- [x] empty-screens test includes `agent`
- [x] `cargo clippy -p minutes-core --no-default-features -- -D warnings` (prior pass)
- [x] Manual SPIKE_OK smoke (authenticated)
- [ ] Optional maintainer smoke: `engine = "agent"` + short meeting

### Maintainer guardrails checklist
- [x] No hardcoded vendor model
- [x] Read-only (`--mode ask`); no `--force` / `--yolo`
- [x] `agent` after existing `detect_agent_cli` entries
- [x] Small PR: invocation + unit tests + docs/WELL_KNOWN

PR #861, opened 2026-08-27
fix: detach the DMG mount from the cleanup trap and make Finder layout best-effort for installs

### LINKED ISSUE #859: create-branded-dmg.sh cleanup trap misses the mounted volume (TMPDIR symlink), leaving the DMG attached on failure
## What happened

While running `./scripts/build.sh --install` from a non-interactive session, the Finder window-layout step in `create-branded-dmg.sh` timed out:

```
Applying Finder window layout...
273:289: execution error: Finder got an error: AppleEvent timed out. (-1712)
```

Fair enough in that context. But the EXIT trap then failed to clean up and left the read/write DMG mounted:

```
rm: /var/folders/.../minutes-dmg.wPVZl3/mount: Resource busy
rm: /var/folders/.../minutes-dmg.wPVZl3: Directory not empty
```

I had to `hdiutil detach` the volume by hand afterwards.

## Why the trap misses the mount

`cleanup()` gates the detach on:

```bash
if mount | grep -Fq "$MOUNT_DIR"; then
    hdiutil detach "$MOUNT_DIR" -force -quiet || true
fi
rm -rf "$WORK_DIR" || true
```

`MOUNT_DIR` comes from `mktemp -d "${TMPDIR:-/tmp}/minutes-dmg.XXXXXX"`, and on macOS `TMPDIR` is a `/var/folders/...` path where `/var` is a symlink to `/private/var`. `mount` prints the resolved `/private/var/folders/...` path, so the grep never matches, the detach is skipped, and `rm -rf` hits a busy mount point. The failed run shows both spellings side by side: `hdiutil attach` reported the mountpoint as `/private/var/folders/...` while the `rm` errors reference `/var/folders/...`.

Since the script runs under `set -euo pipefail`, any failure between `hdiutil attach` and the normal detach path (which has its own retry loop and works fine) goes straight to this trap, so the leftover mount is not specific to the Finder timeout.

## Possible fixes

1. Resolve `WORK_DIR` to a physical path once (`cd "$WORK_DIR" && pwd -P`) so the grep matches
2. Drop the grep gate and always attempt `hdiutil detach ... -force -quiet || true` (already fully guarded)
3. Reuse the retry-then-force detach loop from the main path inside the trap

Happy to send a small PR for whichever direction you prefer.

A related question while I am here: would you take a change making the DMG branding step non-fatal when `--install` was requested (or running the install steps before it)? Today a cosmetic Finder failure blocks the local install that was the point of the invocation. If neither is worth the churn, no hard feelings, feel free to close.

adriannoes:
## Summary

Implements the direction from #859 (options 2 + 3, plus the `--install` question), one commit per point:

- **Cleanup trap** (`create-branded-dmg.sh`): the retry-then-force detach loop moved into a shared `detach_mount_dir()`, used by both the main path and the EXIT trap in place of the broken `mount | grep` gate. The function resolves the physical path once (`pwd -P`), which fixes the `/var` → `/private/var` symlink mismatch, and exits early when nothing is mounted (the trap also fires on successful exits and on failures before the attach, so calling the full retry loop unconditionally would add ~30s of no-op retries there)
- **Best-effort layout** (`create-branded-dmg.sh` + `build.sh`): new `--layout-best-effort` flag turns a Finder window-layout failure into a warning (DMG still produced, window not arranged); `build.sh` passes it only when `--install` was requested, so release packaging keeps the step fatal
- The `${arr[@]+...}` expansion guard in `build.sh` is for the bash 3.2 macOS ships, where an empty array trips `set -u`

Closes #859

## Test plan

- [x] `bash -n` on both scripts
- [x] Full success path: DMG created, no leftover mounts, workdir removed
- [x] Simulated Finder failure (osascript stub on PATH returning the -1712 error) without the flag: script exits 1, the trap detaches the volume and removes the workdir; the #859 leak is gone
- [x] Same simulated failure with `--layout-best-effort`: warning on stderr, DMG still produced, exit 0, no leftover mounts
- [x] Flag plumbing in bash 3.2 with `set -u`: `--install` present and absent both expand correctly
- [x] `./scripts/build.sh --install` end to end on macOS 26: build, DMG, CLI and app installed, exit 0, no leftover mounts

shellcheck did not run (not installed on this machine); happy to address any findings if CI has it.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -148,10 +148,18 @@ from pathlib import Path
 print(json.loads(Path("tauri/src-tauri/tauri.conf.json").read_text())["version"])
 PY
 )"
+# Local installs treat the DMG window cosmetics as best-effort: a Finder
+# AppleEvent failure must not block installing the app that was just built.
+# The ${arr[@]+...} guard keeps set -u happy on the bash 3.2 macOS ships.
+DMG_LAYOUT_ARGS=()
+if [[ " $* " == *" --install "* ]]; then
+    DMG_LAYOUT_ARGS=(--layout-best-effort)
+fi
 ./scripts/create-branded-dmg.sh \
     --app target/release/bundle/macos/Minutes.app \
     --version "$APP_VERSION" \
-    --output "target/release/bundle/dmg/Minutes_${APP_VERSION}_aarch64.dmg"
+    --output "target/release/bundle/dmg/Minutes_${APP_VERSION}_aarch64.dmg" \
+    ${DMG_LAYOUT_ARGS[@]+"${DMG_LAYOUT_ARGS[@]}"}
 
 echo "=== Signing + Installing CLI ==="
 mkdir -p ~/.local/bin
--- a/scripts/check_graph_worker_packaging.mjs
+++ b/scripts/check_graph_worker_packaging.mjs
@@ -9,7 +9,7 @@ import { existsSync, readFileSync } from "node:fs";
 const EXPECTED_SOURCE_SHA256 = {
   release: "3a71697445c183e033afeaa9fc097de990870d934933a2cbba35c2e0289d677a",
   acceptance: "8ae35943181c6d0f248f580587b894c53db9c30257f307c5aa5264a83f13969d",
-  build: "24fca291f683c7c6cc0599e7e88514f4f3374e357983298b857b4fef0acada9f",
+  build: "cbc8570be555a3d5c5679166150cdfaebaa1d15b7eaaea8cda88dc135e77e405",
   dev: "4b8cacb078a06320c9718c6f4d35801125bfa8629e71962caa841a36dcf600b4",
   packageXpc: "450e1cf9d62dc938be45af5f84206fb9d7199f4f865d4f8af37c05f20863573a",
   tauri: "f127e92a1a2a635326d13dd766b3e6cc841655bce30ea2d01d67d9f12c15502c",
--- a/scripts/create-branded-dmg.sh
+++ b/scripts/create-branded-dmg.sh
@@ -89,10 +89,40 @@ STAGING_DIR="$WORK_DIR/staging"
 MOUNT_DIR="$WORK_DIR/mount"
 RW_DMG="$WORK_DIR/Minutes-${VERSION}.rw.dmg"
 
-cleanup() {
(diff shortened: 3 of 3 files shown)


adriannoes:
Ready for review; CI is green.

silverstein:
Verified locally: resealed `build.sh` golden matches, `bash -n` clean, and the `" on $real_dir ("` pattern matches `mount` output on macOS 26 exactly. The not-mounted early exit in the trap is a good call. Thanks for turning this around so quickly.

adriannoes:
Thanks for the careful review and for verifying the mount pattern on your side. Glad the early exit landed well. Count on me for future fixes and contributions; Minutes is part of my daily workflow, so keeping it healthy is in my own interest :)

silverstein:
Likewise — two clean fixes in a day, both with the diagnosis already done, is a gift. Since Minutes is part of your daily workflow: would you be up for sharing a bit about how you actually use it? What you record (calls, meetings, memos), which surfaces (desktop app, CLI, MCP from an agent), what you do with the transcripts afterwards, and the one thing that annoys you most. No pressure and no format required — a few lines here or a Discussion is perfect. That kind of picture from a daily user shapes what gets built next more than anything else.

adriannoes:
Happy to share a bit of the daily picture.

I use Minutes as the capture layer of a local second brain, not as a notetaker I sit in. 

What I record is selective: 
- Interviews, 1:1s, community calls, consulting discovery
- Also short WhatsApp / voice memo drops when someone sends audio
- I don’t record day-to-day internal work meetings. Typical length is 15–80 minutes.

Surfaces: 
- Desktop app for Zoom / Meet / Teams (headphones on; I wait for the “Call detected” banner)
- CLI for vault setup (straight into an Obsidian meetings/ folder) and for processing .opus / .wav. 
- The surface I actually live in afterwards is MCP from a Cursor agent (search_meetings / get_meeting) — that’s how transcripts get folded into a wiki.

Afterwards the file lands in the vault automatically: 
- I run an ingest pass so facts, decisions, and action items become interlinked wiki pages
- I almost never re-read the raw Markdown in Obsidian.. I ask the agent
- Summarization is Cursor Agent (engine = "agent" / agent_command).. I opened a PR for that #522.

Happy to go deeper on any of this.

PR #870, opened 2026-08-27
fix(mcp): stop startup auto-setup from resetting the configured transcription model

### LINKED ISSUE #868: minutes-mcp 0.25.3: startup auto-setup silently resets the configured transcription model to "tiny" on every server spawn
### Environment

- Minutes 0.25.3 (desktop app + Homebrew CLI + minutes-mcp npm, all 0.25.3), macOS 26 (Apple Silicon)
- `[transcription] model = "medium"` configured in `~/.config/minutes/config.toml`, `ggml-medium.bin` downloaded and healthy

### Symptom

Since updating to 0.25.3, `[transcription] model` in `config.toml` keeps reverting from `"medium"` to `"tiny"` many times per day, with no Settings interaction. Meetings silently transcribe on tiny (mangled numbers, destroyed participant names, hallucinated sound tags). Setting medium via the Settings pane, via a file edit, or via `minutes setup --model medium` holds only until the next MCP server spawn, often less than a minute.

### Root cause (verified locally)

Two 0.25.x changes compose into a config-clobbering loop:

1. The 0.25.x CLI wraps `minutes health --json` output in an envelope: `{ "ok": true, "command": "minutes health", "data": { "items": [...] } }`. minutes-mcp's `ensureWhisperModel()` (dist/index.js, the "Auto-setup: download whisper model if missing" block) still expects the old bare-array format: it does `JSON.parse(stdout)` then `Array.isArray(items) && items.find(i => i.label === "Speech model")`. On the envelope this is always falsy, so the ready-check ALWAYS fails, even while the envelope itself says `"Speech model" state "ready", medium installed`.
2. It then falls through to `execFileAsync(MINUTES_BIN, ["setup", "--model", "tiny"], ...)`. Before 0.25.3 this was a harmless re-download no-op. Since #825, `minutes setup` persists the selected model to config.toml (`config setting saved section="transcription" key="model"`), so the fallback now overwrites the user's configured model with "tiny" on every single MCP server startup.

Agent tooling (Claude Code and similar) spawns minutes-mcp on every session start, so this fires dozens of times a day. Confirmed by controlled repro below; also confirmed that `minutes setup --model medium` immediately followed by one `npx minutes-mcp` spawn ends with the file back on tiny within about 10 seconds.

### Repro

1. Set `[transcription] model = "medium"` in config.toml (via Settings, file edit, or `minutes setup --model medium`).
2. Run `npx -y minutes-mcp` for a few seconds and terminate it.
3. `grep '^model' ~/.config/minutes/config.toml` now reads `model = "tiny"` (file rewritten within seconds of the spawn; stderr shows "[Minutes] Whisper model not found - downloading tiny model (~75MB)..." despite medium being installed and healthy).

### Expected

- `ensureWhisperModel()` parses the 0.25.x envelope (ideally accepts both shapes for version skew).
- Auto-setup never downgrades an existing configured model: if it must run `setup`, it should pass the configured model or omit persistence, not hardcode `--model tiny`. Arguably `setup` invoked as an availability check should not rewrite `transcription.model` at all when the config already has a valid value.

### Impact

Silent, recurring transcription-quality degradation for any 0.25.3 user whose MCP server gets spawned regularly (all agent-tool users). Three of my meetings in one day transcribed on tiny before I traced it.

### Related

- #825 ("setup saves the selected speech model") is what turned the pre-existing format mismatch destructive; the CLI-side hardening itself works as designed.
- #818 is the umbrella that spawned #825; different failure mode (config never created on Windows).

silverstein:
## What

Two-part fix for #868 (MCP startup auto-setup silently resetting the configured transcription model):

**CLI (additive, contract-preserving):** `minutes health --json` now reports `data.effective_engine` (via a new public `minutes_core::transcribe::effective_transcription_engine`, which is the existing resolver that turns a retained `parakeet` / `apple-speech` / unselectable `sherpa` into `whisper`) and `data.model` (the configured transcription model). Every existing field is unchanged; the envelope test now asserts the new fields are strings.

**MCP:** `ensureWhisperModel()` in `crates/mcp/src/index.ts` no longer parses TOML at all. It parses the `health --json` envelope (still accepting the legacy bare array) and **mutates config only on positive evidence**:

| evidence | result |
|---|---|
| health failed / `ok: false` / unrecognized shape | skip, log once |
| no `Speech model` item, or its state is not `attention` | skip |
| `Speech model` ready | nothing to do |
| effective engine (or, on an older CLI, configured engine) is not whisper | skip — never download a whisper model for a non-whisper user |
| whisper + CLI reports `model` | `minutes setup --model <that model>` — the only path that can touch an existing config, and it re-persists the same value |
| whisper, older CLI without `model`, config file exists | skip with "run `minutes setup --model <yours>` or upgrade the CLI" |
| whisper, older CLI without `model`, no config file | fresh machine → `setup --model tiny` (the original zero-touch flow) |

Each row is pinned by a test in `crates/mcp/src/index.test.ts`.

## Why

Since `ef2405b3` (Apr 13) `health --json` has printed an envelope, so the MCP's `Array.isArray(JSON.parse(stdout))` ready-check has been false for four months. That was harmless until #825 (Aug 23, shipped in 0.25.3) made `setup --model` persist its choice — after which every MCP spawn (every agent session) rewrote the user's configured model to `tiny`. Full trace in #868.

Rationale deliberately preserved: #825's persistence is untouched (the CLI's `setup` is not changed); the zero-touch first run from `67686e36` still works; the envelope is consumed as the stable contract `ef2405b3` introduced it to be.

## Adversarial review

An adversarial pass on the first revision found three real gaps, all closed here: an existing config *without* a `model` key (CLI default `small`) would still have been persisted to `tiny`; the hand-rolled TOML scanner missed valid TOML (dotted keys, inline tables); and `data.engine` is the *configured* engine, so a retained `parakeet` that actually resolves to whisper was wrongly skipped. The concurrent-spawn download race it also flagged predates this change and is tracked in #871.

One behavior change worth naming: on a fresh machine with the new CLI, auto-setup now downloads the CLI's reported default (`small`, the same default `minutes setup` uses) rather than `tiny`. That is the product default and avoids the quality trap the reporter hit; `tiny` remains the fallback only when an older CLI can't report a model.

## Verification

- `npx vitest run src/index.test.ts -t "Whisper model auto-setup"`: 16 passed; `npm run build` clean
- `cargo test -p minutes-core --no-default-features effective_transcription_engine`, `cargo test -p minutes-cli --no-default-features health_json_envelope`: green
- `cargo fmt --all -- --check`, `cargo clippy --all --no-default-features -- -D warnings`: clean
- Live: `minutes health --json` from this build reports `engine`, `effective_engine`, `model`

## Release

Users get the MCP via `npx`, so this needs a **minutes-mcp 0.25.4** npm release (and the CLI half rides the next CLI release; the MCP is safe against the older CLI in the meantime, per the table).

Closes #868

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -9525,13 +9525,16 @@ life (qmd://life/)
     }
 
     #[test]
-    fn json_envelope_includes_schema_metadata() {
-        let envelope = json_envelope("minutes health", json!({ "engine": "parakeet" }));
+    fn health_json_envelope_includes_effective_transcription_metadata() {
+        let report = health_json_report(&Config::default(), &[]);
+        let envelope = json_envelope("minutes health", report);
         let value = serde_json::to_value(envelope).unwrap();
         assert_eq!(value["ok"], true);
         assert_eq!(value["command"], "minutes health");
         assert_eq!(value["meta"]["schemaVersion"], 1);
-        assert_eq!(value["data"]["engine"], "parakeet");
+        assert!(value["data"]["engine"].is_string());
+        assert!(value["data"]["effective_engine"].is_string());
+        assert!(value["data"]["model"].is_string());
         assert!(value["meta"]["generatedAt"].is_string());
     }
 
@@ -16526,16 +16529,7 @@ fn cmd_health(json: bool) -> Result<()> {
     let items = minutes_core::health::check_all(&config);
 
     if json {
-        let attention_count = items
-            .iter()
-            .filter(|item| item.state == "attention")
-            .count();
-        let report = serde_json::json!({
-            "engine": config.transcription.engine,
-            "all_ready": attention_count == 0,
-            "attention_count": attention_count,
-            "items": items,
-        });
+        let report = health_json_report(&config, &items);
         let envelope = json_envelope("minutes health", report);
         println!("{}", serde_json::to_string_pretty(&envelope)?);
     } else {
@@ -16563,6 +16557,24 @@ fn cmd_health(json: bool) -> Result<()> {
     }
 
     Ok(())
+}
+
+fn health_json_report(
+    config: &Config,
+    items: &[minutes_core::health::HealthItem],
+) -> serde_json::Value {
+    let attention_count = items
+        .iter()
+        .filter(|item| item.state == "attention")
+        .count();
+    serde_json::json!({
+        "engine": config.transcription.engine,
+        "effective_engine": minutes_core::transcribe::effective_transcription_engine(config),
+        "model": config.transcription.model,
+        "all_ready": attention_count == 0,
+        "attention_count": attention_count,
+        "items": items,
+    })
 }
 
 // ──────────────────────────────────────────────────────────────
--- a/crates/core/src/transcribe.rs
+++ b/crates/core/src/transcribe.rs
@@ -488,6 +488,12 @@ pub(crate) fn effective_batch_engine(config: &Config) -> &str {
     }
 }
 
+/// Return the transcription engine that batch/default transcription will
+/// actually use after resolving unavailable or unsupported configured engines.
+pub fn effective_transcription_engine(config: &Config) -> &str {
+    effective_batch_engine(config)
+}
+
 #[cfg(feature = "parakeet")]
 const PARAKEET_LONG_AUDIO_CHUNK_THRESHOLD_SECS: f64 = 60.0;
 #[cfg(feature = "parakeet")]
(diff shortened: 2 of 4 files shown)


silverstein:
Holding this as draft: an adversarial review of the first commit found real gaps — an existing config without a `model` key (CLI default `small`) would still be persisted to `tiny`; the hand-rolled TOML scan misses valid TOML (dotted keys, inline tables); and `data.engine` is the *configured* engine, so a retained `parakeet` that actually resolves to whisper was wrongly skipped. Revision coming: `health --json` gains additive `effective_engine` + `model` fields, the MCP stops parsing TOML, and auto-setup mutates config only on positive evidence (with an explicit fail-safe for the MCP-0.25.4-vs-CLI-0.25.3 skew window). The download race across concurrent MCP spawns is tracked separately in #871.

PR #872, opened 2026-08-27
fix: reattach resumed live transcript stems

silverstein:
## Summary

- keep stalled voice/system StemTail readers alive while the healthy sibling feeds the live sidecar
- re-attach a source to the mix in the same poll where new frames resume
- require a full two-second 16 kHz sibling window before declaring a source stalled
- add a growing-WAV regression covering voice active, voice gap, system-only continuation, and voice rejoin

## Why

PR #869 correctly prevented a header-only or stalled source from blocking the healthy stem, but it set the stalled tail to None. A transient device or route interruption could therefore remove that source from the live transcript for the rest of the call even if frames resumed.

This follow-up changes live-sidecar participation only. It does not change native recording or saved stem preservation.

## Verification

- cargo fmt --all -- --check
- cargo test -p minutes-core --no-default-features --lib stem_tail::tests -- --test-threads=1 (18 passed)
- git diff --check

The broad no-default core suite was also exercised locally; the environment reported existing worker/FFmpeg harness prerequisite failures unrelated to stem_tail.rs. GitHub CI remains the cross-platform authority.

No Minutes Dev app was built, installed, restarted, or interrupted.

### DIFF
--- a/crates/core/src/stem_tail.rs
+++ b/crates/core/src/stem_tail.rs
@@ -244,7 +244,13 @@ const SIBLING_STEM_GRACE: std::time::Duration = std::time::Duration::from_secs(3
 /// not let the stalled source block the healthy source forever. This also
 /// covers a header-only microphone stem when macOS grants capture but delivers
 /// no microphone frames.
-const STEM_STALL_GRACE: std::time::Duration = std::time::Duration::from_secs(2);
+const STEM_STALL_GRACE_SECS: u64 = 2;
+const STEM_STALL_GRACE: std::time::Duration = std::time::Duration::from_secs(STEM_STALL_GRACE_SECS);
+
+/// Require the healthy sibling to have advanced through the full stall grace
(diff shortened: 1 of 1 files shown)


silverstein:
Review (from the lane that raised the concern on #869): this is the right shape. The excluded tail stays open and is still polled; re-attach happens in the same round frames resume (`poll_source` → `reattach_resumed_source`); and a stall now requires the sibling to have advanced a full grace window (`STEM_STALL_MIN_ADVANCE_SAMPLES` = 2 s at 16 kHz) instead of a single late frame — so a both-sources pause during a device switch declares nothing, and a one-sided gap recovers. Alignment restarts from the resume point, which is fine for provisional drafts. The regression test drives a real `StemTail` on a growing WAV through stall → system-only → rejoin. Live-sidecar only; recording isolation untouched. Good to merge once CI is green.

Issue #888, opened 2026-08-29
MCP lockfile integrity check rejects any branch-local SDK source change


silverstein:
`scripts/install_mcp_dependencies.mjs` (run by the `Agent Corpus + MCP Lockfile` CI job) packs the local `crates/sdk` and requires the tarball integrity to equal the *published* `minutes-sdk` pinned in `crates/mcp/package-lock.json`. Since `dist/` is rebuilt from source before packing, **any change under `crates/sdk/src` fails the job** — not just dependency drift. Verified locally on a branch touching only `crates/sdk/src/corpus-lease.ts`: `local minutes-sdk 0.25.4 integrity does not match the MCP lockfile`.

Nothing under `crates/sdk/src` has changed since v0.25.3, which is why this hasn't surfaced; it bit #778 (a devDependency bump changed `package.json`) and now the SDK half of the #802 corpus-lease fix, which has to be deferred until it can ride a release.

The check conflates two things: (a) "the MCP must resolve the published SDK, not a `file:` link" (a real invariant for the shipped bundle) and (b) "the SDK on this branch is byte-identical to the last publish" (not an invariant anyone wants between releases).

Proposed: when the branch's `crates/sdk` differs from the pinned published version's content, install the MCP against the locally packed SDK for test purposes (patching only the disposable checkout lock, as the Windows path already does) and surface a clear notice; keep the strict published-integrity requirement for release workflows (`check_version_sync.mjs --release` and `release-publish.yml`), where it belongs. Related: #778, #802.

PR #889, opened 2026-08-29
test(mcp): CI-only authorization budget for the corpus-lease flake

silverstein:
Tracks #802 (the corpus-lease timing family; failed #831 and #886 this week).

The corpus authorization handshake has a fail-closed **15 s** deadline and caps any caller timeout at that value, so a loaded hosted runner finishing at 15.001 s fails with *authorization budget overrun by 1ms*. Measurement on the branch: worker-backed filesystem attestation dominates (10.5–12.8 s under load on macOS); the protocol itself is ~30 ms.

## What

- `MINUTES_CORPUS_AUTH_TIMEOUT_MS` is honored **only** when `NODE_ENV=test` *and* `MINUTES_TEST_HARNESS=1` are both set, clamped to 120 s; invalid values fail closed. Production behavior is unchanged (15 s cap).
- The CI "SDK/MCP corpus policy tests" steps and `scripts/run_corpus_test_harness.mjs` (used by the MCP `npm test` scripts) set that pair with a 60 s budget; vitest `testTimeout` moves to 75 s so the lease, not the harness, still reports real timeouts.
- Successful authorizations log their duration (content-free) so we can see how close CI runs to the cap.
- Tests: gate ignored without both markers; applied and clamped with them; success diagnostic reports a duration.

## Deliberately deferred

The SDK's twin `corpus-lease.ts` gets the same gate in a follow-up: any `crates/sdk/src` change currently fails the MCP lockfile integrity job until the SDK is republished (#888). The observed flakes were in the MCP copy.

## Verification

Codex's run: exact MCP policy suite 257 passed / 1 platform skip; MCP build, mirror parity, version sync green. On my Mac the corpus-lease suite fails identically on untouched `main` (filesystem attestation >5 s per attempt — this machine's known descriptor problem), so CI is the meaningful gate here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -615,6 +615,10 @@ jobs:
 
       - name: SDK corpus policy tests
         working-directory: crates/sdk
+        env:
+          NODE_ENV: test
+          MINUTES_TEST_HARNESS: '1'
+          MINUTES_CORPUS_AUTH_TIMEOUT_MS: '60000'
         run: npx vitest run src/corpus-lease.test.ts src/secure-read.test.ts
 
       - name: Build SDK production module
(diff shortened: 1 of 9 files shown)


silverstein:
Superseded by #894 — same commits (MCP + SDK mirror of the CI-only authorization budget, plus the checklist note). This PR stopped tracking its branch after the rebase; the branch ref moved but the PR object didn't, and it turned out the PR had been closed by an earlier close/reopen attempt that only half-applied.

PR #932, opened 2026-09-04
feat: add PocketStation application capture

Raphjacksun7:
## Summary

Adds PocketStation as an optional application-audio backend on Windows and Linux.

Minutes continues to own microphone capture, alignment, recording, transcription, and every later processing step. PocketStation supplies only the selected application's audio through the existing `SystemAudioBackend` interface. The current macOS capture code is unchanged.

## Use

Build with the opt-in feature:

```bash
cargo build --release -p minutes-cli --features pocketstation-capture
```

Then configure the application by exact name, application ID, or process ID:

```toml
[recording]
capture_backend = "pocketstation"

[recording.sources]
voice = "default"
call = "Zoom"
```

## Implementation

- resolves `pocketstation 1.1.8` from crates.io with `native-capture` only;
- does not add Opus or Relay dependencies to Minutes;
- requests 10 ms application-audio frames;
- converts PocketStation's 48 kHz frames into Minutes' existing 16 kHz mono chunks;
- keeps application audio and the existing microphone source separate;
- reports full downstream queues and capture failures through existing Minutes diagnostics;
- rejects `call = "auto"` so an ambiguous application is never selected silently.

## Verification

- Linux ARM64: the proof downloaded `pocketstation 1.1.8` from crates.io, captured a selected PipeWire application, and wrote separate application and microphone stems through Minutes' existing recorder. All 80 100 ms slots remained aligned.
- Windows 11 ARM64: selected per-process WASAPI audio and the configured microphone reached the same recorder as separate sources. This VM run proves behavior, not latency.
- A separate Windows check resolves `pocketstation 1.1.8` from crates.io, confirms that the capture-only dependency tree contains neither `opus` nor `audiopus_sys`, and passes all eight focused adapter tests.
- Focused Rust tests, the feature-only CLI build, formatting, and Clippy with warnings denied pass.

The virtual test sources are deterministic fixtures. They do not claim physical-device or latency qualification.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,19 @@ jobs:
         # breakage here only surfaces when Prompter next runs cargo build.
         run: cargo build -p minutes-core --no-default-features --features streaming
 
+      - name: Build PocketStation capture (Windows and Linux)
+        if: runner.os == 'Windows' || runner.os == 'Linux'
+        run: cargo build -p minutes-core --no-default-features --features pocketstation-capture
+
+      - name: Test PocketStation capture (Windows and Linux)
+        if: runner.os == 'Windows' || runner.os == 'Linux'
+        run: cargo test -p minutes-core --no-default-features --features pocketstation-capture --lib -- --test-threads=1 --nocapture
+        timeout-minutes: 15
+
+      - name: Clippy PocketStation capture (Windows and Linux)
+        if: runner.os == 'Windows' || runner.os == 'Linux'
+        run: cargo clippy -p minutes-core --no-default-features --features pocketstation-capture -- -D warnings
+
       - name: Build core with parakeet (cross-platform)
         # Guards against Linux-parakeet breakage that today only surfaces
         # at tag push via release-cli.yml. The parakeet feature compiles
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4066,6 +4066,7 @@ dependencies = [
  "objc2-app-kit",
  "objc2-av-foundation",
  "ort",
+ "pocketstation",
  "pyannote-rs",
  "rusqlite",
  "schemars 1.2.2",
@@ -5277,6 +5278,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "pocketstation"
+version = "1.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "419bd4c5dac0735627be925273e2c92d4ab4c8d75982c248707797ff01eaafa5"
+dependencies = [
+ "alsa",
+ "cc",
+ "cpal",
+ "hound",
+ "libc",
+ "libloading 0.8.9",
+ "pipewire",
+ "rtrb",
+ "serde",
+ "serde_json",
+ "thiserror 1.0.69",
+ "tokio",
+ "wasapi",
+ "windows 0.58.0",
+ "windows-core 0.58.0",
+]
+
+[[package]]
 name = "polling"
 version = "3.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5984,6 +6008,12 @@ dependencies = [
  "untrusted",
  "windows-sys 0.52.0",
 ]
+
+[[package]]
+name = "rtrb"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fae8ee26b0371a29a77d2b2d6b3ae13aa81def6f9bf1b1b92a32d279a5e709b7"
 
 [[package]]
 name = "rusqlite"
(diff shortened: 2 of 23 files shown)


silverstein (changes requested):
Raphael, thank you for the thoughtful contribution. Per-application capture without a virtual audio device would be genuinely useful for Minutes users on Windows and Linux, and I appreciate that you kept Minutes in control of the microphone, recording, and later processing.

I walked through the adapter and the current CI at exact SHA `27fee255`. The overall shape is promising, but I think we need three safeguards before merging this into the recording path:

1. Please add a Windows and Linux CI step that explicitly builds and tests `minutes-core` with `--no-default-features --features pocketstation-capture`, plus a strict Clippy step for that feature. All current checks are green, but the existing CI only enables `streaming`, so the new feature-gated module is not compiled or tested by those checks.

2. Please make shutdown prompt and cover it with a regression test. `PocketStationCaptureStream::drop` waits on the worker, and the worker calls the orderly `RunningSession::stop()`. PocketStation documents that orderly stop may drain accepted work before joining. Minutes currently waits for that drop before it seals the WAV and stem files, so a stuck backend could leave the user on Stop Recording and delay file finalization. Since this route only feeds polled audio, cancellation looks like the more appropriate shutdown policy, but I am happy for you to choose another bounded design if there is a better PocketStation-native option.

3. Please prove or adjust the shared-timeline claim. The adapter currently discards PocketStation frame timing, assigns `Instant::now()`, and starts its chunk counter at zero. Minutes starts the microphone first and then pairs both sources by chunk index. A delayed PocketStation startup could therefore pair its first application chunk with the microphone first chunk while shifting the actual audio in time. A deterministic test with an intentional source startup delay would resolve this. If the current behavior is intentional, please narrow the documentation rather than promise that a late source is padded instead of shifted.

I do not think this PR needs to turn PocketStation on in the normal release builds yet. Keeping it clearly experimental and opt-in is a sensible first landing. Once the three points above are covered, I would be glad to take another quick pass.

Raphjacksun7:
Thanks, Mat. These were good catches. Let me know if there’s anything still missing from these changes.

silverstein:
Raphael — you turned these around in two days and then I left you waiting three weeks. That's on me, and I'm sorry. Thank you for the patience.

I've gone back through all three points and they're covered:

**1. CI.** The build, test, and `-D warnings` Clippy steps for `--features pocketstation-capture` are there for Windows and Linux. The feature-gated module is actually compiled and exercised now, which it wasn't before.

**2. Shutdown.** This was the one I cared most about, and cancellation is the right call. `running.cancel()` before the join, plus a regression test asserting that dropping the stream "must not wait for orderly delivery" — that's exactly the invariant I wanted held. A wedged backend can no longer strand someone on Stop Recording with unsealed files.

**3. Timeline.** You took the narrow-the-documentation option, and you wrote it plainly: this backend doesn't measure a common start time between the microphone and application capture, so a late application start isn't backfilled with silence. I'd rather ship an honest limitation than an unproven claim, so I'm accepting it as written. Worth revisiting if this graduates out of experimental — a measured common start time is the obvious next step, but it's not a condition of landing.

Merging. Staying opt-in behind the build flag and out of release builds, which is the right first landing for this.

Genuinely useful contribution — per-application capture without a virtual audio device is something people have asked for, and you kept Minutes in control of the microphone and the recording path throughout, which made this reviewable in the first place.

silverstein:
Quick note on what I just pushed to your branch, so nothing looks mysterious.

Main moved seven times today and this branch fell into conflict, so I merged main in rather than make you do another round. Three conflicts, none of them in your adapter:

- **`crates/core/src/screen.rs`** — we both fixed the same cfg/return-shape bug independently. I took main's version because it's a superset: it also threads a `width` parameter that landed after you branched. Your fix was right, it just got there second.
- **`Cargo.lock`** — regenerated. Re-resolves `pocketstation =1.1.10` and keeps a `nnnoiseless` change from earlier today that closes an `atty` advisory.
- **`scripts/check_graph_worker_packaging.mjs`** — resealed the `cliCargo` golden hash. It covers `crates/cli/Cargo.toml`, which both sides touched, so your earlier reseal was stale regardless. Only delta against main is your one feature line.

Your work is untouched — both feature declarations, the gated `mod pocketstation_capture`, the adapter, the tests, the CI steps, the docs. `cargo check` is clean and the packaging guard passes.

Merging once CI goes green.

Raphjacksun7:
No problem at all, Mat — and thank you for the thoughtful review and for taking care of the final merge conflicts. I’m very glad to contribute to Minutes, and I’m looking forward to seeing what great things we can build together!

silverstein:
Glad to have it in. The capture path you added is already getting attention from elsewhere: #1055 is turning on `--all-targets` clippy across the supported routes, and the PocketStation feature is one of the jobs it covers, so that code will be getting linted harder than it was when it landed.

If you want more in that area, #1057 is a Bluetooth HFP mic stem coming out digitally silent on the core-audio-tap backend. Different capture route, same neighborhood.

PR #985, opened 2026-09-14
Fix graph projection under high file limits

### LINKED ISSUE #984: Graph projection commands fail when RLIMIT_NOFILE exceeds 100k
# Graph projection commands fail when RLIMIT_NOFILE exceeds 100k

## Summary

- `minutes people`, `minutes commitments` and `minutes vocabulary rebuild` fail with `Error: policy graph inputs could not be captured` whenever the soft descriptor limit is above 100000.
- `raise_descriptor_limit` in `crates/core/src/graph/macos_kqueue.rs` reads `RLIMIT_NOFILE` to scan `0..rlim_cur` counting open descriptors, and returns an error instead of bounding the scan when the ceiling is too high.
- `crates/core/src/graph_worker.rs:1277` drops the underlying `GraphError` in `map_err(|_| ...)`, so the real cause reaches nobody: not the CLI output, not `minutes logs`, not `minutes events`, not `RUST_LOG=debug`.
- Any Node parent process triggers this, because Node raises `RLIMIT_NOFILE` toward the hard limit at startup. That includes Claude Code, so the Minutes Assistant surface (`Minutes.app` -> `claude` -> shell) hits it on every session.
- Workaround: `( ulimit -n 8192; minutes people )`.

## Reproduction

Verified on macOS 26.6.2 (Darwin 25.6.0, build 25G83), arm64, minutes 0.26.2, both the shipped `/Applications/Minutes.app/Contents/MacOS/minutes` and a `cargo build --release -p minutes-cli` from source at the same version.

```
$ ulimit -n
1048576
$ minutes people
Error: policy graph inputs could not be captured

$ ( ulimit -n 8192; minutes people )
INFO Index rebuilt people=596 meetings=94 commitments=965 topics=646 aliases=217 elapsed_ms=77
TOP CONTACTS (by relationship score)
  ...
```

The threshold is exact: it fails above 100000 and succeeds at or below.

## The offending check

```rust
// crates/core/src/graph/macos_kqueue.rs
let scan_ceiling = usize::try_from(limit.rlim_cur)
    .map_err(|_| io::Error::other("macOS graph descriptor limit was not representable"))?;
if scan_ceiling > 100_000 {
    return Err(io::Error::other(
        "macOS graph descriptor occupancy could not be bounded",
    ));
}
let mut open_count = 0usize;
let mut high_water = 0usize;
for descriptor in 0..scan_ceiling {
    let status = unsafe { libc::fcntl(descriptor as libc::c_int, libc::F_GETFD) };
    ...
}
```

Refusing a 1M-iteration `fcntl` scan is reasonable. Failing the whole projection because of it is not: a high `rlim_cur` means descriptors are plentiful, which is the case where the headroom calculation matters least.

## How the real error was obtained

The message is only reachable by patching the discard:

```rust
// crates/core/src/graph_worker.rs:1277
- .map_err(|_| "policy graph inputs could not be captured".to_string())?;
+ .map_err(|e| format!("policy graph inputs could not be captured: {e}"))?;
```

```
Error: policy graph inputs could not be captured: IO error: macOS graph descriptor occupancy could not be bounded
```

At least five distinct failures inside `capture_policy_graph_snapshot` (missing directory, unavailable lease, journal failure, corrections snapshot, source collection) currently collapse into that one sentence.

## Suggested fixes

| # | Topic | Suggestion |
|---|---|---|
| 1 | Descriptor scan | Bound the scan instead of failing: `let scan_ceiling = scan_ceiling.min(100_000)`, or get the high-water mark from `proc_pidinfo(PROC_PIDLISTFDS)` and skip the linear walk |
| 2 | Error opacity | Keep the `GraphError` in the message, or log it before mapping. The current `map_err(\|_\| ...)` makes every failure in this path undiagnosable |
| 3 | Note | A `DirNotFound` also surfaces as the same sentence, which sends people looking at their corpus when the directory simply does not exist |

## Impact

The Minutes Assistant runs as a Claude Code session launched by `Minutes.app`, so it inherits Node's raised limit. Every assistant session that reaches for `people`, `commitments` or `vocabulary rebuild` gets the generic failure. `search`, `actions`, `consistency`, `list` and `insights` are unaffected.

Happy to send a PR for this if you think the approach above is the right one.

silverstein:
`minutes people`, `minutes commitments`, and vocabulary rebuild failed whenever a Node or Claude parent raised macOS `RLIMIT_NOFILE` above 100,000. Journal setup rejected the large occupancy scan before opening any vnode.

For an already-high soft limit, this skips the linear occupancy scan and proceeds with the existing bounded journal construction. The 8,192-watch cap remains unchanged, low descriptor limits retain the occupancy-aware raise logic, and individual vnode opens still fail closed if capacity is actually exhausted. The outer error remains path-neutral because the underlying graph error may contain local filesystem paths.

Thanks @adriannoes for the precise reproduction and root-cause report.

Fixes #984

Validation:
- macOS 26.6.2 arm64: all 7 `graph::macos_graph_journal::tests` pass
- macOS 26.6.2 arm64: `ulimit -n 1048576; minutes people` succeeds against an isolated synthetic corpus
- `cargo clippy -p minutes-core --no-default-features -- -D warnings` passes on macOS and Linux with Rust 1.95.0
- Linux core suite reached 1,795/1,800 passing in parallel; the four unrelated `knowledge.rs` filesystem flakes all passed when rerun serially (one test is ignored by design)


adriannoes:
I’m glad we were able to solve this problem today, @silverstein ! :)

PR #997, opened 2026-09-16
i18n: add Brazilian Portuguese (pt-BR) and generalize the locale runtime to N languages

PedroPackness:
## What

Adds Brazilian Portuguese as a third UI language, and generalizes the existing localization runtime from "English plus one" to N locales so further languages are a data-only addition.

## Why

The i18n runtime shipped with a single extra locale hardcoded throughout: `valueFor()` returned early unless the locale was literally `'zh-CN'`, the catalog was a pair of module-level `exactMap` / `patterns` globals, the CJK font fallback was keyed to the locale rather than to the script, and `initialLocale()` / `setLocale()` compared against a two-value allowlist. Adding a language meant editing all of those.

Rather than add a second set of hardcoded branches, this makes the locale a lookup key.

## Changes

**Rust (`crates/core/src/i18n.rs`)**
- `Locale::PtBr` plus `locales/pt-BR.json` (48 strings: tray, native menus, notifications)
- catalog lookup is table-driven instead of a `zh`-only branch in `tr()`
- `resolve_locale` maps any tag starting with `pt` to `PtBr`, so `language = "auto"` picks it up from a pt-BR OS locale
- `parse_catalog` is shared, and logs which locale failed to parse

**Frontend (`tauri/src/scripts/i18n.js`)**
- `SUPPORTED` list and a per-locale catalog map replace the module-level globals
- `hasTranslation()` now tests the union across all shipped catalogs. This matters: a text node captured while the UI is in English must still be translatable after an in-place language switch, so capture cannot be locale-dependent
- the CJK font fallback is keyed to a `CJK_LOCALES` set, leaving the Latin type stack untouched under pt-BR
- `initialLocale()` falls back to `navigator.language` for `pt`

**Catalog (`tauri/src/locales/pt-BR.js`)**
958 strings and 27 interpolation patterns, covering the main window, the Coach onboarding sheet and settings, the dictation overlay, the meeting prompt, the command palette and the terminal window.

**Wiring**
`pt-BR.js` is loaded by all six frontend windows, `Português (Brasil)` is added to the Settings language picker, and `cmd_set_ui_language` accepts and resolves the new tag.

## One upstream fix included

`zh-CN.js` has a `"Have questions about this "` key with a trailing space. The runtime trims before lookup, so that key can never match — the string renders in English under zh-CN today. It is also built by concatenation with the content type, so the node is `"Have questions about this meeting?"`, not the prefix. The pt-BR catalog handles it as four exact keys, one per content type. The zh-CN catalog is left alone; flagging it here rather than changing it in a PR about a different language.

## Notes for review

- No English source string was modified, so nothing changes for existing users. Locale resolution is unchanged for `en` and `zh-CN`.
- Translation choice worth knowing about: `Record Memo` and `Quick Thought` are the same action (the capture-mode picker's `memo` segment forwards its click to the hidden Quick Thought button). Translating them as two different Portuguese terms would invent a third name for one feature, so both read "ideia rápida" in pt-BR. `Debrief Memo` / `Decision Memo` keep "memorando" — there the word means a written document, not a voice capture.
- Uncovered strings are deliberate: keyboard shortcuts, URLs, env var names, shell commands, brand and feature names (Minutes, Coach, Recall), and concatenation fragments that never form a whole text node. Fragments whose joined result *is* a node are handled as patterns.

## Testing

- `cargo fmt --all -- --check` clean
- `cargo test -p minutes-core --no-default-features`: 1830 passed. Two `knowledge::` tests fail under parallel execution with `NotFound` errors, and a different pair fails on each run; all 169 `knowledge::` tests pass with `--test-threads=1`. Pre-existing flakiness in a module this PR does not touch.
- `cargo clippy --all --no-default-features -- -D warnings` reports 4 `chunks_exact_to_as_chunks` errors in `transcribe.rs`, `apple_speech_worker.rs`, `stem_probe.rs` and `voice.rs` — all files untouched here. These fire on Rust 1.98 but not on the 1.95 pinned in `rust-toolchain.toml`; I do not have rustup installed, so this is the local-vs-CI drift `CONTRIBUTING.md` warns about, not a regression. Left unfixed to keep the diff focused.
- Built and ran the desktop app end to end: verified the embedded asset is byte-identical to the source catalog, that the picker switches language in place without a reload, and that switching back to English restores the original strings losslessly.


silverstein:
Reviewed the diff in full. This is the right shape: making the locale a lookup key instead of adding a second set of branches is what I would have asked for, and the write up saved me a lot of reading.

What I checked and agree with:

- `tr()` is now table driven and English returns early, so the hot path did not get slower for the default locale
- `resolve_locale` keeps `zh` and `en` behavior byte identical; only a new `pt` branch was added
- `hasTranslation` taking the union across catalogs is correct and load bearing, and your comment explains why. A node captured in English has to stay translatable after an in place switch, so capture cannot be locale dependent
- `CJK_LOCALES` keyed by script rather than locale means pt-BR keeps the Latin stack
- all 27 interpolation patterns are anchored with `^...$`, so none of them can reach into a meeting title or transcript text. That was my main risk when I saw a pattern list this long
- the catalog is wired into all six windows, the picker, and `cmd_set_ui_language`, with `locale_tag` updated so the round trip is closed

Good catch on the zh-CN `"Have questions about this "` key, and thank you for leaving it alone rather than fixing it inside a PR about a different language. Filed as its own issue.

On the clippy noise you saw: those four `chunks_exact_to_as_chunks` errors are local toolchain drift, not yours. `rust-toolchain.toml` pins 1.95 and CI honors it. Nothing to do here.

One gate before this merges, and it is on me, not on you. CLAUDE.md requires a click test in the dev app bundle for any change to `tauri/src/index.html`, and that has to happen on a Mac. CI is green across the board and the code review is done, so that verification pass is all that is left. I will run it and report back.

silverstein:
The zh-CN key is #999.

silverstein:
Merged. Thank you for this one, and for the write up that made it reviewable.

For the record, here is what the verification pass actually covered, since it was done with computer-use automation against the installed dev app rather than by hand.

Verified:

- pt-BR renders in place with no reload. The Settings modal came back as Configurações / Geral / Captura / Transcrição / IA e privacidade / Avançado, with IDENTIDADE, Seu nome, Seus endereços de e-mail, Variações do nome and Idioma all translated
- user data was untouched while the chrome moved around it: name, email list and name variants stayed exactly as entered. This was the property I most wanted to see, and it held
- layout survives the longer Portuguese strings, with no clipping or overflow in the modal
- the Latin type stack stayed put, which is the CJK_LOCALES change doing its job, and the Geist Mono tabs and serif heading are unchanged
- the picker itself lists all four options correctly, including Português (Brasil)

Not verified, and worth saying plainly: I did not re-confirm zh-CN rendering or the round trip back to English through the UI. The synthetic key events would not reach the WebKit select, so those two clicks did not happen. I am comfortable merging on that basis because the PR does not touch the capture or restore mechanism that makes the round trip lossless, and because widening hasTranslation only captures more nodes, each of which still records its own English source. If anything does strand a string, it will be in that path, so that is where I would look first.

Follow ups from your notes are filed: #999 for the zh-CN key, and the catalog drift problem more generally is the thing I want to solve before a third language lands. There is no check today that a new English UI string has a catalog entry, which is exactly how the zh-CN key went unnoticed.

PR #1023, opened 2026-09-17
fix(config): preserve newer settings and support isolated voice profiles

silverstein:
## Summary
- Preserve unknown config tables, nested fields, comments and unchanged arrays during full saves; honor explicit removal of known optional settings.
- Refuse malformed/incompatible files and symlinks; atomically replace using a private temporary file and detect intervening changes.
- Add MINUTES_CONFIG_PATH for process-scoped voice testing without changing other tools XDG settings.

## Verification
- 66 config tests pass on Mac, including five new regression tests.
- Core voice-feature Clippy passes with warnings denied; formatting passes.
- Private local voice profile provisioned and shared config confirmed unchanged.
- Signed Minutes Dev installation and native persistence acceptance in progress.

Edited arrays remain atomic replacements. This does not retrofit already installed older binaries or claim a cross-process transaction protocol.


silverstein:
Native Mac acceptance passed. Canonical scripts/install-dev-app.sh installed source 6795e0ec7ae3113a34c9dfeae6178beef13d36a5 (existing desktop voice branch plus this patch) as ~/Applications/Minutes Dev.app, bundle com.useminutes.desktop.dev, Apple Development signing authority, TeamIdentifier 63TMLKT8HN. Strict deep signature verification and native CGEventTap diagnostic passed. Installed executable modified 09:14:07 EDT; fresh process after profile restoration started 09:14:43 EDT, PID 50635.

Through the installed app Settings UI, toggled completion alerts true -> false -> true, confirmed both persisted writes, and verified unknown top-level test table plus newer voice fields survived both. Removed only the temporary test table afterward. Separate voice-profile SHA remained identical across both app saves; runtime override loaded Kore/Morris/Extended Thinking and enabled tools. CLI dogfood source 0787ab1c, binary SHA256 0c0b488f6c90f57ce894cf7020c2858620e4dcb20678a4d86da2795d464c9893. Production bundle untouched; idle production process quit. No microphone session started. Full CI still pending; 66 config tests and Clippy pass locally.

PR #1027, opened 2026-09-17
Add optional verified Orukeet setup for the sherpa engine

Nathan-Roll1:
Add `minutes setup --sherpa-model orukeet` to install and select an optional INT8 model for the existing sherpa engine. It downloads a pinned [Hugging Face release](https://huggingface.co/oruk/orukeet), verifies the required manifest and every model file, and selects the model only after verification. The required integrity manifest fetch participates in Hugging Face’s normal download accounting. Valid cached files need no network. Defaults and inference/capture code remain unchanged; audio stays local, with no telemetry or extra counting requests.

Disclosure: I’m Nathan Roll, founder of Oruk; Orukeet is our model.

The installer now refuses legacy `setup --sherpa` writes into manifest-managed directories, including custom paths and symlink aliases. Downloads have exact byte limits and finite timeouts; temporary files protect existing files on failure. It reloads the current config after downloading so unrelated desktop edits survive. File replacement is atomic individually, not a transaction over the whole release.

The [published NeMo evaluation](https://huggingface.co/oruk/orukeet#evaluation) reports 9.85% versus 11.01% pooled FLEURS WER across 25 languages. This motivates the optional choice; it does not establish this export’s accuracy or speed in Minutes.

Validation on macOS arm64:

- Seven installer tests cover verified offline reuse, same-length corrupt manifests, bad checksums, interrupted/short/endless downloads, and a real HTTP server stalling mid-body.
- CLI regressions cover conflicting/invalid selectors, edits and malformed config during a download, protected custom directories, and symlink aliases with missing markers.
- Formatting, workspace Clippy with warnings denied, the CLI build, all 17 setup-related CLI tests, generated release constants, site checks, and the site production build pass.
- Full core suite: 1,876 passed, two failed, one ignored. The unchanged macOS child-ceiling test passed on isolated retry. The unchanged QMD stdout-budget assertion still fails in isolation, matching the previously reproduced failure on pristine upstream `cbe693a`. No unrelated code or tests were changed to suppress either failure.
- This updated CLI built with `engine-sherpa` fetched and verified the required Hugging Face manifest into an isolated cache, reused verified model weights, and completed two subsequent cached setups with networking denied. A corrupt manifest left config unchanged; legacy repair of a custom Orukeet directory with a missing encoder preserved every remaining file and the config.
- This updated CLI and the existing native sherpa plugin recognized English, German, Spanish, and French twice after cold loads with identical output, with networking denied. The same 14-process run returned empty text for 0.1/1/5-second silence using the existing `min_words=0` setting. The default three-word guard remains unchanged. GUI/microphone capture was not exercised; the isolated cache used the existing fixed-window fallback because it had no Silero VAD model.

Upstream CI for this update is pending.


silverstein:
I ran an adversarial pass over this. The trust chain and path handling came back clean.

One thing has to change before I merge.

Selecting Orukeet saves its directory as `sherpa_model_dir`, and our existing `cmd_setup_sherpa` uses that same value as its download target. That installer pulls from a different repo's mutable `resolve/main`, checks only minimum file sizes, and takes no hashes. It also doesn't take your `.install.lock`.

So: install Orukeet, have a file go missing or truncated, run `minutes setup --sherpa`, which is the repair the loader itself suggests, and unverified bytes land in the directory you just verified. The plugin loads them without hashing. That old downloader predates your PR, but pointing it at a verified install is new, and it undoes the thing the rest of this code is careful about. The verified release needs a directory that path can't target, or that path needs to refuse to write into a manifest-managed one.

Smaller stuff:

The command clones config from CLI startup, downloads for several minutes, then writes that whole stale struct back, so anything the desktop app changed in between gets overwritten. `cmd_setup_sherpa` already reloads after downloading, so this is just matching what's there.

`std::io::copy` reads to EOF and the length check happens after, with no timeouts. A request for what should be an 1,867 byte manifest can return an endless body and write until the disk fills, or stall while holding the install lock. Our own `download_file` has the same gap and I'm fixing that separately, but since this adds a new endpoint it's worth bounding here too.

`cmd_setup_orukeet` landed between an existing doc comment and `cmd_setup_sherpa`, so that comment now documents your function instead.

Two things that aren't defects. Could the flag be generic, something like `--sherpa-model orukeet`? Same functionality, but then the CLI doesn't name one model, and the next one is a value rather than another flag. And please add a line noting you founded Oruk. You told me by email and I've no issue with it at all, but the repo is public so it belongs here.

On the tests, neither blocking: "failures preserve existing files" holds per file rather than per release, so a good encoder followed by a failed decoder leaves a mixed install with no rollback. And the corrupt-manifest fixture has the wrong length as well as the wrong hash, so that test alone doesn't prove hash rejection, though your bad/new one does.

None of it's a rewrite. Happy to merge once the directory thing is sorted.

silverstein:
This is in. Thanks for taking every point seriously, including the ones I didn't ask for, like proving a damaged config during download stays untouched and the live-socket test for the timeout. Nice work.

PR #1051, opened 2026-09-23
fix(attendees): only people established as present get written

### LINKED ISSUE #245: Speaker attribution can override identity.name with a name discussed in transcript
## Summary

When the user has \`identity.name\` correctly set in \`config.toml\`, the diarization-to-speaker-name mapping can still misidentify them by picking up a different name mentioned in the transcript content and assigning it to their speaker.

## Reported by

@allaire in #227:

> the transcript succeeded, but it identifies me as "Charles" (a name that was discussed), even though my name is correctly in the setting.

## Why this matters

\`identity.name\` is supposed to be authoritative for the user's own speaker. The mapping should:
1. Bind the user's speaker label (typically the dominant speaker on their voice stem) to \`identity.name\`.
2. Bind other speaker labels to inferred names from the transcript only.

Today, the LLM-based name extraction can apparently latch onto a name mentioned in dialogue ("Charles") and assign it to the user's speaker, overriding the configured identity. This is the "wrong names are worse than anonymous" failure mode and erodes trust fast: the user sees their own meeting note attributed to someone else.

## Affected code paths

Best candidates to investigate:
- \`crates/core/src/diarize.rs\` (speaker attribution + speaker_map)
- \`crates/core/src/summarize.rs\` (SPEAKER_MAPPING_PROMPT and how identity is threaded into it)
- \`crates/core/src/pipeline.rs\` (where identity and diarization output are joined)

## Acceptance

With \`identity.name = "Allaire"\` in config:
- Recording where "Charles" is mentioned in dialogue: user's own utterances should be labeled "Allaire", other speakers labeled with their actual names if inferable.
- Confidence-aware system from CLAUDE.md: identity.name binding should be L0 confidence (deterministic, can't be overridden by L1 LLM suggestions).

Likely fix is to enforce that identity.name applies to the user's stem-dominant speaker before any LLM rewrite, and that the LLM mapping pass cannot rewrite the user's speaker label.

## Related

- #227 (parent report)
- CLAUDE.md notes "Wrong names are worse than anonymous, only High-confidence attributions rewrite transcript labels"

silverstein:
Fixes #245.

A reporter recorded a two person Teams call and got eight names in `attendees`. The extra six were people whose access permissions were being discussed on the call, not people on it.

## Root cause

`crates/core/src/summarize.rs` asks the model for:

```
6. Participants (names of people present or mentioned in the conversation)
```

That answer comes back as `summary.participants`, and `pipeline.rs` passes it to `merge_attendees(&frontmatter.attendees, &summary_participants)`, which appends every name with no confidence gate. `knowledge_extract.rs` then walks `normalized_attendees()` and files every decision in the meeting under every name in it, under a comment reading "We file them under each attendee present."

So six people who were absent were recorded as parties to decisions made without them. The model answered the question it was asked; the pipeline read "mentioned" as "attended".

It is not an attribution bug. The attribution system does gate on confidence and CLAUDE.md caps LLM name suggestions at Medium. This path routes around all of it, because those names never enter as attributions. They enter as a field treated as ground truth.

Both pipelines are affected: `write_transcript_artifact` and `process_with_progress_and_sidecar` both call `merge_attendees`. Batch CLI recordings most likely stayed clean because they matched a calendar event or nobody absent got named, not because that path is safe.

## Fix

`attribute_meeting_speakers` already takes trusted attendees and the merged LLM-derived list as separate arguments, so the seam existed. The merged list was just also the one being written to disk.

`confirmed_attendees` narrows only what gets persisted, to:

- the calendar event's attendee list (L0, deterministic), and
- speakers the attribution pass resolved at `Confidence::High`, the same bar CLAUDE.md sets for rewriting user-visible labels.

Everywhere a superset is harmless or useful the merged list is untouched: speaker mapping candidates, entity links, and the name correction pool. Mentioned people therefore keep flowing into `people:` and `entities.people` exactly as before. Only the field that carries a presence claim narrows.

## Behavior change worth calling out

A meeting with no calendar event and no high confidence attribution now writes an empty `attendees:` where it used to write the summarizer's guess. That is the intended answer under the project's "a wrong rewrite is worse than none" rule, not a regression, but it is a visible change and should be read as a deliberate one.

## Tests

Five cases on `confirmed_attendees`: mentioned names dropped, Medium and Low attributions dropped while High is kept, empty when nothing establishes presence, a high confidence speaker absent from the merged list still added, and case-insensitive matching that does not renormalize spelling.

Verified locally: parses and `rustfmt --check` clean. The compile and test run is CI's, since this VM does not build.

## Still open

The reporter's redacted frontmatter would confirm the ordering (two real participants first, six appended) and let a regression test pin that exact shape. Asked for in the issue thread.


silverstein:
Codex reviewed this and rejected it as a complete fix. It was right about the part that mattered, so the PR now has a second commit.

**What it caught.** Narrowing `attendees` did not close the bug. `knowledge_extract` never read that field for its presence claim: it asserted `Attended meeting` at Strong confidence for every entry in `entities.people`, and `entities.people` is built from the broad merged list. The six people discussed on the reporter's two person call still came out recorded as having been on it, through a different door. Presence now comes from the attendee list, and everyone else gets `Mentioned in meeting` at Inferred confidence instead. Being named in a meeting is worth recording; it is just not attendance.

**Also caught.** All five helper tests would have passed if someone reverted both call sites and went back to persisting the merged list. The helper was never the fix; calling it is. A guard test now pins the wiring at both persist points. Matching is also split-aware now, so a composite entry like `Gert and Liam` is findable from an attribution for either of them.

**One finding I could not reproduce.** The review said trusted `Dana` plus merged `Dana Whitfield` emits both. Tracing it, the merged entry is dropped and only `Dana` is emitted, so the list is narrowed rather than duplicated. The real cost in that case is losing the fuller spelling, which is a smaller problem than the one this PR exists to fix.

**Two findings deliberately left open.**

Calendar membership is treated as sufficient evidence of presence. An invitee who never joined is still written as an attendee, and decisions are still filed under them. That is a real gap, but the calendar is the L0 deterministic source CLAUDE.md already designates, so tightening it is a separate decision about what `attendees` means, not a bug fix. Worth its own issue.

The name correction participant gate still starts from the broad merged list, so a fuzzy transcript rewrite can still pull toward someone who was only discussed. Pre-existing, unchanged by this PR, and it deserves its own change rather than being folded into this one.

**Verification, stated honestly.** Parses and is `rustfmt --check` clean. I could not compile or run a single test: this VM does not build Minutes and stonebook's SDK is too old for minutes-core. Everything below the parse level is CI's word, not mine.

